PR #19, opened 2024-06-03
refactor(request): move api specific logic outside the fetch method

icatalina:
## What?

Move logic that is specific to a particular type of endpoint outside the generic fetch function.

## Why?

This type of coupling is messy, making it hard to reason about the way fetch works.
Moving this out makes each execution path easier to follow.

## Testing / Proof

When I learn how to run the functional tests, I will.
For the most part, this is just copy/paste.

Revert.

@bigcommerce/b2b-buyer-portal

### DIFF
--- a/apps/storefront/src/shared/service/request/fetch.ts
+++ b/apps/storefront/src/shared/service/request/fetch.ts
@@ -1,22 +1,16 @@
-import { snackbar } from '@/utils';
-
-import { RequestType } from './base';
-
 const originFetch = window.fetch;
 
-const responseResult = (path: string, res: any, resolve: any, init: any) => {
-  if (path.includes('current.jwt')) return res.text();
-
+const responseResult = (res: Response, resolve: (val?: unknown) => void, init: RequestInit) => {
   if (init.method === 'DELETE') {
     resolve();
   }
   return res.json();
 };
 
-function b3Fetch(path: string, init: any, type?: string, customMessage = false) {
+function b3Fetch(path: string, init: RequestInit): any {
   return new Promise((resolve, reject) => {
     originFetch(path, init)
-      .then((res: Response) => responseResult(path, res, resolve, init))
+      .then((res: Response) => responseResult(res, resolve, init))
       .then(async (res) => {
         if (res?.code === 500) {
           const data = res?.data || {};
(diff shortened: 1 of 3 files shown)


apps/storefront/src/shared/service/request/fetch.ts
```diff
@@ -1,51 +1,25 @@
...
-
 const originFetch = window.fetch;
 
-const responseResult = (path: string, res: any, resolve: any, init: any) => {
-  if (path.includes('current.jwt')) return res.text();
-
+const responseResult = (res: any, resolve: any, init: any) => {
   if (init.method === 'DELETE') {
     resolve();
   }
```
icatalina, line 6: This `init.method === 'DELETE'` is a bit weird, it'll mark any failed request as a success for the DELETE method.

I don't think that's desired, as it'll masquerade any errors while deleting.

apps/storefront/src/shared/service/request/fetch.ts
```diff
@@ -1,51 +1,25 @@
-import { snackbar } from '@/utils';
-
-import { RequestType } from './base';
-
 const originFetch = window.fetch;
 
-const responseResult = (path: string, res: any, resolve: any, init: any) => {
-  if (path.includes('current.jwt')) return res.text();
```
icatalina, line 8: The only place where we fetch `current.jwt` is here: 

https://github.com/bigcommerce/b2b-buyer-portal/blob/b4c3191363a2a60322ffcae539394e59d7951c31/apps/storefront/src/shared/service/bc/api/login.ts#L12-L13

Which uses a vanilla fetch.

apps/storefront/src/shared/service/request/fetch.ts
```diff
@@ -1,51 +1,25 @@
...
-
+const responseResult = (res: any, resolve: any, init: any) => {
```
bc-marco, line 3: It seems that you remove it `path` param instad of `init`, that's causing a problem

bc-marco (changes requested):
There is a minor error, please take a look

apps/storefront/src/shared/service/request/fetch.ts
icatalina: 🤦 we need tests

icatalina:
@bc-marco ♻️

deov31 (approved):
👍

bc-marco (approved):

PR #26, opened 2024-06-06
fix: add role=progressbar for loading indicators

bc-micah:
## What?

- add role=progressbar to loading indicators / text

## Why?

- allows our e2e test to properly wait
- follows the [accessibility standards](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role)

Revert

@bigcommerce/b2b-buyer-portal


deov31 (approved):
👍

bc-marco (approved):

PR #29, opened 2024-06-11
fix: memoize b3lang to avoid re-renders

icatalina:
## What?

Wrap the b3lang function in a `useCallback` to avoid re-renders.

## Why?

A lot of the re-renders that the app suffers are due to the b3lang function not being memoized.

After this change we should be able to remove some of the ignores for `react-hooks/exhaustive-deps`.

## Testing / Proof

This is a fairly innocuous change.

Revert this PR.

@bigcommerce/b2b-buyer-portal

### DIFF
--- a/apps/storefront/src/pages/registered/RegisteredFinish.tsx
+++ b/apps/storefront/src/pages/registered/RegisteredFinish.tsx
@@ -39,7 +39,7 @@ export default function RegisteredFinish(props: {
       return isAutoApproval ? (
         <StyleTipContainer>
           {b3Lang('global.registerFinish.autoApproved.tip', {
-            storeName,
+            storeName: storeName ?? '',
           })}
         </StyleTipContainer>
       ) : (
@@ -71,7 +71,7 @@ export default function RegisteredFinish(props: {
       return (
         <StyleTipContainer>
           {b3Lang('global.registerFinish.bcSuccess.tip', {
-            storeName,
+            storeName: storeName ?? '',
           })}
         </StyleTipContainer>
       );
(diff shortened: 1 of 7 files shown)


apps/storefront/src/utils/b3DateFormat/index.ts
```diff
@@ -38,7 +41,7 @@ const formatCreator =
       case 'parseDate':
         return fmt.parseDate(dateObject, display) || '';
       default:
-        return null;
+        throw new Error('Invalid value');
```
icatalina, line 44: This should never happen, the value of `handler` is protected by TS.

This removes `null` from the output of this method, removing some of the instances of the translated value being `null`.

packages/lang/useB3Lang.ts
```diff
@@ -1,26 +1,22 @@
+import { useCallback } from 'react';
 import { useIntl } from 'react-intl';
 
 export type LangFormatFunction = (
   id: string,
-  // TODO: If we pass `undefined | null`, the translation will probably be not correct.
-  // Ensure code using the values parameter do not pass `undefined | null`.
-  values?: Record<string, string | number | Date | undefined | null>,
+  values?: Record<string, string | number | Date>,
```
icatalina, line 6: As suggested by the deleted comment, `undefined` and `null` are not valid `values`.

apps/storefront/src/pages/registered/RegisteredFinish.tsx
```diff
@@ -71,7 +71,7 @@ export default function RegisteredFinish(props: {
       return (
         <StyleTipContainer>
           {b3Lang('global.registerFinish.bcSuccess.tip', {
-            storeName,
+            storeName: storeName ?? '',
```
icatalina, line 74: This is not worse than passing `null` to the translator. We should probably refactor the code so `storeName` is defined when getting to this part of the code.

bc-marco (approved):

bc-victor (approved):

PR #30, opened 2024-06-11
refactor: simplify usage of graphql fetchers

icatalina:
## What?

- Abstract some of the endpoint differences into the fetcher methods.
- Fix some typos.

## Why?

We use different methods if the user is b2b or not, some of these return different keys in their result. Homogenise some of the responses so the consumer needs to know less about the method they are calling.

Also fixed some typos.

## Testing / Proof

Will run the functional tests, also.

Revert

@bigcommerce/b2b-buyer-portal


bc-marco (approved):
Thanks!
It simplify a lot of code

This is just a minor suggestion, but could you also remove the parameter `channelId`?
That is a global variable, so it could imported directly on the graphql queries instead of the import it unnecessarily in the components files, like in `apps/storefront/src/pages/orderDetail/components/OrderShoppingList.tsx` or `apps/storefront/src/pages/quote/QuotesList.tsx`

icatalina:
Sure can do...

icatalina:
So, was having a look at the usages of `channelId`, and even though we can remove them, I'd prefer if we do so on a follow up PR. There's plenty more we could be removing.

deov31 (approved):

PR #31, opened 2024-06-11
refactor: use useRef instead of useState to define dropdown/popover targets

icatalina:
## What?

Replace `useState` with `useRef` when defining popover/dropdown targets.

## Why?

Using useState, removes the target upon closing, which impedes animations. This is also not the best practice in terms of storing/passing elements around.

## Testing / Proof

https://github.com/bigcommerce/b2b-buyer-portal/assets/4542735/b851b071-1b31-4552-acb4-fa2b56c0c994

Revert this PR

@bigcommerce/b2b-buyer-portal

### DIFF
--- a/apps/storefront/src/pages/orderDetail/components/OrderAction.tsx
+++ b/apps/storefront/src/pages/orderDetail/components/OrderAction.tsx
@@ -443,6 +443,7 @@ export default function OrderAction(props: OrderActionProps) {
             role={role}
             ipStatus={ipStatus}
             invoiceId={invoiceId}
+            key={item.key}
           />
         ))}
     </OrderActionContainer>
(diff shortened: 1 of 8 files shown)


apps/storefront/src/pages/orderDetail/components/OrderAction.tsx
```diff
@@ -443,6 +443,7 @@ export default function OrderAction(props: OrderActionProps) {
             role={role}
             ipStatus={ipStatus}
             invoiceId={invoiceId}
+            key={item.key}
```
icatalina, line 446: This was causing an error in the console, `key` props shouldn't be spread.

bc-victor (approved):

bc-marco (approved):

PR #38, opened 2024-06-15
fix(BUN-2614): fixing email verification validation

bc-victor:
## What?

Fixing a bug where email validation for approved company accounts is now broken. 

## Why?

Bug found in support

revert

@bigcommerce/b2b-buyer-portal


apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -185,7 +187,19 @@ function AccountSetting() {
...
+    const {
+      [key]: { userType },
+    }: CustomFieldItems = await fn({
+      email: emailValue,
+      channelId,
+    });
+
+    const isValid = !isBCUser ? [1].includes(userType) : ![2].includes(userType);
+
+    if (!isValid) {
```
bc-marco, line 202: @icatalina made a refactor about how to fetch data from graphql

Please take a look and take it as base for this kind of validation to decide which fetch should be request -> https://github.com/bigcommerce/b2b-buyer-portal/pull/30/files#diff-7ecfe89a8df11fedd9c965ef7ea2965c47158903d683ecdb90ab29a4722653ef

bc-marco (changes requested):

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -188,15 +190,25 @@ function AccountSetting() {
...
   const validateEmailValue = async (emailValue: string) => {
     if (customer.emailAddress === trim(emailValue)) return true;
+    const payload = {
+      email: emailValue,
+      channelId,
+    };
+
+    const { userType }: CustomFieldItems = isBCUser
+      ? await checkUserBCEmail(payload)
+      : await checkUserEmail(payload);
```
icatalina, line 200: This looks great, but we can further improve by exposing an `isValid` field directly from the two methods. It is quite cheap.

```tsx
export const checkUserEmail = (data: CustomFieldItems) =>
  B3Request.graphqlB2B({
    query: checkUserB2BEmail(data),
  }).then((res) => ({
    ...res.userEmailCheck,
    isValid: res.userEmailCheck.userType === UserTypes.DOESNT_EXIST,
  }));

export const checkUserBCEmail = (data: CustomFieldItems) =>
  B3Request.graphqlB2B({
    query: checkCustomerBCEmail(data),
  }).then((res) => ({
    ...res.customerEmailCheck,
    isValid: res.customerEmailCheck.userType !== UserTypes.B2C,
  }));
```

that will leave this code like:

```tsx
const { isValid } = isBCUser
  ? await checkUserBCEmail(payload)
  : await checkUserEmail(payload);
```

This kind of refactoring also exposes a missing concept in the API, the validity of a user. This logic should live in the API and not the consumer.

PR #48, opened 2024-06-20
refactor: remove unnecessary isVisible from the LoginWidget component

icatalina:
## What/Why?
Remove unnecessary prop, if we don't want to show the component, we should just not render it.

## Rollout/Rollback
Revert

## Testing
TS + Manual checks


bc-marco (approved):

deov31 (approved):
👍

PR #50, opened 2024-06-26
test: add a snapshot serializer for @emotion + update snapshots

icatalina:
## What/Why?
Without it, classes are less stable and can cause issues when re-running tests.

With it, it also includes the CSS for review at the top of the snapshot.

## Rollout/Rollback
Revert this PR

## Testing
This changes tests, you can explore the changes in the snapshot.


icatalina:
Opted for removing the snapshot test.

PR #53, opened 2024-06-27
fix: shopping list description bun2621

BrianJiang2021:
Jira: [BUN-2621](https://bigcommercecloud.atlassian.net/browse/BUN-2621)

## What/Why?
**Current issues:**

description is not escaped, graphql reports an error, description needs to be escaped

When we use strings in GraphQL queries or mutations, if the string contains special characters (such as newlines, double quotes, etc.), these characters may be mistaken as part of the syntax, resulting in parsing errors. For example, unescaped newlines can break the structure of the query and prevent it from parsing correctly

If we use newlines directly in the query, it will cause an error
mutation { updateDescription(description: "test\n") }
This causes parsing errors because newlines are mistaken for part of the syntax

To avoid this we need to use escape characters:
mutation { updateDescription(description: "test\\n") }
The GraphQL parser will recognize \n as an actual newline character

**Better way: use variables**

More convenient and avoids manual handling of escape characters, This allows GraphQL queries to be parsed and executed correctly

## Rollout/Rollback
undo this pr

## Testing

https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/c5560f75-a10c-43d7-be78-67a6965269b7

b2b:

https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/82cd1b67-bfff-4adf-a24a-3fdc25144dcb

bc:

https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/cbc961a3-b5c6-4ea4-8857-84438d077943

### DIFF
--- a/apps/storefront/src/pages/ShoppingListDetails/index.tsx
+++ b/apps/storefront/src/pages/ShoppingListDetails/index.tsx
@@ -230,7 +230,7 @@ function ShoppingListDetails({ setOpenPage }: ShoppingListDetailsProps) {
       const params: UpdateShoppingListParamsProps = {
         id: +id,
         name: shoppingListInfo?.name || '',
-        description: shoppingListInfo?.description || '',
+        description: shoppingListInfo?.description.split('\n').join('\\n') || '',
       };
 
       if (isB2BUser) {
--- a/apps/storefront/src/pages/ShoppingListDetails/index.tsx
+++ b/apps/storefront/src/pages/ShoppingListDetails/index.tsx
@@ -230,7 +230,7 @@ function ShoppingListDetails({ setOpenPage }: ShoppingListDetailsProps) {
       const params: UpdateShoppingListParamsProps = {
         id: +id,
         name: shoppingListInfo?.name || '',
-        description: shoppingListInfo?.description.split('\n').join('\\n') || '',
+        description: shoppingListInfo?.description || '',
       };
 
       if (isB2BUser) {
--- a/apps/storefront/src/pages/ShoppingLists/AddEditShoppingLists.tsx
+++ b/apps/storefront/src/pages/ShoppingLists/AddEditShoppingLists.tsx
@@ -72,18 +72,12 @@ function AddEditShoppingLists(
     handleSubmit(async (data) => {
       setAddUpdateLoading(true);
       try {
-        const { description } = data;
-        const submitData = data;
-        if (description.indexOf('\n') > -1) {
-          submitData.description = description.split('\n').join('\\n');
-        }
         const params: Partial<ShoppingListsItemsProps> = {
           ...data,
         };
 
         let fn = isB2BUser ? createB2BShoppingList : createBcShoppingList;
         let successTip = b3Lang('shoppingLists.addSuccess');
-
         if (type === 'edit') {
           if (isB2BUser) {
             fn = updateB2BShoppingList;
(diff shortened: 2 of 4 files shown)


icatalina:
hey @BrianJiang2021 , can you elaborate a bit more on the What and the Why? I expect some description of the error and why double escaping `\n` solves it.

BrianJiang2021:
> hey @BrianJiang2021 , can you elaborate a bit more on the What and the Why? I expect some description of the error and why double escaping `\n` solves it.

I have updated the 'what/why?' of pr

icatalina:
🤔 Shouldn't that escaping be handled by the backend?

BrianJiang2021:
> 🤔 Shouldn't that escaping be handled by the backend?

If the backend handles this, it will be more troublesome; if the frontend handles it, it will be relatively simple.
When the problem was first discovered, the result of the discussion with the backend colleagues was to handle it by the frontend.

kity-bundle (approved):

icatalina:
Hey @BrianJiang2021, can you explain why it is more troublesome for the BE to handle this? It looks like we are double escaping something just for the BE no to have to. The issue is, from now onward, that logic will live in the FE, and any client having to use this API will have to do the same escaping.

BrianJiang2021:
> Hey @BrianJiang2021, can you explain why it is more troublesome for the BE to handle this? It looks like we are double escaping something just for the BE no to have to. The issue is, from now onward, that logic will live in the FE, and any client having to use this API will have to do the same escaping.

In fact, there are already two places where double escaping has been done by the front end.
image
So here we just follow the old practice.

icatalina:
That's no explanation. And also not a valid reason to keep doing it. Can you, or some BE, explain why it is more complex to do it in the BE? there might be a good reason, but I want to hear it.

JamessenLiu (approved):

BrianJiang2021:
hey, @icatalina 
Regarding issues related to BE review again, if the previous method of passing parameters does not follow the request format standard of application/json, the back-end code cannot be reached, so I have made modifications again and added variables to make it better. Solve the problem of double definition

icatalina:
Hey @BrianJiang2021 !

I don't understand, is the BE handling the escaping now?

BrianJiang2021:
@icatalina 

**Current issues:** 

description is not escaped, graphql reports an error, description needs to be escaped


When we use strings in GraphQL queries or mutations, if the string contains special characters (such as newlines, double quotes, etc.), these characters may be mistaken as part of the syntax, resulting in parsing errors. For example, unescaped newlines can break the structure of the query and prevent it from parsing correctly

If we use newlines directly in the query, it will cause an error
`
mutation {
  updateDescription(description: "test\n")
}
`
This causes parsing errors because newlines are mistaken for part of the syntax

To avoid this we need to use escape characters:
`
mutation {
  updateDescription(description: "test\\n")
}
`
The GraphQL parser will recognize \\n as an actual newline character

**Better way: use variables**

More convenient and avoids manual handling of escape characters, This allows GraphQL queries to be parsed and executed correctly

bc-micah:
@icatalina - thanks for asking the team to fix the source of the problem.
Are you good with the solution @BrianJiang2021 has implemented to resolve the double escaping of strings?

From my understanding they switched to use variables instead of trying to inline the string into the query - which was the cause of the original issue and need for double escaping.

bc-micah:
@BrianJiang2021  - can you please update the PR description with the real fix that you just implemented?

cc: @LeoChowChina

icatalina:
can we please stop using sprint branches?

PR #56, opened 2024-06-28
test: add basic tests for the ForgotPassword page

icatalina:
## What/Why?
This commit sets the basis for testing across the app.

Adds several testing tools used to ensure the app works correctly.

## Rollout/Rollback
Revert this PR

## Testing
This PR is mostly about unit testing, the build should prove everything works.

### DIFF
--- a/apps/storefront/src/shared/service/request/fetch.ts
+++ b/apps/storefront/src/shared/service/request/fetch.ts
@@ -1,5 +1,3 @@
-const originFetch = window.fetch;
-
 const responseResult = (res: Response, resolve: (val?: unknown) => void, init: RequestInit) => {
   if (init.method === 'DELETE') {
     resolve();
@@ -9,7 +7,7 @@ const responseResult = (res: Response, resolve: (val?: unknown) => void, init: R
 
 function b3Fetch(path: string, init: RequestInit): any {
   return new Promise((resolve, reject) => {
-    originFetch(path, init)
+    fetch(path, init)
       .then((res: Response) => responseResult(res, resolve, init))
       .then(async (res) => {
         if (res?.code === 500) {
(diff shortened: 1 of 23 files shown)


apps/storefront/src/shared/service/request/fetch.ts
```diff
@@ -9,7 +7,7 @@ const responseResult = (res: Response, resolve: (val?: unknown) => void, init: R
 
 function b3Fetch(path: string, init: RequestInit): any {
   return new Promise((resolve, reject) => {
-    originFetch(path, init)
+    fetch(path, init)
```
icatalina, line 10: We need to use straight native fetch in order for MSW to work correctly

apps/storefront/src/components/RegisteredCloseButton.tsx
```diff
@@ -1,5 +1,6 @@
 import { Dispatch, SetStateAction, useContext } from 'react';
 import { useNavigate } from 'react-router-dom';
+import { useB3Lang } from '@b3/lang';
```
bc-marco, line 3: This import is not necessary

bc-marco (changes requested):
Overall looks good to me
Great test btw, just a couple of errors they seem to be related to fetch operations

image

apps/storefront/src/components/RegisteredCloseButton.tsx
icatalina: noticed this, but was afk, will fix in Monday

icatalina: fixed

icatalina:
Uh, this is interesting, they all pass fine on local:
image

I'll investigate a bit further

deov31:
I'm also seeing those errors while running locally the tests 👀 
image

icatalina:
sorted, was missing a few env. variables in `vite.config.ts`

bc-marco (approved):

deov31 (approved):

PR #59, opened 2024-06-28
refactor: add a PageProps interface to simplify the definition of the standard props for a page

icatalina:
## What/Why?
We are currently repeating the same definition on each page, and type checking the list of routes is a bit more complex than it needs to be.

This introduces a type for `PageProps` that all pages can conform to and can be used to type the RouteItem.

## Rollout/Rollback
Revert

## Testing
typescript

### DIFF
--- a/apps/storefront/src/components/B3Card.tsx
+++ b/apps/storefront/src/components/B3Card.tsx
@@ -1,16 +1,11 @@
-import { Dispatch, ReactNode, SetStateAction } from 'react';
+import { PropsWithChildren } from 'react';
 
-import { OpenPageState } from '@/types/hooks';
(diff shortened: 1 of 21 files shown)


apps/storefront/src/components/B3Card.tsx
```diff
@@ -1,16 +1,11 @@
-import { Dispatch, ReactNode, SetStateAction } from 'react';
+import { PropsWithChildren } from 'react';
 
-import { OpenPageState } from '@/types/hooks';
+import { type SetOpenPage } from '@/pages/SetOpenPage';
```
icatalina, line 3: In cases where this is not a whole page, use the `SetOpenPage` definition.

deov31 (approved):

PR #63, opened 2024-07-03
refactor: simplify login page (Step 1)

icatalina:
## What/Why?
This commit extracts some of the logic in the login outside of it and simplifies the syntax a bit.

Also simplifies the creation and use of `loginInfo`, ensuring it is always defined.

## Rollout/Rollback
Revert

## Testing
Most of the refactor here is automated.

Page still loads fine:
image

### DIFF
--- a/apps/storefront/src/pages/Login/index.tsx
+++ b/apps/storefront/src/pages/Login/index.tsx
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useState } from 'react';
+import { useCallback, useContext, useEffect, useState } from 'react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import { useB3Lang } from '@b3/lang';
 import { Alert, Box, ImageListItem } from '@mui/material';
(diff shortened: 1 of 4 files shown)


apps/storefront/src/pages/Login/index.tsx
```diff
@@ -443,7 +436,8 @@ export default function Login(props: PageProps) {
                           }}
                         >
                           <LoginPanel
-                            loginInfo={loginInfo}
+                            createAccountButtonText={loginInfo.createAccountButtonText}
+                            widgetBodyText={loginInfo.widgetBodyText}
```
icatalina, line 438: This ensures only necessary data is shared between components. It also helps with re-renders.

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -148,76 +169,42 @@ export default function Login(props: PageProps) {
...
       } finally {
         setLoading(false);
       }
     };
 
     init();
-    // disabling as we only need to run this on the first render and its causing infinite loops when loging in
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [b3Lang, endMasquerade, isLoggedIn, isMasquerade, searchParams]);
```
icatalina, line 179: This doesn't seem to cause infinite loops anymore:

https://github.com/bigcommerce/b2b-buyer-portal/assets/4542735/e7868291-779e-4b24-98e1-f590877aeb08

bc-victor (approved):

deov31:
Hey @icatalina overall looks good to me, I just noticed something really weird and I'm still not sure but I think it could related with this part https://github.com/bigcommerce/b2b-buyer-portal/pull/63/files#diff-fd4aa15ef6c953e480fa16141985f695bad80e0aad2b88b01a2163044e23c8d4R159-R160

After clicking Sign In is not displaying that "Loading" so it feels like nothing is happening after clicking the button until of course it is redirected to the account dashboard.
image

icatalina:
Hey @deov31, good catch, the `useEffect` for init was re-firing due to the `endMasquerade` function getting recreated on re-render. I fixed it by wrapping it in a `useCallback`.


https://github.com/bigcommerce/b2b-buyer-portal/assets/4542735/ae5765e1-7336-4ba7-b4eb-54bdab793864

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -30,31 +30,54 @@ import { getCurrentCustomerInfo } from '@/utils/loginInfo';
...
+
+const useMasquerade = () => {
+  const isAgenting = useAppSelector(({ b2bFeatures }) => b2bFeatures.masqueradeCompany.isAgenting);
+  const b2bId = useAppSelector(({ company }) => company.customer.b2bId);
+  const salesRepCompanyId = useAppSelector(({ b2bFeatures }) => b2bFeatures.masqueradeCompany.id);
+  const storeDispatch = useAppDispatch();
+
+  const isMasquerade = isAgenting && typeof b2bId === 'number';
+
+  const endMasquerade = useCallback(async () => {
```
icatalina, line 48: here's the fix for the loading not happening

deov31: Ah, of course. Thanks for the fix.

deov31 (approved):
thanks @icatalina !

PR #65, opened 2024-07-04
refactor: make showPageMask into a hook to remove the need to pass dispatch

icatalina:
## What/Why?
Create a hook to expose `showPageMask` avoiding the need for a method taking dispatch from outside.

## Rollout/Rollback
Revert this PR

## Testing
Manual


deov31 (approved):
👍

bc-marco (approved):

PR #66, opened 2024-07-09
feat: sprint57 Iteration task

BrianJiang2021:
Jira: [BUN-2601](https://bigc-b2b.atlassian.net/browse/BUN-2601)
[BUN-2662](https://bigc-b2b.atlassian.net/browse/BUN-2662)
[BUN-2621](https://bigcommercecloud.atlassian.net/browse/BUN-2621)

## What/Why?

1. description is not escaped, graphql reports an error, description needs to be escaped

When we use strings in GraphQL queries or mutations, if the string contains special characters (such as newlines, double quotes, etc.), these characters may be mistaken as part of the syntax, resulting in parsing errors. For example, unescaped newlines can break the structure of the query and prevent it from parsing correctly

If we use newlines directly in the query, it will cause an error
mutation { updateDescription(description: "test\n") }
This causes parsing errors because newlines are mistaken for part of the syntax

To avoid this we need to use escape characters:
mutation { updateDescription(description: "test\\n") }
The GraphQL parser will recognize \n as an actual newline character

**Better way: use variables**

More convenient and avoids manual handling of escape characters, This allows GraphQL queries to be parsed and executed correctly

2. B2B Edition Shopping Lists do not have "Add to Cart" Option while logged in as a B2C User

3. Improve Buyer Portal compatibility for Beatify Theme

When clicking the account icon, in most cases, you cannot enter the buyer portal page, but stay on a blank page.
Reason: The theme code is not compatible with the class name or other specific identifiers corresponding to the icon; this pr adds specific identifiers in the code to be compatible with the theme.

## Rollout/Rollback
undo pr

## Testing

1.
b2b:

https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/82cd1b67-bfff-4adf-a24a-3fdc25144dcb

bc:

https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/cbc961a3-b5c6-4ea4-8857-84438d077943

2.
c55c3198ebceca03e884e36fe12b992e

3.
https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/a769583d-d460-4173-bfdf-36cb44ef727a

### DIFF
--- a/apps/storefront/src/components/layout/B3Mainheader.tsx
+++ b/apps/storefront/src/components/layout/B3Mainheader.tsx
@@ -7,7 +7,7 @@ import { Box, Button, Typography } from '@mui/material';
 import { CART_URL } from '@/constants';
 import useMobile from '@/hooks/useMobile';
 import { CustomStyleContext } from '@/shared/customStyleButton';
-import { rolePermissionSelector, useAppSelector } from '@/store';
+import { isB2BUserSelector, rolePermissionSelector, useAppSelector } from '@/store';
 import b3TriggerCartNumber from '@/utils/b3TriggerCartNumber';
 
 import { getContrastColor } from '../outSideComponents/utils/b3CustomStyles';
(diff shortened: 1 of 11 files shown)


libruce (approved):
ATM

LeoChowChina (approved):

JamessenLiu (approved):
atm

PR #70, opened 2024-07-11
fix: set guest customerGroupId is 0

BrianJiang2021:
Jira: [BUN-2637](https://bigc-b2b.atlassian.net/browse/BUN-2637)

## What/Why?

When the user is guest, customerGroupId should be 0, not 1,
causing calculated_price to obtain an incorrect price

## Rollout/Rollback

undo pr

## Testing

https://github.com/bigcommerce/b2b-buyer-portal/assets/80307788/b797a4b4-e27f-4e3c-aa75-db2b9c785623

### DIFF
--- a/apps/storefront/src/store/slices/company.ts
+++ b/apps/storefront/src/store/slices/company.ts
@@ -35,7 +35,8 @@ const initialState: CompanyState = {
     firstName: '',
     lastName: '',
     emailAddress: '',
-    customerGroupId: 1,
+    /** the customerGroupId for the guest user is 0 */
+    customerGroupId: 0,
     role: CustomerRole.GUEST,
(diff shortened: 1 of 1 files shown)


apps/storefront/src/store/slices/company.ts
```diff
@@ -35,7 +35,7 @@ const initialState: CompanyState = {
     firstName: '',
     lastName: '',
     emailAddress: '',
-    customerGroupId: 1,
+    customerGroupId: 0,
```
deov31, line 38: hey @BrianJiang2021 only one comments, could you please add a comment here so anyone reading this understands why this should be set up with 0 as default?

deov31 (approved):

kity-bundle (approved):

PR #81, opened 2024-07-30
fix: update translation variable

sunilit42:
Issue - 
purchasedProducts.quickOrderPad.insufficientStockSku showing as msg when item not in stock

Result 
- It should be display translate msg

image


icatalina (approved):

deov31 (approved):

sunilit42:
First Pull request merge with bigcommerce

PR #88, opened 2024-08-07
fix: allocate different dependent modules

BrianJiang2021:
**Jira: [BUN-2784](https://bigc-b2b.atlassian.net/browse/BUN-2784)

## What/Why?

PageSpeed checks that the react-setup file takes too long to parse.
​
react-setup is unpacked to perform an optimization

## Rollout/Rollback
undo pr

## Testing

old: 

old

new:

new

**

### DIFF
--- a/apps/storefront/vite.config.ts
+++ b/apps/storefront/vite.config.ts
@@ -103,30 +103,33 @@ export default defineConfig(({ mode }) => {
             const { name } = info;
             return name.includes('headless') ? '[name].js' : '[name].[hash].js';
           },
+          manualChunks: {
+            reactVendor: ['react', 'react-dom'],
+            intl: ['react-intl'],
+            mui: ['@mui/material'],
+            muiIcon: ['@mui/icons-material'],
+            redux: ['react-redux'],
+            dateFns: ['date-fns'],
+            lang: ['@b3/lang'],
+            pdfobject: ['pdfobject'],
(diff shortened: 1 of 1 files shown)


libruce (approved):
ATM

apps/storefront/vite.config.ts
```diff
@@ -103,31 +103,34 @@ export default defineConfig(({ mode }) => {
...
+            pdf: ['react-pdf'],
+            toolkit: ['@reduxjs/toolkit'],
+            form: ['react-hook-form'],
+            router: ['react-router-dom'],
+            lodashEs: ['lodash-es'],
+            dropzone: ['react-dropzone'],
+            draggable: ['react-draggable'],
+            eCache: ['@emotion/cache'],
+            eReact: ['@emotion/react'],
+            eStyled: ['@emotion/styled'],
```
LeoChowChina, line 125: do we need to put `redux` and `toolkit` into one trunk?  So `dropzone` and `draggable` do.

LeoChowChina (approved):

kity-bundle (approved):

apps/storefront/vite.config.ts
BrianJiang2021: I think it is necessary because the main entry file uses redux and toolkit. After packaging, the main entry file is too large and needs to be split

PR #90, opened 2024-08-09
fix: empty Page showing when you back using browser button

BrianJiang2021:
Jira: [BUN-2742](https://bigc-b2b.atlassian.net/browse/BUN-2742)

## What/Why?
add to quote After the product is added, the browser returns a blank page. You should actually return to the pdp page

## Rollout/Rollback
undo pr

## Testing

https://github.com/user-attachments/assets/144b2bbb-e6dc-4d08-a1bd-54aa92964342

### DIFF
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -335,6 +335,21 @@ export default function App() {
   }, [isOpen]);
 
   useEffect(() => {
+    const { hash = '' } = window.location;
+
+    const handleHashChange = () => (!hash || hash === '#/') && setOpenPage({ isOpen: false });
+
+    window.addEventListener('hashchange', handleHashChange);
+
+    return () => {
+      window.removeEventListener('hashchange', handleHashChange);
+    };
+    // ignore setOpenPage
+    // due they are funtions that do not depend on any reactive value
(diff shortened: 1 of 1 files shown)


LeoChowChina (approved):

kity-bundle (approved):

apps/storefront/src/App.tsx
```diff
@@ -334,6 +334,21 @@ export default function App() {
...
+
+    window.addEventListener('hashchange', handleHashChange);
+
+    return () => {
+      window.removeEventListener('hashchange', handleHashChange);
+    };
+    // ignore setOpenPage
+    // due they are funtions that do not depend on any reactive value
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
icatalina, line 350: 🤔 what's the harm in honoring `setOpenPage` here? it'll just remount the event listener.

icatalina: I noticed we consistently override `react-hooks/exhaustive-deps`, this is an antipattern. We should be wrapping functions that we don't want to change in `useCallbacks` and adding them to the array. This can cause unexpected issues, like changes not updating the UI and incorrect function/values being executed/passed around.

PR #99, opened 2024-08-23
fix: updating message for po payments

deov31:
Jira: [BUN-1641](https://bigc-b2b.atlassian.net/browse/BUN-1641)

## What/Why?
I'm changing the message of the payments details when seeing any order detail because is confusing for buyers and sellers who still owe payment via Invoice

## Rollout/Rollback
Revert PR 

## Testing
image

### DIFF
--- a/apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
+++ b/apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
@@ -285,7 +285,18 @@ export default function OrderAction(props: OrderActionProps) {
     orderId,
     ipStatus = 0,
     invoiceId,
+    poNumber,
   } = detailsData;
+
+  const getPaymentMessage = () => {
+    const message = poNumber ? 'orderDetail.paidWithPo' : 'orderDetail.paidInFull';
+    return (
+      createAt &&
+      b3Lang(message, {
+        paidDate: displayFormat(createAt, true),
+      })
+    );
+  };
 
   if (!orderId) {
     return null;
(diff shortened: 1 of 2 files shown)


apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
```diff
@@ -285,8 +285,19 @@ export default function OrderAction(props: OrderActionProps) {
...
+  const getPaymentMessage = () => {
+    const message = poNumber ? 'orderDetail.paidWithPo' : 'orderDetail.paidInFull';
+    return (
+      createAt &&
+      b3Lang(message, {
+        paidDate: displayFormat(createAt, true),
+      })
+    );
+  };
+
```
bc-marco, line 300: As you're using a prop you should use `useCallback`
```suggestion
  const getPaymentMessage = useCallback(() => {
    const message = poNumber ? 'orderDetail.paidWithPo' : 'orderDetail.paidInFull';
    let message = ''
    
    if(createdAt) {
      if(poNumber) {
        message = b3Lang('orderDetail.paidWithPo', {
          paidDate: displayFormat(createAt, true),
        })
      } else {
        message = b3Lang('orderDetail.paidInFull', {
          paidDate: displayFormat(createAt, true),
        })
      }
    }

    return message
  }, [poNumber, createAt]);

```

apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
```diff
@@ -407,11 +418,7 @@ export default function OrderAction(props: OrderActionProps) {
     {
       header: b3Lang('orderDetail.payment'),
       key: 'payment',
-      subtitle: createAt
-        ? b3Lang('orderDetail.paidInFull', {
-            paidDate: displayFormat(createAt, true),
-          })
-        : '',
+      subtitle: getPaymentMessage() || '',
```
bc-marco, line 421: With the suggestion above there is no need to validate here, just use what is returned from `getPaymentMessage()`
```suggestion
      subtitle: getPaymentMessage(),
```

bc-marco (changes requested):
Overall looks good, there are suggestions to improve readability

Feel free to ask any questions

apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
deov31: 👍

bc-marco (approved):

bc-victor (approved):

PR #100, opened 2024-08-26
fix: fix invoice pay issue sup-1866

BrianJiang2021:
Jira: [SUP-1866](https://bigc-b2b.atlassian.net/browse/SUP-1866)

## What/Why?
Due to rounding, the data is biased. So when the back end verifies based on the front end data, the value does not match, resulting in an error.

Modified to: When calling the invoice payment interface, the value is not rounded.

## Rollout/Rollback
undo pr

## Testing
image


libruce (approved):
ATM

LeoChowChina (approved):

kity-bundle (approved):

PR #105, opened 2024-08-27
fix: fix logout issues

BrianJiang2021:
Jira: [BUN-2869](https://bigc-b2b.atlassian.net/browse/BUN-2869)

## What/Why?
fix logout issues
  Page flashing issue when exiting

## Rollout/Rollback
undo pr

## Testing

https://github.com/user-attachments/assets/049374d4-0448-42a3-86f8-6e1178019b71

### DIFF
--- a/apps/storefront/src/components/RegisteredCloseButton.tsx
+++ b/apps/storefront/src/components/RegisteredCloseButton.tsx
@@ -20,7 +20,8 @@ export default function RegisteredCloseButton(props: CloseButtonProps) {
   const navigate = useNavigate();
 
   const handleCloseForm = () => {
-    if (isCloseGotoBCHome) {
+    const isInLoginPage = window.location.hash.startsWith('#/login');
+    if (isCloseGotoBCHome || isInLoginPage) {
       window.location.href = '/';
     } else {
       navigate('/');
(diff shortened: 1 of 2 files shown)


apps/storefront/src/components/RegisteredCloseButton.tsx
```diff
@@ -20,6 +20,8 @@ export default function RegisteredCloseButton(props: CloseButtonProps) {
   const navigate = useNavigate();
 
   const handleCloseForm = () => {
+    const isInLoginPage = window.location.href.includes('#/login');
```
icatalina, line 23: 🤔 do we want to check for the hash?

`window.location.hash.startsWith('#/login')`

BrianJiang2021: Here is to determine whether it is on the login page

icatalina: yes, all I am saying is you can use the `location.hash` instead of `location.href`, and it makes the check easier

BrianJiang2021: Well, I have made the replacement, thanks for your suggestion.

kity-bundle (approved):

libruce (approved):
ATM

LeoChowChina (approved):

PR #106, opened 2024-08-29
fix: checkout link open register page instead of sending mail

BrianJiang2021:
Jira: [BUN-2841](https://bigc-b2b.atlassian.net/browse/BUN-2841)

## What/Why

The passwordless login link email is not being sent correctly. Instead, it redirects the user to the Buyer Portal account registration.  Because the Buyer Portal has changed the link to jump to other

The solution:
Buyer Portal only needs to change the login and registration links

## Rollout/Rollback
undo pr

## Testing

https://github.com/user-attachments/assets/8a6bed75-029e-430b-8f46-06b2d3f64c91

### DIFF
--- a/apps/storefront/src/hooks/useB3AppOpen.ts
+++ b/apps/storefront/src/hooks/useB3AppOpen.ts
@@ -64,6 +64,16 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
     return isSearchNode;
   };
 
+  const handleJudgeCheckoutNormalHref = (element: MouseEvent) => {
+    if (window?.location?.pathname !== CHECKOUT_URL) return false;
+
+    const target = element.target as HTMLAnchorElement;
+
+    if (target.getAttribute('href') && target.getAttribute('href') === '#') return true;
+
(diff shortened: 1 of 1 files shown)


apps/storefront/src/hooks/useB3AppOpen.ts
```diff
@@ -64,6 +64,23 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
...
+    const target = element.target as HTMLAnchorElement;
+    const checkoutNormalHref = ['href'];
+
+    checkoutNormalHref.forEach((key) => {
+      if (target.getAttribute(key) && target.getAttribute(key) === '#') {
+        isCheckoutNormalHref = true;
+      }
+    });
+
+    return isCheckoutNormalHref;
```
icatalina, line 81: do we need the for each and all this complexity to check a single attribute?

`element.target.getAttribute('href')`

we can add the loop later if needed

LeoChowChina (approved):

kity-bundle (approved):

PR #133, opened 2024-10-14
fix: create manual chunk for mui with all dependencies

bc-marco:
Jira: [B2B-1506](https://bigcommercecloud.atlassian.net/browse/B2B-1506)

## What/Why?

Fix bug by putting all MUI dependencies in one package

## Rollout/Rollback

Undo this PR

### Before fix
image

### After fix
image

[B2B-1506]: https://bigcommercecloud.atlassian.net/browse/B2B-1506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/vite.config.ts
+++ b/apps/storefront/vite.config.ts
@@ -106,7 +106,7 @@ export default defineConfig(({ mode }) => {
           manualChunks: {
             reactVendor: ['react', 'react-dom'],
             intl: ['react-intl'],
-            mui: ['@mui/material'],
+            mui: ['@emotion/react', '@emotion/styled', '@mui/material'],
             muiIcon: ['@mui/icons-material'],
             redux: ['react-redux'],
             dateFns: ['date-fns'],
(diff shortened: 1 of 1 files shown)


bc-micah (approved):

bc-victor (approved):

jp06:
@bc-marco Is this the fix for the exports issue from emotion? I was getting something like this on the production build which originates from a certain emotion-elements:
```
Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'exports')
```
Applying this change seems to have fixed the issue.

bc-marco:
> @bc-marco Is this the fix for the exports issue from emotion? I was getting something like this on the production build which originates from a certain emotion-elements:
> 
> ```
> Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'exports')
> ```
> 
> Applying this change seems to have fixed the issue.

Yes, this is the fix for that error

PR #146, opened 2024-10-30
feat: adding login/logout callbacks

deov31:
Jira: [B2B-1532](https://bigcommercecloud.atlassian.net/browse/B2B-1532)

## What/Why?
Right now for a next js store or any other headless store the buyer portal takes the responsibility to handle the session management. I'm adding new callbacks to allow a headless store using buyer portal login page to access a bc storefront login token so they can set up the token in their app context.

## Rollout/Rollback
Revert PR

## Testing

[B2B-1532]: https://bigcommercecloud.atlassian.net/browse/B2B-1532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bc-victor (approved):
validated and it works in my local

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -131,51 +132,52 @@ export default function Login(props: PageProps) {
...
+      setLoading(false);
+    } finally {
+      setLoading(false);
+      dispatchLogoutEvent();
+    }
+  });
+
+  useEffect(() => {
+    logoutEffect();
+  }, [logoutEffect]);
```
bc-micah, line 180: This one I am not sure is smart.

`useB2BCallback` does not memoize itself (it will regenerate the function on every render).

and now we are running this effect logic every render essentially.
This is a big departure from previous logic, which only ran this effect when these values changed:

`[b3Lang, endMasquerade, isLoggedIn, isMasquerade, searchParams]`.

@bc-victor - Am I missing something here? Shouldn't this logoutEffect function be created within the useEffect like it was being done before to make sure we aren't introducing some unexpected new behaviour?

bc-victor: Yeah that is a good observation - somehow React is not regenerating on every render which makes me think that the runtime is doing some memoization for us although not sure if it will work on every case.

@deov31 do you think you can follow Micah's suggestion and put the function back in the useEffect?

deov31: @bc-victor 👍

apps/storefront/tests/utils/eventManager.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    const consoleSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+    expect(cbManager.callbacks.size).toBe(0);
+    expect(consoleSpy).toBeCalledTimes(0);
+
+    // @ts-expect-error:next-line
+    cbManager.addEventListener(B2BEvent.OnLogin);
+
+    expect(cbManager.callbacks.size).toBe(0);
+    expect(consoleSpy).toHaveBeenCalled();
```
bc-micah, line 27:
```suggestion
    // @ts-expect-error:next-line
    expect(cbManager.callbacks.size).toBe(0);
    expect(consoleSpy).toBeCalledTimes(0);

    // @ts-expect-error:next-line
    cbManager.addEventListener(B2BEvent.OnLogin);

    // @ts-expect-error:next-line
    expect(cbManager.callbacks.size).toBe(0);
    expect(consoleSpy).toHaveBeenCalled();
```

this should fix the errors in build stage: https://github.com/bigcommerce/b2b-buyer-portal/actions/runs/11631477020/job/32392630075?pr=146

bc-victor (approved):

apps/storefront/src/utils/b3CallbackManager.ts
```diff
@@ -22,25 +22,31 @@ export default class CallbackManager {
    * @param callback The callback function to register.
    * @returns A unique hash identifying the registered callback.
    */
-  addEventListener(callbackKey: CallbackKey, callback: Callback): void {
+  addEventListener(callbackKey: B2BEvent, callback: Callback): void {
+    if (typeof callback !== 'function') {
+      console.error('callback should be a function');
```
bc-micah, line 27: I think we need to add eslint ignore for this.

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -145,36 +146,38 @@ export default function Login(props: PageProps) {
...
+            if (isMasquerade) {
+              await endMasquerade();
+            }
+          } catch (e) {
+            b2bLogger.error(e);
+          } finally {
+            // SUP-1282 Clear sessionStorage to allow visitors to display the checkout page
+            window.sessionStorage.clear();
+            logoutSession();
+            window.b2b.callbacks.dispatchEvent(B2BEvent.OnLogout, null);
```
bc-micah, line 170: we don't need to pass `null` since the data is optional.

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -212,98 +215,105 @@ export default function Login(props: PageProps) {
...
+            login: {
+              result: { token, storefrontLoginToken },
+              errors,
+            },
+          } = await b2bLogin({ loginData });
+
+          storeDispatch(setB2BToken(token));
+          customerLoginAPI(storefrontLoginToken);
+
+          dispatchOnLoginEvent(storefrontLoginToken);
```
bc-micah, line 254: can we pass the data as an object with `storefrontLoginToken` as a key/value so that if we every have to extend the login event in the future we don't need to create a new event.

packages/hooks/index.d.ts
```diff
@@ -2,8 +2,9 @@ declare interface Window {
   b2b: {
     callbacks: {
       dispatchEvent: (
-        callbackKey: import('./useCustomCallbacks').CallbackKey,
-        data: unknown,
+        callbackKey: import('./useB2BCallback').B2BEvent,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        data?: any,
```
bc-micah, line 7: maybe we should enforce the `Record` on the data field with keys being string, and values being anything.

https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

apps/storefront/src/utils/b3CallbackManager.ts
```diff
@@ -84,3 +90,5 @@ export default class CallbackManager {
     return success;
   }
 }
+
+export default new CallbackManager();
```
bc-micah, line 94: Is this a good idea?
Seems like it'd make it harder to unit test + not sure if this behaviour will be consistent.

bc-micah (changes requested):
I have some concerns - let me know what you think.

apps/storefront/src/utils/b3CallbackManager.ts
deov31: This changed was required because the b2b object is created under an `useEffect` and is listening to the `b2btoken` and i was losing the callbacks if i set up them before logging into the application.

bc-victor: You can see what Fani is referring to in the HeadlessController.tsx file. Basically, we create this window object in this effect that depends on the B2BToken in order to serve it in other functions, so this made the effect run when login logout happens and reset the class instance losing all your callbacks. 

By using exporting this singleton, we maintain the references across all renders.

bc-micah: @deov31 - can we solve that issue another way.

i.e create the singleton outside of useEffect & the component.

packages/hooks/index.d.ts
deov31: I pushed a change, but using with `Record` with any as values is basically the same as having that any. If we want be specific about types we would required to build a proper interface every single type that's being used or might be used in the future and would require way more effort.

bc-micah (approved):
Looks good!

PR #150, opened 2024-10-31
fix: validate platform before made login requests

bc-marco:
Jira: [B2B-1544](https://bigcommercecloud.atlassian.net/browse/B2B-1544)

## What/Why?

A validation was implemented to ensure that login requests are only made for`'bigcommerce'` platform stores.

## Rollout/Rollback

Revert this PR

## Testing

NA

[B2B-1544]: https://bigcommercecloud.atlassian.net/browse/B2B-1544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/Login/index.tsx
+++ b/apps/storefront/src/pages/Login/index.tsx
@@ -21,7 +21,7 @@ import {
 } from '@/store';
 import { setB2BToken } from '@/store/slices/company';
 import { CustomerRole, UserTypes } from '@/types';
-import { channelId, getB3PermissionsList, loginJump, platform, snackbar, storeHash } from '@/utils';
+import { channelId, getB3PermissionsList, loginJump, snackbar, storeHash } from '@/utils';
 import b2bLogger from '@/utils/b3Logger';
 import { logoutSession } from '@/utils/b3logout';
 import { deleteCartData } from '@/utils/cartUtils';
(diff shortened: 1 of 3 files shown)


bc-victor:
Hey this works but one question. Why was this token used and is it not needed in headless? Is this something that catalyst already owns or developers have to take care of?

bc-marco:
> Hey this works but one question. Why was this token used and is it not needed in headless? Is this something that catalyst already owns or developers have to take care of?

This is used in stores that use `bigcommerce` as platform but in the rest of platforms there is no need to use it as we rely in b2b graphql endpoint using b2b token

bc-victor (approved):

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -244,7 +244,10 @@ export default function Login(props: PageProps) {
         } = await b2bLogin({ loginData });
 
         storeDispatch(setB2BToken(token));
-        customerLoginAPI(storefrontLoginToken);
+
+        if (platform === 'bigcommerce') {
+          customerLoginAPI(storefrontLoginToken);
+        }
```
bc-micah, line 250: shouldn't we move this logic inside `customerLoginAPI` since that api call will only work when platform is `bigcommerce`?

otherwise we'd have to do this if statement if we use this api anywhere else also.

bc-marco: You're right, let me update that

PR #151, opened 2024-11-01
fix: handling password reset error on login

bc-victor:
Jira: [B2B-1559](https://bigcommercecloud.atlassian.net/browse/B2B-1559)

## What/Why?
Headless checkout login api call needs to show error message properly to the user when the request to login /internalapi/checkout/* fails.

## Rollout/Rollback
revert

## Testing
Screenshot 2024-10-31 at 9 36 00 p m

[B2B-1559]: https://bigcommercecloud.atlassian.net/browse/B2B-1559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/Login/index.tsx
+++ b/apps/storefront/src/pages/Login/index.tsx
@@ -222,11 +222,19 @@ export default function Login(props: PageProps) {
 
     if (isCheckout) {
       try {
-        await loginCheckout(data);
-        window.location.href = CHECKOUT_URL;
+        const response = await loginCheckout(data);
+
+        if (response.status === 400 && response.type === 'reset_password_before_login') {
+          b2bLogger.error(response);
+          await getForcePasswordReset(data.emailAddress);
+        } else {
+          window.location.href = CHECKOUT_URL;
+        }
       } catch (error) {
         b2bLogger.error(error);
-        getForcePasswordReset(data.emailAddress);
+        await getForcePasswordReset(data.emailAddress);
+      } finally {
+        setLoading(false);
       }
     } else {
       try {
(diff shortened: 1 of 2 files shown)


apps/storefront/src/pages/Login/index.tsx
```diff
@@ -222,11 +222,19 @@ export default function Login(props: PageProps) {
...
     if (isCheckout) {
       try {
-        await loginCheckout(data);
-        window.location.href = CHECKOUT_URL;
+        const response = await loginCheckout(data);
+
+        if (response.status === 400 && response.type === 'reset_password_before_login') {
+          b2bLogger.error(response);
+          await getForcePasswordReset(data.emailAddress);
+        } else {
```
bc-marco, line 230: I would change this to:
```suggestion
        if (!response.ok) {
          b2bLogger.error(response);
          if(response.type === 'reset_password_before_login') {
            await getForcePasswordReset(data.emailAddress);
          }
        } else {
```

bc-marco (changes requested):
Please check the suggestion below

apps/storefront/src/pages/Login/index.tsx
bc-victor: the problem with this is that there's more error types than that so I'd rather leave it like this - if a similar support ticket is raised where nothing happens we'll know that its a different case than this

bc-marco: It makes sense

bc-marco (approved):

Issue #24, opened 2024-06-05
 "Signature verification failed" error when trying to login as a user from the admin area of a BC store


cjsanchezBW:
**Please fill out this template to help us diagnose and address your issue effectively.**

**Type:** (Select one)
- [X] Bug Report
- [ ] Feature Request

**Summary:**

* Briefly describe the issue or feature request.
Platform is BigCommerce (not headless). Login as user not working for the open source version of Buyer portal.
**Expected Behavior:**
* Describe what you expect to happen under normal circumstances.
Logging in as a user via the admin area should allow you to navigate the website as if you had logged in normally as that user.
**Actual Behavior:**
From the BC admin go to the customers section and login as a user. You will see it's not possible to login.

* Describe what is actually happening. 
Cant access buyer portal pages after logging in as a user from the BC admin.

**Steps to Reproduce (Bug Reports Only):**

   1. List the specific steps necessary to reproduce the bug.

1 -Build the buyer portal code as instructed in the read me. Then deploy to the store's webdav.
2 - Try to login as a user via the admin area.
3- The app produces a blank loading screen and gets stuck

**Additional Information:**

* Include any other relevant details that might help us understand the issue.
Tested using chrome and firefox on windows.

**Backend Request Information (Optional):**

There's a graphQL request that comes back with an error when you attempt to log in as a user. We found that the same error comes up if you open a new tab and attempt to open to buyer portal.

   - If the issue involves interaction with a backend API, please provide details about the request:
      - API endpoint URL
      -https://api-b2b.bigcommerce.com/graphql

      - HTTP method (GET, POST, PUT, DELETE, etc.)
      Post
      - Request body (if applicable)
      {
[...]

PR #158, opened 2024-11-13
fix: app client id change in env example

bc-victor:
NO JIRA

## What/Why?
per this thread, this value is wrong https://bigcommerce.slack.com/archives/C01SAU2HH1V/p1731456396702289?thread_ts=1715630373.016079&cid=C01SAU2HH1V

## Rollout/Rollback
revert

### DIFF
--- a/apps/storefront/.env-example
+++ b/apps/storefront/.env-example
@@ -11,7 +11,7 @@ VITE_TRANSLATION_SERVICE_URL=https://api-b2b.bigcommerce.com
 VITE_CATPCHA_SETKEY=captcha_setkey
 
 # Client ID issued by B2B Edition for the storefront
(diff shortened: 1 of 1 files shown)


bc-marco (approved):

jp06:
@bc-victor @bc-marco Can you expand on why this value is updated? Currently, we're actually using `dl7c39mdpul6hyc489yk0vzxl6jesyx` as we encountered an issue with `qxvapwlk4fbb9dyogdcxk9o50d9jqjo`. This is also what seems to be used with B2B Edition footer script:
image

I think this is the issue I encountered when I used `qxvapwlk4fbb9dyogdcxk9o50d9jqjo`:
https://github.com/bigcommerce/b2b-buyer-portal/issues/24

PR #168, opened 2024-11-20
feat: add the script that will run with js app deployment process

bc-micah:
Jira: [B2B-1714](https://bigcommercecloud.atlassian.net/browse/B2B-1714)

## What/Why?
- use `config/deploy/after.sh` to create the revision for the buyer portal when deployed with lobby
- fixed typo
- added cspell to lintstaged

## Rollout/Rollback
reverting this PR should be enough

## Testing
Video of integration and staging stores working

https://github.com/user-attachments/assets/df824947-6847-4aa1-a2fb-d782aaf20e1f

[B2B-1714]: https://bigcommercecloud.atlassian.net/browse/B2B-1714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bc-victor (approved):
looks good, thanks for the improvements - let's just notify partners for the next release

apps/storefront/src/shared/service/request/base.ts
```diff
@@ -1,6 +1,18 @@
...
+  local: import.meta.env.VITE_B2B_URL ?? 'http://localhost:9000',
+  integration: 'https://api-b2b.integration.zone',
+  staging: 'https://api-b2b.staging.zone',
+  production: 'https://api-b2b.bigcommerce.com',
+};
+
+function getAPIBaseURL() {
+  const environment: Environment = window.B3.setting.environment ?? Environment.Production;
+  return ENVIRONMENT_B2B_API_URL[environment];
+}
```
bc-marco, line 13: This function it's creating problems to run the project locally

Where should I configure this value?
Run it with this branch
image

Modify the branch
image
image

bc-marco (changes requested):
Found an error that doesn't let me to run the project locally
I tried to change the value of `window.B3.setting.environment` but I couldn't find where, could you update the readme

apps/storefront/src/shared/service/request/base.ts
bc-micah: Ah good point - local needs to have this set to Local.
I'll add a `VITE_ENVIRONMENT` for local

bc-micah: I made it default to `Environment.Local` so that should help.

apps/storefront/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "tsc && vite build",
     "build:staging": "tsc && vite build --mode staging",
     "build:tier1": "tsc && vite build --mode tier1",
-    "lint": "yarn lint:eslint && yarn lint:dependencies",
+    "lint": "yarn lint:dependencies && yarn lint:eslint",
```
bc-marco, line 12: Why the order changed?

bc-marco (approved):
LGTM
Thanks for the improvements! Just a question below

PR #189, opened 2024-12-16
feat: catalyst redirect

bc-victor:
Jira: [B2B-1932](https://bigcommercecloud.atlassian.net/browse/B2B-1932)

## What/Why?
Catalyst cart redirect

## Rollout/Rollback
revert

## Testing
https://github.com/user-attachments/assets/e561d434-0281-43a0-a1c1-d5112a7fa54c

[B2B-1932]: https://bigcommercecloud.atlassian.net/browse/B2B-1932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,4 +1,4 @@
 {
   "**/*.json": "prettier --write",
-  "**/*.{js,jsx,ts,tsx,md}": "npx cspell lint"
+  "*.{js,jsx,ts,tsx,md}": "npx cspell lint"
 }
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,4 +1,4 @@
 {
   "**/*.json": "prettier --write",
-  "*.{js,jsx,ts,tsx,md}": "npx cspell lint"
+  "**/*.{js,jsx,ts,tsx,md}": "npx cspell lint"
 }
--- a/apps/storefront/.lintstagedrc.json
+++ b/apps/storefront/.lintstagedrc.json
@@ -1,4 +1,3 @@
 {
-  "*.{ts,tsx,js}": "eslint --fix",
-  "*.{js,jsx,ts,tsx,md}": "npx cspell lint"
+  "*.{ts,tsx,js}": "eslint --fix"
 }
--- a/apps/storefront/.lintstagedrc.json
+++ b/apps/storefront/.lintstagedrc.json
@@ -1,3 +1,4 @@
 {
-  "*.{ts,tsx,js}": "eslint --fix"
+  "*.{ts,tsx,js}": "eslint --fix",
+  "*.{js,jsx,ts,tsx,md}": "npx cspell lint"
 }
--- a/apps/storefront/src/components/layout/B3Mainheader.tsx
+++ b/apps/storefront/src/components/layout/B3Mainheader.tsx
@@ -4,11 +4,11 @@ import { B2BEvent, useB2BCallback } from '@b3/hooks';
 import { useB3Lang } from '@b3/lang';
 import { Box, Button, Typography } from '@mui/material';
 
+import { CART_URL } from '@/constants';
 import useMobile from '@/hooks/useMobile';
 import { CustomStyleContext } from '@/shared/customStyleButton';
 import { isB2BUserSelector, rolePermissionSelector, useAppSelector } from '@/store';
 import b3TriggerCartNumber from '@/utils/b3TriggerCartNumber';
-import getCartUrl from '@/utils/getCartUrl';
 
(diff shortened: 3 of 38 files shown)


apps/storefront/src/components/layout/B3Mainheader.tsx
```diff
@@ -15,6 +15,7 @@ import { getContrastColor } from '../outSideComponents/utils/b3CustomStyles';
 import B3AccountInfo from './B3AccountInfo';
 import B3StatusNotification from './B3StatusNotification';
 
+// Cspell:ignore Mainheader
```
bc-micah, line 18: This one we should just refactor rename - no reason to keep this name.

apps/storefront/src/components/layout/B3Mainheader.tsx
```diff
@@ -33,8 +34,10 @@ export default function B3Mainheader({ title }: { title: string }) {
     },
   } = useContext(CustomStyleContext);
 
+  // Cspell:ignore purchasability
```
bc-micah, line 37: This word is reused in a lot of places - so worth adding to cspell.json file

apps/storefront/src/components/layout/B3Mainheader.tsx
```diff
@@ -45,7 +48,7 @@ export default function B3Mainheader({ title }: { title: string }) {
       return;
     }
 
-    window.location.href = CART_URL;
+    window.location.href = window.B3.setting.platform === 'catalyst' ? '/cart' : CART_URL;
```
bc-micah, line 51: @bc-victor - shouldn't we provide a way to customize this for other platforms?
I think we should change the `CART_URL` variable dynamically based on `window.B3.setting.cart_url`.

The fallback value of `CART_URL` when `window.B3.setting.cart_url` is not defined can be `/cart.php` when platform is `bigcommerce` vs `/cart` for other platforms.

bc-micah (changes requested):
Please see my comments

apps/storefront/src/utils/getCartUrl.ts
```diff
@@ -0,0 +1,8 @@
+export default function getCartUrl() {
+  const {
+    setting: { platform, cart_url: cartUrl },
+  } = window.B3;
+  const fallbackValue = platform === 'bigcommerce' ? '/cart.php' : '/cart';
+
+  return cartUrl || fallbackValue;
```
bc-micah, line 7: @bc-victor 

can we move this logic to the `src/constants/index.ts` file?

and replace the `CART_URL` value there? That should allow you to not need to change all the places it is currently used.

bc-victor:
@bc-micah I disabled cspell as it introduced several errors, we should tackle them in a separate PR before incorporating again

apps/storefront/.lintstagedrc.json
```diff
@@ -1,4 +1,3 @@
 {
-  "*.{ts,tsx,js}": "eslint --fix",
-  "*.{js,jsx,ts,tsx,md}": "npx cspell lint"
```
icatalina, line 3: ? why remove the spell check?

bc-victor: Is there a way I can commit without this rule? this is a really small change, if I leave it on It starts flagging a bunch of cases in my local environment and the PR would get bigger and kind of lose its purpose

bc-micah: good callout ignacio. I discussed w/ victor - and have a plan of keeping spell check as the goal is to keep improving our spelling issues through each PR.

So the PR having changes outside it's original scope is expected (although if it gets out of hand we can ignore specific files/words/ come up with a better strategy other than disabling it fully)..

bc-victor: Just added back the setting :+1:

bc-micah (approved):

.lintstagedrc.json
```diff
@@ -1,4 +1,4 @@
 {
   "**/*.json": "prettier --write",
-  "**/*.{js,jsx,ts,tsx,md}": "npx cspell lint"
+  "*.{js,jsx,ts,tsx,md}": "npx cspell lint"
```
bc-micah, line 3: oh this change should also be reverted.

my bad I missed this. otherwise - looks good!

bc-micah (changes requested):
my bad - I missed a small change - `.lintstagedrc.json` files still have some issues.

bc-micah (approved):

deov31 (approved):

PR #242, opened 2025-02-05
refactor(account-settings): refactor formValidation, updateAccountSettings & dataProcessing

icatalina:
Jira: [B2B-2158](https://bigcommercecloud.atlassian.net/browse/B2B-2158)

#### refactor(account-settings): reorganise account settings update

Separate payload preparation from the update process.
Purify and simplify handleGetUserExtraFields.

#### refactor(account-settings): separate formValidation from error notifications

The current logic conflates form validation with form error
display/translations.

We should probably include some form library, but for now we'll just
separate the methods doing the validation from the methods displaying
the errors/alerts.

#### refactor(account-settings): separate getAccountSettingsFields & getPasswordModifiedFields

These two methods are used at different stages of the process and for
different reasons. Separating them allows us to only execute them when
they are needed and more importantly _when_ they are needed.

eg. the `getAccountSettingFiles` is not required for bc users.

Remove the configuration for the `xs` prop, the only use of it is `12`.

Translate the passwordFields directly upon receiving them, as opposed to
doing it in the component.

Hard code the list of fields directly in the method, we had a method for
indirection that wasn't needed.

Bonus: flip all the `!isBcUser` ternaries

#### refactor(account-settings): rename isEdit to pristine in the dataProcessing methods

Current code uses `isEdit`, but then exposes it as `!isEdit`, this is
ultimately confusing. Change the flag name to `pristine`, and change the
method to either return the payload, if we need to save it, or undefined
if not modified.

## Rollout/Rollback
Revert

## Testing
Manual

[B2B-2158]: https://bigcommercecloud.atlassian.net/browse/B2B-2158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/AccountSetting/config.ts
+++ b/apps/storefront/src/pages/AccountSetting/config.ts
@@ -18,47 +18,15 @@ export interface GetFilterMoreListProps {
   size: string;
 }
 
-interface GetAccountSettingFilesReturnProps {
-  accountB2BFormFields: GetFilterMoreListProps[];
-  passwordModified: GetFilterMoreListProps[];
-}
-
-interface PasswordKeysProps {
-  name: string;
-  label: string;
-  idLang: string;
-}
-
-export const getPasswordKeys = (): PasswordKeysProps[] => [
-  {
-    name: 'currentPassword',
-    label: 'Current Password',
-    idLang: 'accountSettings.form.currentPassword',
-  },
-  {
-    name: 'password',
-    label: 'Password',
-    idLang: 'accountSettings.form.password',
-  },
-  {
-    name: 'confirmPassword',
-    label: 'Confirm Password',
-    idLang: 'accountSettings.form.confirmPassword',
-  },
-];
-
-export const getAccountSettingFiles = (
-  xs: number,
-  b3Lang: LangFormatFunction,
-): GetAccountSettingFilesReturnProps => {
-  const accountB2BFormFields = [
+export const getAccountSettingsFields = (b3Lang: LangFormatFunction): GetFilterMoreListProps[] => {
+  return [
     {
       name: 'company',
       label: b3Lang('accountSettings.form.company'),
       required: false,
       default: '',
       fieldType: 'text',
-      xs,
+      xs: 12,
       variant: 'filled',
       size: 'small',
     },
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -237,10 +223,32 @@ function AccountSetting() {
...
+        if (!isValid) {
+          setError('email', {
+            type: 'custom',
+            message: b3Lang('accountSettings.notification.emailExists'),
+          });
+        }
+
         const emailFlag = emailValidation(data);
 
+        if (!emailFlag) {
```
leeBigCommerce, line 232: these `flag` names aren't very descriptive, what do you think to turning them into something like

```tsx
if (isValidEmail(data.email)) { 
    // etc
```

in another commit/PR/ticket?

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -132,6 +130,7 @@ function AccountSetting() {
 
         const { [key]: accountSettings } = await fn(params);
 
+        const accountFormAllFields = await getB2BAccountFormFields(isBCUser ? 1 : 2);
```
leeBigCommerce, line 133: "how many fields do you want? One, two..?" 😆

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -40,11 +40,42 @@ function useData() {
...
 
-  return { isBCUser, companyId, customer };
+  const validateEmailValue = async (emailValue: string) => {
+    if (customer.emailAddress === trim(emailValue)) return true;
+    const payload = {
+      email: emailValue,
+      channelId,
+    };
+
+    const { isValid }: { isValid: boolean } = isBCUser
```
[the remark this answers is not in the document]
icatalina: the problem is this functions are used in many places, we'll need to update the definition, but that creates a cascade of types that I'll prefer to postpone to a future PR.

apps/storefront/src/pages/AccountSetting/index.tsx
icatalina: can do

apps/storefront/src/pages/AccountSetting/index.tsx
icatalina: I wish that was what it means...

leeBigCommerce (approved):

apps/storefront/src/pages/AccountSetting/config.ts
```diff
@@ -86,26 +54,44 @@ export const getAccountSettingFiles = (
...
-    variant: 'filled',
-    size: 'small',
-    idLang: item.idLang,
-  }));
-
-  return {
-    accountB2BFormFields,
-    passwordModified,
-  };
+export const getPasswordModifiedFields = (b3Lang: LangFormatFunction): GetFilterMoreListProps[] => {
```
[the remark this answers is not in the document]
icatalina: all this is gonna be absorbed by `useData`, we can rethink it after

PR #243, opened 2025-02-06
refactor(quote-list): extract getQuotesList to useData

icatalina:
Jira: [B2B-2149](https://bigcommercecloud.atlassian.net/browse/B2B-2149)

## What/Why?
﻿﻿﻿### refactor(quote-list): extract getAvailableFilters to useData

Extract the getAvailableFilters method, remove some of the indirection
and further separate the b2b path from the b2c path.

## Rollout/Rollback
Revert

## Testing
Manual

[B2B-2149]: https://bigcommercecloud.atlassian.net/browse/B2B-2149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/QuotesList/index.tsx
+++ b/apps/storefront/src/pages/QuotesList/index.tsx
@@ -147,7 +147,15 @@ function useData() {
 
   const companyId = companyB2BId || salesRepCompanyId;
 
-  return { companyId, isB2BUser, draftQuoteListLength, customer };
+  const getQuotesList = (
+    params: Partial<FilterSearchProps>,
+  ): ReturnType<typeof getB2BQuotesList | typeof getBCQuotesList> => {
+    return isB2BUser
+      ? getB2BQuotesList({ ...params, channelId })
+      : getBCQuotesList({ ...params, channelId });
+  };
+
+  return { companyId, isB2BUser, draftQuoteListLength, customer, getQuotesList };
 }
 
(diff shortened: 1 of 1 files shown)


apps/storefront/src/pages/QuotesList/index.tsx
```diff
@@ -147,7 +147,15 @@ function useData() {
 
   const companyId = companyB2BId || salesRepCompanyId;
 
-  return { companyId, isB2BUser, draftQuoteListLength, customer };
+  const getQuotesList = (
+    params: Partial<FilterSearchProps>,
+  ): ReturnType<typeof getB2BQuotesList | typeof getBCQuotesList> => {
```
leeBigCommerce, line 85: isn't this just giving us a return type of `any`?

also, wouldn't it'd be better for `getQuotesList` to adhere to a type owned by the view (defining what the view needs) rather than the being tied to a type of an implementation detail?

icatalina: let me change this

icatalina: actually, having a look, this creates a cascading effect with types... gonna keep it like this for now, and fix it in a future PR.

leeBigCommerce (approved):

PR #247, opened 2025-02-10
B2B-2160 Refactor B3Table and B3TablePaginationTable towards type safety

leeBigCommerce:
Jira: [B2B-2160](https://bigcommercecloud.atlassian.net/browse/B2B-2160)

## What/Why?

- `B3Table` and `B3TablePaginationTable` made prolific use of `any`, losing type safety and a clear view of values
- This PR moves away from the use of `any` by adding more accurate types
- 99% just type definition changes, except for a new `isNodeWrapper` type-guard

**N.B.**

Due to the 'quirky' implementation of these components, some states cannot be legally represented by Typescript.
E.g. if a `selectedSymbol` prop is defined, then it must be a populated key on the `Row` value object

Typescript _could_ represent this if the prop was mandatory, but this is not the case. As such `B3Table` and `B3TablePaginationTable` have numerous type suppressions such as:

```tsx
<Checkbox
  // @ts-expect-error typed previously as an any
  checked={selectCheckbox.includes(node[selectedSymbol])}
  // etc etc
/>
```

The idea is to make these bits of code redundant through future refactoring. 

## Rollout/Rollback

- Revert this PR

## Testing

- Manual tests

[B2B-2160]: https://bigcommercecloud.atlassian.net/browse/B2B-2160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/components/table/B3PaginationTable.tsx
+++ b/apps/storefront/src/components/table/B3PaginationTable.tsx
@@ -1,4 +1,5 @@
 import {
+  FC,
   ReactElement,
   Ref,
   useCallback,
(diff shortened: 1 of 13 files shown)


apps/storefront/src/components/table/B3PaginationTable.tsx
```diff
@@ -131,24 +138,25 @@ function PaginationTable<GetRequestListParams, ListItem>(
...
       const copyCacheAllList = [...cacheAllList];
 
-      edges.forEach((item: CustomFieldItems) => {
-        const option = item?.node || item;
-        const isExist = cacheAllList.some((cache: CustomFieldItems) => {
-          const cacheOption = cache?.node || cache;
+      edges.forEach((item) => {
+        const option = isNodeWrapper(item) ? item.node : item;
+        const isExist = cacheAllList.some((cache) => {
+          const cacheOption = isNodeWrapper(cache) ? cache.node : cache;
```
icatalina, line 158: 🤔 instead of a typeguard, should we write some unmarshaller? 

`getWithPossibleNode(cache)`

icatalina (approved):

apps/storefront/src/components/table/B3PaginationTable.tsx
leeBigCommerce: I considered that, but went with this as it's closer to the original code and I wanted to keep this PR as tiny/easy to read as possible

I really wanted to do this with PR with just type definitions and no code changes, but this one forced my hand

PR #252, opened 2025-02-14
refactor: lift default value setting to the AccountSettings page

icatalina:
Jira: [B2B-2158](https://bigcommercecloud.atlassian.net/browse/B2B-2158)

## What/Why?
This further separates the fetching of the field information from the data. Allowing for a cleaner API separation.

## Rollout/Rollback
Revert

## Testing
Manual

image

[B2B-2158]: https://bigcommercecloud.atlassian.net/browse/B2B-2158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/AccountSetting/index.tsx
+++ b/apps/storefront/src/pages/AccountSetting/index.tsx
@@ -3,6 +3,7 @@ import { useForm } from 'react-hook-form';
 import { useNavigate } from 'react-router-dom';
 import { useB3Lang } from '@b3/lang';
 import { Box } from '@mui/material';
+import { keyBy, mapValues } from 'lodash-es';
 import trim from 'lodash-es/trim';
 
 import { B3CustomForm } from '@/components';
@@ -23,12 +24,12 @@ import {
 } from '@/shared/service/b2b';
 import { isB2BUserSelector, useAppSelector } from '@/store';
 import { Fields, ParamProps } from '@/types/accountSetting';
-import { B3SStorage, channelId, snackbar } from '@/utils';
+import { B3SStorage, channelId, snackbar, validatorRules } from '@/utils';
 
 import { deCodeField, getAccountFormFields } from '../Registered/config';
 
 import { getAccountSettingsFields, getPasswordModifiedFields } from './config';
-import { b2bSubmitDataProcessing, bcSubmitDataProcessing, initB2BInfo, initBcInfo } from './utils';
+import { b2bSubmitDataProcessing, bcSubmitDataProcessing } from './utils';
 
 function useData() {
   const isB2BUser = useAppSelector(isB2BUserSelector);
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -280,11 +275,37 @@ function AccountSetting() {
...
+      field_email: emailValidate,
+    };
+
+    const formFields = mapValues(keyBy(accountSettings.formFields, 'name'), 'value');
+    const extraFields = mapValues(keyBy(accountSettings.extraFields, 'fieldName'), 'fieldValue');
+
     return accountInfoFormFields.map((item) => ({
       ...item,
+      default:
+        defaultValueMap[item.fieldId ?? ''] ??
```
leeBigCommerce, line 297: aren't `fieldId`s mandatory?

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -280,11 +275,37 @@ function AccountSetting() {
...
+        formFields[item.bcLabel ?? ''] ??
+        extraFields[item.label ?? ''] ??
+        item.default ??
```
leeBigCommerce, line 300: I'm struggling to mentally map this one

is there a possibility of (A) multiple objects (`defaultValueMap`, `formFields`, etc) holding a value for a single `item`?
Or (B) would we only expect _one_ of them to contain a value?


---
if (A) the ordering of the object look ups matters, e.g. 
`formFields[item.bcLabel] ?? extraFields[item.label]` vs 
`extraFields[item.label] ?? formFields[item.bcLabel]`

---

if (B) the ordering _doesn't_ matter, as we only every expect one of them to be truth-y

---

at the moment I don't if we need to care about the ordering or not 🤔

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -133,32 +136,24 @@ function AccountSetting() {
...
-        const roleItem = all.find((item) => item.name === 'role');
-
-        if (roleItem?.fieldType) roleItem.fieldType = 'text';
+          ? [...contactInformation, ...additionalInformation, ...getPasswordModifiedFields()]
+          : [
+              ...contactInformation,
+              ...getAccountSettingsFields(),
+              ...additionalInformation,
+              ...getPasswordModifiedFields(),
+            ];
```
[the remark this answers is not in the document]
icatalina: Not a huge fan of this pattern because of readability, but sure...

apps/storefront/src/pages/AccountSetting/index.tsx
leeBigCommerce: gotcha, I guess ideally we'd go for something that makes that clearer in code, e.g.

```tsx
    const defaultValuesByType = {
      defaultValueMap,
      formFields,
      extraFields,
    };

    return accountInfoFormFields.map((item) => ({
      ...item,
      default: defaultValuesByType[item.type][item.label ?? ''] ?? item.default ?? '',
     // etc etc
```

but these types are a long way from allowing something like that (or something better)

worth a comment to explain the mutually exclusivity in the mean time?

leeBigCommerce (approved):

PR #256, opened 2025-03-07
feat: color palette for catalyst platforms

bc-victor:
## What/Why?
Color palette for catalyst platforms only.

## Rollout/Rollback
Revert

## Testing
Screenshot 2025-03-07 at 3 23 42 p m


apps/storefront/src/components/table/B3NoData.tsx
```diff
@@ -30,11 +31,20 @@ export default function B3NoData({
...
   minHeight,
+  showBorder = false,
   isLoading = false,
 }: B3NoDataProps) {
   const b3Lang = useB3Lang();
   return (
-    <NoDataContainer backgroundColor={backgroundColor} minHeight={minHeight}>
+    <NoDataContainer
+      style={{
+        boxShadow: showBorder
```
icatalina, line 41: Not a huge fan of these kind of booleans, it might be better to make this part of the theme, something like `panelShadow`. Just give it a name and toggle it as part of theme.

```tsx
// theme.tsx

declare module '@mui/material' {
  interface Theme {
    styles: {
      panelShadow: string;
    };
  }
  interface ThemeOptions {
    styles: {
      panelShadow: string;
    };
  }
}

// ...

createTheme(
  {
    styles: {
      panelShadow:
        '0px 2px 1px -1px rgb(0 0 0 / 20%), 0px 1px 1px 0px rgb(0 0 0 / 14%), 0px 1px 3px 0px rgb(0 0 0 / 12%)',
    },
	// ...
```

and then in here:

```tsx
const theme = useTheme();

return (
  <NoDataContainer
    backgroundColor={backgroundColor}
    style={{ boxShadow: theme.styles.panelShadow }}
```

apps/storefront/src/theme.tsx
```diff
@@ -39,10 +40,10 @@ function B3ThemeProvider({ children }: Props) {
       {
         palette: {
           background: {
-            default: backgroundColor,
+            default: platform === 'catalyst' ? '#FFFFFF' : backgroundColor,
           },
           primary: {
-            main: primaryColor || '#1976d2',
+            main: platform === 'catalyst' ? '#000000' : primaryColor || '#1976d2',
```
icatalina, line 46: 🤔 do we receive a color from the configuration if none is selected in the dashboard?

This change will lock the color to `black` and `white` when using catalyst, not sure if that's the best course of action here, as we'll be disabling color configuration from the dashboard.

Ideally we want to use these new defaults if no color was selected.

bc-victor:
closing as this task is back in refinement

PR #263, opened 2025-03-14
refactor: extract logout method from the Login Page

icatalina:
Jira: [B2B-2164](https://bigcommercecloud.atlassian.net/browse/B2B-2164)

## What/Why?
- Extract the logout method from the login page
- Replace custom components/callbacks with standard mui components

## Rollout/Rollback
Revert

### Before:
image

### After:
image

[B2B-2164]: https://bigcommercecloud.atlassian.net/browse/B2B-2164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/Login/index.tsx
+++ b/apps/storefront/src/pages/Login/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext, useEffect, useState } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import { B2BEvent, useB2BCallback } from '@b3/hooks';
 import { useB3Lang } from '@b3/lang';
@@ -11,24 +11,14 @@ import { useMobile } from '@/hooks';
 import { CustomStyleContext } from '@/shared/customStyleButton';
 import { defaultCreateAccountPanel } from '@/shared/customStyleButton/context/config';
 import { GlobalContext } from '@/shared/global';
-import {
-  endUserMasqueradingCompany,
-  getBCForcePasswordReset,
-  superAdminEndMasquerade,
-} from '@/shared/service/b2b';
-import { b2bLogin, bcLogin, bcLogoutLogin, customerLoginAPI } from '@/shared/service/bc';
-import {
-  clearMasqueradeCompany,
-  isLoggedInSelector,
-  useAppDispatch,
-  useAppSelector,
-} from '@/store';
+import { getBCForcePasswordReset } from '@/shared/service/b2b';
+import { b2bLogin, bcLogin, customerLoginAPI } from '@/shared/service/bc';
+import { isLoggedInSelector, useAppDispatch, useAppSelector } from '@/store';
 import { setB2BToken } from '@/store/slices/company';
 import { CustomerRole, UserTypes } from '@/types';
 import { AlertColor, LoginFlagType } from '@/types/login';
 import { b2bJumpPath, channelId, loginJump, snackbar, storeHash } from '@/utils';
 import b2bLogger from '@/utils/b3Logger';
-import { logoutSession } from '@/utils/b3logout';
 import { getCurrentCustomerInfo } from '@/utils/loginInfo';
 
 import { type PageProps } from '../PageProps';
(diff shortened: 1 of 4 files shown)


apps/storefront/src/pages/Login/useLogout.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { B2BEvent } from '@b3/hooks';
+
+import { endUserMasqueradingCompany, superAdminEndMasquerade } from '@/shared/service/b2b';
+import { bcLogoutLogin } from '@/shared/service/bc';
+import { clearMasqueradeCompany, useAppDispatch, useAppSelector } from '@/store';
+import b2bLogger from '@/utils/b3Logger';
+import { logoutSession } from '@/utils/b3logout';
+
+export const useMasquerade = () => {
+  const isAgenting = useAppSelector(({ b2bFeatures }) => b2bFeatures.masqueradeCompany.isAgenting);
```
leeBigCommerce, line 11: `isAgenting` is basically `isMasquerading`?

apps/storefront/src/pages/Login/useLogout.tsx
```diff
@@ -0,0 +1,57 @@
+import { useCallback } from 'react';
+import { B2BEvent } from '@b3/hooks';
+
+import { endUserMasqueradingCompany, superAdminEndMasquerade } from '@/shared/service/b2b';
+import { bcLogoutLogin } from '@/shared/service/bc';
+import { clearMasqueradeCompany, useAppDispatch, useAppSelector } from '@/store';
+import b2bLogger from '@/utils/b3Logger';
+import { logoutSession } from '@/utils/b3logout';
+
+export const useMasquerade = () => {
```
leeBigCommerce, line 10: `useEndMasquerade` if it's doing nothing else?

apps/storefront/src/pages/Login/useLogout.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return async () => {
+    try {
+      const { result } = (await bcLogoutLogin()).data.logout;
+
+      if (result !== 'success') {
+        return;
+      }
+
+      await Promise.all([endCompanyMasquerading(), endMasquerade()]);
```
leeBigCommerce, line 47: would we ever want to do these individually? I'm wondering what the benefit is of the split,

e.g. why not roll `endCompanyMasquerading` into `endMasquerade`?

apps/storefront/src/pages/Login/useLogout.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const useLogout = () => {
+  const endMasquerade = useMasquerade();
+  const endCompanyMasquerading = useEndCompanyMasquerading();
+
+  return async () => {
+    try {
+      const { result } = (await bcLogoutLogin()).data.logout;
+
+      if (result !== 'success') {
```
leeBigCommerce, line 43: can the promise resolve to any other state?

if not, I'd consider chaining with a `then` 🍹

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -126,7 +98,7 @@ export default function Login(props: PageProps) {
   };
 
   useEffect(() => {
-    const logout = async () => {
+    const asyncLogout = async () => {
```
leeBigCommerce, line 101: still grim, but maybe we can swap this immediate definition and invocation with an async IIFE?

```tsx
  useEffect(() => {
    (async () => {
       // stuff...
    })();
  }, [b3Lang, endMasquerade, isLoggedIn, isAgenting, searchParams, selectCompanyHierarchyId]);
```

otherwise we have an `asyncLogout` and a `logout` (which is also async)

apps/storefront/src/pages/Login/useLogout.tsx
icatalina: atm they are two separate actions, they get enabled at different places and ending is part of the logout process. Ideally, we'll store these in the session and remove the need for these calls.

apps/storefront/src/pages/Login/useLogout.tsx
icatalina: this is not the resolve state of the promise, but rather the response from the server. By looking at the code I'm assuming there's a scenario where it can fail.

leeBigCommerce (approved):

PR #265, opened 2025-03-14
fix: bug where email address is not shown for the reset password alert

icatalina:
Jira: [B2B-2164](https://bigcommercecloud.atlassian.net/browse/B2B-2164)

## What/Why?
Simplify the code and fix a bug where the email variable is not replaced correctly when rendering the reset password alert on the login page.

## Rollout/Rollback
Revert

### Before:
image

### After:
image


apps/storefront/src/pages/Login/config.ts
```diff
@@ -167,6 +167,10 @@ export const loginType: LoginTypeConfig = {
...
 };
 
-export const isLoginFlagType = (value: string | null): value is LoginFlagType => {
-  return Object.keys(loginType).includes(value as keyof typeof loginType);
+export const isLoginFlagType = (value?: string): value is LoginFlagType => {
+  if (typeof value !== 'string') {
+    return false;
+  }
+
+  return Object.keys(loginType).includes(value);
```
leeBigCommerce, line 175: I think we can remove some of the optionality if we move to a zod-esq parser

```tsx
const isLoginFlagType = (value: unknown): value is LoginFlagType =>
  typeof value === 'string' && Object.keys(loginType).includes(value);

export const parseLoginFlagType = (value: unknown): LoginFlagType | undefined =>
  isLoginFlagType(value) ? value : undefined;
```

```tsx
// usage
const loginFlag = parseLoginFlagType(searchParams.get('loginFlag'));

// no more or undefined
const tipInfo = (loginFlag: LoginFlagType, email = '') => // etc

// elsewhere
const tip = flag && tipInfo(flag, loginAccount?.emailAddress);
```

we also get back all the `have no overlap` errors should we write an incorrect branch statement

image

icatalina: Not sure how the _zod-esq parser_ helps 🤔 

```tsx
if (isLoginFlagType(flag)) {
  setLoginFlag(flag);
}
```

and

```tsx
const tip = flag && tipInfo(flag, loginAccount?.emailAddress);
```

seem to do the trick without needing the "parser"

Just bear in mind we need the `useState`, as that flag is set everywhere in the file, separate from the searchParam...

icatalina:
@leeBigCommerce ♻️ 

I did add some of the suggestions, but the parser didn't seem necessary.

apps/storefront/src/pages/Login/config.ts
leeBigCommerce: it was mostly for keeping the tighter type for the rest of the code (like in my IDE screenshot)

the normal "parse, don't validate" benefit of it no-longer being just a `string` 🍹

leeBigCommerce (approved):

PR #275, opened 2025-03-20
fix(b2b): b2b-2364  fix permission issue for junior buyer on shopping list creation

bc-amandalan:
Jira: [B2B-2364](https://bigcommercecloud.atlassian.net/browse/B2B-2364)

## What/Why?
Junior buyer does not have permissions to create shopping list on PDP.

The fix is to set `containOrEqual` to `contain` instead of using default `equal` 
My understanding is that Junior buyer's permission contains submit for approval permission, it also has other permissions.

## Rollout/Rollback
Revert

## Testing
Manual test with Junior Buyer, Senior Buyer and Admin role and the results are as expected
Junior Buyer Before: Cannot create shopping list

https://github.com/user-attachments/assets/e2ed7fef-a8f1-4031-960c-14a6b11edcdc

Junior Buyer After: Can create shopping list and the status is `draft`

https://github.com/user-attachments/assets/51244583-4155-4c23-a903-b9707bffda79

Senior Buyer behaviour is the same before and after: Can create shopping list and it's approved.

https://github.com/user-attachments/assets/fd85f69c-ed33-43ba-a360-4f115f47b522

Admin behaviour is the same before and after: Can create shopping list and it's approved.

https://github.com/user-attachments/assets/3f328662-97f4-4511-973d-362de130099c

[B2B-2364]: https://bigcommercecloud.atlassian.net/browse/B2B-2364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/utils/b3ShoppingList/b3ShoppingList.ts
+++ b/apps/storefront/src/utils/b3ShoppingList/b3ShoppingList.ts
@@ -20,6 +20,7 @@ CreateShoppingListParams) => {
 
   if (isB2BUser) {
     const submitShoppingListPermission = validatePermissionWithComparisonType({
+      containOrEqual: 'contain',
       code: b2bPermissionsMap.submitShoppingListPermission,
     });
     const selectCompanyHierarchyId =


bc-amandalan:
Hi @icatalina, would you mind to have a look at this PR and see if my understanding about the permission is correct? I see you have a refactoring in this area. Thank you so much in advance.

bc-amandalan:
Hi @leeBigCommerce, I see you are working on adding test for shopping list page, would you mind to give me some instruction on the test? Basically, before the change, junior buyer cannot create shopping list, and now it can create successfully. Thanks in advance 🙇

leeBigCommerce (approved):
right now, the code doesn't provide any nice entry points for testing this kind of behaviour

given the simplicity of the change, plus the manual testing, this is good enough for now

icatalina (approved):

PR #282, opened 2025-03-27
refactor(account): B2B-2146 Use local instead of global state for login after registration

icatalina:
Jira: [B2B-2146](https://bigcommercecloud.atlassian.net/browse/B2B-2146)

### refactor(account): B2B-2146 Use local instead of global state for login after registration

The login after registration functionality was trying to find email &
password on the global "registration" state. Since the state is quite
big and convoluted it meant looping through all the fields to find the
customer's email and password.

By explicitly exposing it we remove the dependency on global state and
make the code simpler.

### refactor(account): B2B-2146 Rename emailAddress to email and remove storeName from RegisteredContext

Rename for consistency:

- `emailAddress` -> `email`
- `pass` -> `password`

Remove `storeName` from the RegisteredContext, all we are doing is
reading it from global and saving it in local context.

## Rollout/Rollback
Revert

## Testing
Manual

[B2B-2146]: https://bigcommercecloud.atlassian.net/browse/B2B-2146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/ForgotPassword/index.tsx
+++ b/apps/storefront/src/pages/ForgotPassword/index.tsx
@@ -49,14 +49,14 @@ export function ForgotPassword({
   });
 
   const navigate = useNavigate();
-  const emailAddressReset = watch('emailAddress');
+  const emailReset = watch('email');
 
   useEffect(() => {
     if (captchaKey || !isEnabledOnStorefront) setIsCaptchaMissing(false);
   }, [captchaKey, isEnabledOnStorefront]);
 
   const handleLoginClick: SubmitHandler<LoginConfig> = async (data) => {
-    const { emailAddress } = data;
+    const { email } = data;
 
     if (isEnabledOnStorefront && !captchaKey) {
       setIsCaptchaMissing(true);
@@ -67,7 +67,7 @@ export function ForgotPassword({
       setLoading(true);
       if (isEnabledOnStorefront && captchaKey) {
         try {
-          await requestResetPassword(captchaKey, emailAddressReset);
+          await requestResetPassword(captchaKey, emailReset);
           navigate('/login?loginFlag=receivePassword');
           setLoading(false);
         } catch (e) {
@@ -76,7 +76,7 @@ export function ForgotPassword({
       }
 
       if (!isEnabledOnStorefront) {
-        await sendEmail(emailAddress);
+        await sendEmail(email);
         setLoading(false);
         navigate('/login?loginFlag=receivePassword');
       }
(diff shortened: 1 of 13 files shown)


apps/storefront/src/pages/ForgotPassword/index.tsx
```diff
@@ -49,14 +49,14 @@ export function ForgotPassword({
   });
 
   const navigate = useNavigate();
-  const emailAddressReset = watch('emailAddress');
+  const emailReset = watch('email');
```
leeBigCommerce, line 52: this keeps making me think it'll be a boolean, e.g. "yes, email was reset"

`newEmail`..?

apps/storefront/src/pages/ForgotPassword/index.tsx
```diff
@@ -169,7 +169,7 @@ export function ForgotPassword({
               <Captcha
                 siteKey={storefrontSiteKey}
                 size="normal"
-                email={emailAddressReset}
+                email={emailReset}
```
leeBigCommerce, line 172: odd that a `Captcha` needs to know about the new email

apps/storefront/src/pages/Login/config.ts
```diff
@@ -73,9 +70,9 @@ export const loginCheckout = (data: LoginConfig) => {
   ).then((response) => response.json());
 };
 
-export const sendEmail = (emailAddress: string) => {
```
leeBigCommerce, line 76: I'm not sold on the renaming of `emailAddress` to `email`, they're different things 🤷 

most of the time the shorthand of `email` is okay-ish, but this one is a bit rough. People would be forgiven for thinking `email: string` in the signature of `sendEmail` was the content of the email, rather than the recipient's address

icatalina: i think I'd call that body or content... but fair point

apps/storefront/src/pages/ForgotPassword/index.tsx
icatalina: checked, email is not required, did some refactoring while in there

apps/storefront/src/pages/ForgotPassword/index.tsx
icatalina: turns out this was only "needed" for captcha, and captcha didn't need it 🤷

leeBigCommerce (approved):

PR #290, opened 2025-04-09
refactor(shopping-lists): B2B-2131 introduce enum for shopping list status and updated all hardcoded number to use the enum

isaacchen-bc:
Jira: [B2B-2131](https://bigcommercecloud.atlassian.net/browse/B2B-2131)

## What/Why?
It's not ideal to use hardcoded numbers for the status, which could also cause status got misused. In fact, `20` is supposed to be for `deleted` for shopping lists but it somehow got misused as `rejected`. This PR is to introduce enum for shopping list status and updated all hardcoded number to use the enum.

## Rollout/Rollback
merge/revert

## Testing
Replacing hardcoded with enum only so passing ci would be sufficient


icatalina (approved):
thanks 🙇

bc-amandalan (approved):

PR #292, opened 2025-04-13
fix(shopping-lists): B2B-2131 correct the usage of shopping list status deleted and rejected

isaacchen-bc:
Jira: [B2B-2131](https://bigcommercecloud.atlassian.net/browse/B2B-2131)

## What/Why?
According to the [public docs](https://developer.bigcommerce.com/b2b-edition/apis/rest-management/shopping-list#update-a-shopping-list) and the [back-end](https://github.com/bigcommerce/b2b-edition-api/blob/0fe45ae1894bb7e0650e1be7e1e7d31255fbfc49/applications/v2/modules/shopping_list/models.py#L18), status 20 is for deleted but it has been misused as rejected in the front-end. This PR is to correct the usage for status deleted and rejected. 

Changes:

- Use rejected instead of deleted for handleUpdateShoppingList
- When the front-end receives a deleted shopping-list, it would be considered as rejected. Have checked with product that this is the lesser evil approach we are taking for now, because there is no way to tell if shopping lists with status code 20 were actually deleted or were rejected because of the previous buyer portal behaviour.
- When filter by rejected shopping lists, it would send requests to fetch both rejected and deleted shopping lists
- Removed the usage of useGetFilterShoppingListStatus in ShoppingStatus.tsx as it's unnecessary coupling and rename ShoppingStatus to ShoppingListStatusTag.tsx
- Added tests to cover the changes and introduced builder along the way

## Rollout/Rollback
merge/revert

## Testing
TBD


bc-akshatkaushik:
Hey @isaacchen-bc 
There are quite a bit of changes here and 10 commits as well, would it be possible to squash a few commits to make it a bit easier for reviews? 
For example there are 3 commits around testing

isaacchen-bc:
> Hey @isaacchen-bc There are quite a bit of changes here and 10 commits as well, would it be possible to squash a few commits to make it a bit easier for reviews? For example there are 3 commits around testing

@bc-akshatkaushik Should be better now 👍 Squashed some of the commits

apps/storefront/src/pages/ShoppingLists/ShoppingStatus.tsx
```diff
@@ -1,59 +1,57 @@
...
+  };
 
-    return newStatus;
+  // Status code 20 was previously misused as Rejected in the frontend, which is actually Deleted
+  // Now when we want to fetch rejected shopping lists, we need to fetch deleted ones as well
+  const deletedStatus = {
+    value: ShoppingListStatus.Deleted,
+    label: b3Lang('global.shoppingLists.status.rejected'),
+    color: '#7A6041',
+    textColor: '#FFFFFF',
```
bc-akshatkaushik: Do we need const here or can we just return one big static array of values? 💅 
Not too particular about this, just curious

apps/storefront/src/pages/ShoppingLists/index.test.tsx
```diff
@@ -0,0 +1,577 @@
...
+    bcId: '109',
+  },
+}));
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+it('shows approved shopping list with "approved" label', async () => {
+  const approvedStatusCode = 0;
```
bc-akshatkaushik, line 68: Do we have an enum for this (`0`) ? 💅

apps/storefront/src/pages/ShoppingLists/index.test.tsx
```diff
@@ -0,0 +1,577 @@
...
+      company: companyStateWithSuperAdminUser,
+    },
+  });
+
+  await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
+  expect(screen.getByText(/Approved/));
+});
+
+it('shows draft shopping list with "draft" label', async () => {
+  const draftStatusCode = 30;
```
bc-akshatkaushik, line 110: Same as https://github.com/bigcommerce/b2b-buyer-portal/pull/292/commits/966865d71b7e3d1cdae77037e2c767f77d051dd5#r2041361053

apps/storefront/src/pages/ShoppingLists/index.test.tsx
isaacchen-bc: We do but I intentionally use hardcoded numbers in the tests. This is because if we use enum, then let say somehow the enum got messed up again (such doing rejected: 20 again), then the tests might not be able to pick it up.

apps/storefront/src/pages/ShoppingLists/ShoppingStatus.tsx
isaacchen-bc: yeah that was my initial approach but since we have `const b3Lang = useB3Lang();` (a custom hook), we would have to use it within a component

apps/storefront/src/pages/ShoppingLists/index.test.tsx
```diff
@@ -0,0 +1,577 @@
...
+  });
+
+  await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
+  expect(screen.getByText('shopping list 1')).toBeInTheDocument();
+  expect(screen.getByText('shopping list 2')).toBeInTheDocument();
+  expect(screen.queryByRole('button', { name: /delete/i })).not.toBeInTheDocument();
+});
+
+it('allows b2b users to delete their own rejected shopping list', async () => {
+  const rejectedStatusCode = 50;
```
bc-amandalan, line 348: If my memory surfs, we introduced enum for this already. Can we use in testing, instead of having another const?

isaacchen-bc: Yeah we did introduced enum. I did this intentionally https://github.com/bigcommerce/b2b-buyer-portal/pull/292#discussion_r2041368374

apps/storefront/src/pages/ShoppingLists/index.test.tsx
bc-akshatkaushik: Not a huge concern but I think might be over-kill, we ended up here because we were using numbers directly like this test instead of Enum
I don't see it would be as easy to mess up Enums like it was for numbers so trying to catch for that might not be needed

apps/storefront/src/pages/ShoppingLists/index.test.tsx
```diff
@@ -329,7 +329,7 @@ it('hides delete button for shopping lists that are approved or ready for approv
   const companyState = buildCompanyStateWith({
     customer,
     companyInfo,
-    permissions: [{ code: 'create_shopping_list', permissionLevel: 1 }],
+    permissions: [{ code: 'submit_shopping_list_for_approval', permissionLevel: 1 }],
```
bc-akshatkaushik, line 332: 👏

apps/storefront/src/pages/ShoppingListDetails/index.test.tsx
```diff
@@ -0,0 +1,407 @@
...
+  });
+
+  await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
+
+  expect(screen.getByText('Shopping List 1')).toBeInTheDocument();
+  expect(screen.queryByText(/add to list/i)).not.toBeInTheDocument();
+});
+
+describe('when user approves a shopping list', () => {
+  it('sends a mutation to approve the shopping list', async () => {
```
bc-akshatkaushik, line 218:
```suggestion
  it('fires a request to update shopping list status to approved', async () => {
```

apps/storefront/src/pages/ShoppingListDetails/index.test.tsx
```diff
@@ -0,0 +1,407 @@
...
+    );
+    const shoppingListsUpdateMutationVariables = shoppingListsUpdateMutationBody?.variables;
+
+    expect(shoppingListsUpdateMutationVariables).toBeDefined();
+    expect(shoppingListsUpdateMutationVariables.shoppingListData).toHaveProperty('status', 0);
+  });
+});
+
+describe('when user rejects a shopping list', () => {
+  it('sends a mutation to reject the shopping list', async () => {
```
bc-akshatkaushik, line 314: Similar to https://github.com/bigcommerce/b2b-buyer-portal/pull/292/commits/966865d71b7e3d1cdae77037e2c767f77d051dd5#r2041462842

Mutation is implementation detail that could be abstracted from the test case

bc-akshatkaushik (approved):
Minor suggestions otherwise LGTM 👌

apps/storefront/src/pages/ShoppingListDetails/index.test.tsx
isaacchen-bc: updated

apps/storefront/src/pages/ShoppingListDetails/index.test.tsx
isaacchen-bc: updated

apps/storefront/src/pages/ShoppingLists/index.test.tsx
```diff
@@ -0,0 +1,577 @@
...
+
+import { B2BRequest } from '@/shared/service/request/b3Fetch';
+import { CompanyStatus, Customer, CustomerRole, LoginTypes, UserTypes } from '@/types';
+import * as utilsModule from '@/utils';
+
+import ShoppingLists from '.';
+
+const { server } = startMockServer();
+
+const buildCustomerWith = builder<Customer>(() => ({
```
RobinOng, line 21: Normally for builder, we will use faker to randomise the param rather than hardcoding it

apps/storefront/tests/storeStateBuilders/companyStateBuilder.ts
```diff
@@ -0,0 +1,52 @@
+import { PersistPartial } from 'redux-persist/es/persistReducer';
+import { builder } from 'tests/builder';
+
+import { CompanyState } from '@/store/slices/company';
+import { CompanyStatus, CustomerRole, LoginTypes, UserTypes } from '@/types';
+
+export const buildCompanyStateWith = builder<CompanyState & PersistPartial>(() => ({
```
RobinOng, line 7: Same comment as https://github.com/bigcommerce/b2b-buyer-portal/pull/292/files#r2041490394

Also, I think I would duplicate this builder in both test file rather sharing it because both pages can evolve differently and `CompanyState` can have different interfaces from each other

apps/storefront/src/pages/ShoppingLists/ShoppingListStatusTag.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  if (statusItem) {
+    return (
+      <B3Tag color={statusItem.color} textColor={statusItem.textColor}>
+        {statusItem.label}
+      </B3Tag>
+    );
+  }
+
+  return null;
```
RobinOng, line 67: Rather than doing this check here, could we move to work out the status in the parent? Then when we reach this level, we already know it's definitely one of the valid status, and then based on that render the correct label, color and text color

apps/storefront/src/pages/ShoppingLists/ShoppingListsCard.tsx
```diff
@@ -69,7 +69,13 @@ function ShoppingListsCard(props: OrderItemCardProps) {
...
+      if (
+        // Status code 20 was previously misused as Rejected in the frontend, which is actually Deleted
+        // We need to add Deleted here so that the shopping lists that were previously rejected remain the same behavior
+        status === ShoppingListStatus.Deleted ||
+        status === ShoppingListStatus.Draft ||
+        status === ShoppingListStatus.Rejected
+      )
         return false;
       return true;
     }
```
RobinOng, line 81: not 100% related to your PR, but this will be easier to read

```
const getDeletePermissions = (status: number) => {
  if (!submitShoppingListPermission) return;

  const nonDeletableStatuses = [
    ShoppingListStatus.Deleted, // Previously misused as Rejected
    ShoppingListStatus.Draft,
    ShoppingListStatus.Rejected,
  ];

  return !nonDeletableStatuses.includes(status);
};
```

apps/storefront/src/pages/ShoppingLists/index.test.tsx
isaacchen-bc: Yeah I thought about it as well but currently there is no faker in this project. So I go with hardcoded values for now.

apps/storefront/tests/storeStateBuilders/companyStateBuilder.ts
isaacchen-bc: This company state is actually for redux. As long as a component is making use of apps/storefront/src/store/slices/company.ts, it can reuse this for tests.

apps/storefront/src/pages/ShoppingLists/index.test.tsx
isaacchen-bc: If there is no strong preference, I would lean towards keeping them as they are because it's an extra layer of safety 🙇

apps/storefront/src/pages/ShoppingLists/ShoppingListStatusTag.tsx
isaacchen-bc: For this one, I think the ideal situation would be the prop `status` in this component would refer to the `status` in `ShoppingListsItemsProps` in apps/storefront/src/pages/ShoppingLists/config.ts. To do that, it would require more testing. I would suggest we do that as a separate ticket. WDYT?

apps/storefront/src/pages/ShoppingLists/ShoppingListsCard.tsx
isaacchen-bc: Yeah just updated to something similar 👍

apps/storefront/src/pages/ShoppingLists/ShoppingListStatusTag.tsx
RobinOng: Sounds good to me

apps/storefront/src/pages/ShoppingLists/index.test.tsx
isaacchen-bc: let me add a comment around there to remind ourselves to change it once we have faker in place

bc-amandalan (approved):
LGTM

apps/storefront/src/pages/ShoppingLists/ShoppingListStatusTag.tsx
isaacchen-bc: just created: https://bigcommercecloud.atlassian.net/browse/CUST-3861

RobinOng (approved):

PR #296, opened 2025-04-16
feat(carts): on cart created callback

bc-victor:
## What/Why?
Creating a new event `on-cart-created` so headless storefronts can sync cart ids easier

## Rollout/Rollback
Revert

## Testing
Created a cart and listened to the event

### DIFF
--- a/commit-validation.json
+++ b/commit-validation.json
@@ -7,6 +7,8 @@
     "account",
     "shopping-lists",
     "orders",
-    "catalyst"
+    "catalyst",
+    "carts",
+    "orders"
   ]
 }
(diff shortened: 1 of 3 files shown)


commit-validation.json
```diff
@@ -6,6 +6,7 @@
     "permissions",
     "account",
     "shopping-lists",
-    "orders"
+    "orders",
+    "catalyst"
```
isaacchen-bc, line 10: Looking at the existing entries we have, would using `carts` be better here?

bc-victor: makes sense

icatalina:
🤔 when does buyer portal create a cart that catalyst needs to sync?

PR #298, opened 2025-04-17
feat(common): adding dynamic staging url to support integration domain

bc-victor:
## What/Why?
While deploying a Catalyst in an integration store we realized the default staging domain will not work as it is different. Adding a dynamic variable to override this url while we migrate all staging stores to integration.

## Rollout/Rollback
Revert

## Testing
NA


bc-victor:
closed as no longer needed.

PR #312, opened 2025-05-09
feat(headless): Remove the need to provide a sku in ShoppingLists & Quotes

icatalina:
Jira: [B2BCAT-25](https://bigcommercecloud.atlassian.net/browse/B2BCAT-25)

## What/Why?
Figure out product SKU based on product options. This removes the need to pass an sku from catalyst or any headless implementation.

## Rollout/Rollback
Revert

## Testing
https://github.com/user-attachments/assets/f61de06f-8eb8-4bae-b2af-2da53f83b3f1


apps/storefront/src/components/addSku.tsx
```diff
@@ -0,0 +1,51 @@
...
+  data: {
+    site: {
+      productWithSelectedOptions: {
+        sku: string;
+      };
+    };
+  };
+}
+
+export const addSku = async (product: LineItem): Promise<LineItem> => {
```
leeBigCommerce, line 23: the `add` makes it read a bit like a mutation 

`withSku`? Or `getSku` and let the client code do the tiny bit of merging?

leeBigCommerce (approved):

apps/storefront/src/components/addSku.tsx
icatalina: `getSku` it is

PR #315, opened 2025-05-13
refactor(user-management): B2BCAT-22 add tests for editing and deleting in Users Management

leeBigCommerce:
Jira: [B2BCAT-22](https://bigcommercecloud.atlassian.net/browse/B2BCAT-22)

**N.B.**
- This PR should not introduce any behaviour changes, it should be purely test focussed
- Best read with "Hide whitespace"
- Some graphql queries have been given "operation names" to allow for mocking within tests, this does not change the query or the schema of the response

## What/Why?

- Add happy path test for editing a user
- Add happy path test for deleting a user
- Add happy path test for moving to the next page
- Add happy path test for moving to the previous page

## Rollout/Rollback
- Revert if needed

## Testing
- New tests added

[B2BCAT-22]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-22?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/UserManagement/index.test.tsx
+++ b/apps/storefront/src/pages/UserManagement/index.test.tsx
@@ -2,6 +2,7 @@ import { graphql, HttpResponse } from 'msw';
 import {
   buildCompanyStateWith,
   builder,
+  bulk,
   faker,
   renderWithProviders,
   screen,
@@ -11,6 +12,8 @@ import {
   within,
 } from 'tests/test-utils';
 
+import B3LayoutTip from '@/components/layout/B3LayoutTip';
+import { DynamicallyVariableProvider } from '@/shared/dynamicallyVariable';
 import { UserExtraFieldsInfoResponse, UsersResponse } from '@/shared/service/b2b/graphql/users';
 
 import UserManagement from './index';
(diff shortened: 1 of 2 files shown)


apps/storefront/src/pages/UserManagement/index.test.tsx
```diff
@@ -196,6 +199,132 @@ it('displays users with custom roles', async () => {
...
+    expect(sallyCinnamonHeading).not.toBeInTheDocument();
+
+    // once queries/mutations are changed to use real graphql variables,
+    // we can spy on the request "variables" instead of this hacky string matching
+    expect(getUsersQuerySpy).toHaveBeenCalledWith(expect.stringContaining('offset: 0'));
+  });
+});
+
 describe('when the user clicks on the "edit" button for a user', () => {
   it('opens the "Edit User" modal', async () => {
```
icatalina, line 329: feel a bit awkward about this sort of describes, they look overly specific to the point there will mostly only be one `it` per block

leeBigCommerce: I normally aim for

```tsx
describe('the context in which the assertion applies', () => {
    it('defends this business assertion which must hold true', () => {
```

(using no `describe` when there are no pre-conditions or specific contexts required by the assertion)

Meaning they don't always end up being parents of multiple `it`s.

Would you prefer we collapsed them into something like the following?

```tsx
it('opens the "Edit User" modal when the user clicks on the "edit" button for a user', () => {
```

icatalina: I don't mind keeping it the way it is, just not used to it

leeBigCommerce: let's go with it as is, if it gets in the way or doesn't feel clear enough in test results we can change it up

PR #317, opened 2025-05-13
fix(quotes): B2B-2415 getting site configs stencil request

deov31:
## What/Why?
When adding a new product to a draft quote we are seeing the price and currency always as the default currency instead of the one selected in the storefront. That happens because when we are getting the active currency we are using the bc proxy request to get the information, so the session in the proxy is not being sync properly since the change is coming from a stencil store. In stencil stores we don't have to use the proxy. Adding a validation to handle the request based on the platform will have stencil and headless stores using the correct flow.

## Rollout/Rollback
Revert PR

## Testing
https://github.com/user-attachments/assets/5b697758-d933-4133-a429-1a1b1620810f

### DIFF
--- a/apps/storefront/src/utils/b3CurrencyFormat.ts
+++ b/apps/storefront/src/utils/b3CurrencyFormat.ts
@@ -1,7 +1,7 @@
 import { store } from '@/store';
 
 import b2bLogger from './b3Logger';
-import { getActiveCurrencyInfo, getDefaultCurrencyInfo } from './currencyUtils';
+import { getActiveCurrencyInfo } from './currencyUtils';
 
 interface MoneyFormat {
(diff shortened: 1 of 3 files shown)


apps/storefront/src/utils/b3CurrencyFormat.ts
```diff
@@ -13,10 +13,7 @@ interface MoneyFormat {
 }
 
 export const currencyFormatInfo = () => {
-  const currentCurrency =
-    import.meta.env.VITE_IS_LOCAL_ENVIRONMENT === 'TRUE'
```
deov31, line 17: We don't need this validation to make a difference between local and other envs, we can use the value from the activeCurrencyInfo

bc-micah (approved):

PR #318, opened 2025-05-14
fix(headless): B2BCAT-25 Only allow products with sku or variantId to be added to quotes

icatalina:
Jira: [B2BCAT-25](https://bigcommercecloud.atlassian.net/browse/B2BCAT-25)

## What/Why?
We can only process products with a SKU or a variantId. This is a known limitation of some of the APIs. This code only allows products containing at least one of the two to be added to the quote.

## Rollout/Rollback
Revert

## Testing

https://github.com/user-attachments/assets/5e7f441a-03fd-41e6-9f88-065d4d1901a9


apps/storefront/src/components/HeadlessController/getSku.tsx
```diff
@@ -20,8 +21,17 @@ interface ProductsWithOptionSelections {
   };
 }
 
+const graphqlRequest: typeof B3Request.graphqlBC | typeof B3Request.graphqlBCProxy = ({
```
icatalina, line 24: we shouldn't use the proxy for stencil storefronts

bc-micah (approved):
Looks good.

PR #320, opened 2025-05-14
feat(common): B2B-2880 allow a way to override buyer portal event listener for sign in

bc-alvaroolvera:
Jira: [B2B-2880](https://bigcommercecloud.atlassian.net/browse/B2B-2880)

## What/Why?
When loading catalyst checkout as a guest and clicking the Sign In button it gets redirected into a stencil page, becase the buyer portal configs map an event listener into the sign in and gets tracked by the apps/storefront/src/hooks/useB3AppOpen.ts file, so in order to detach this event listener and others, we need to expose a function that could be use by B2B checkout and override this.

## Rollout/Rollback
Revert this PR

## Testing

https://github.com/user-attachments/assets/7ca8a851-6a0f-4d27-95b9-a9056c0b1a08

[B2B-2880]: https://bigcommercecloud.atlassian.net/browse/B2B-2880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/index.d.ts
+++ b/apps/storefront/src/index.d.ts
@@ -40,7 +40,7 @@ declare interface Window {
     utils: {
       openPage: (page: import('./constants').HeadlessRoute) => void;
       getRoutes: () => import('@/shared/routeList').BuyerPortalRoute[];
-      setB3ConfigEventListeners: (key: string, value: string) => void;
+      setConfig: (key: string, value: string) => void;
       quote: {
         addProductFromPage: (item: import('@/utils').LineItems) => void;
         addProductsFromCart: () => Promise<void>;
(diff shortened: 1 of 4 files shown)


apps/storefront/src/index.d.ts
```diff
@@ -40,6 +40,7 @@ declare interface Window {
     utils: {
       openPage: (page: import('./constants').HeadlessRoute) => void;
       getRoutes: () => import('@/shared/routeList').BuyerPortalRoute[];
+      setB3ConfigEventListeners: (key: string, value: string) => void;
```
bc-micah, line 43:
```suggestion
      setConfig: (key: string, value: string) => void;
```

we shouldn't expose `B3` as that is a reference to `bundleb2b` and we don't want that in our public api.

```javascript
b2b.utils.setConfig('dom.registerElement')
```

bc-micah (commented):
Naming changes for public api, otherwise looks good

PR #337, opened 2025-05-23
fix(shopping-lists): B2B-2559 fixed an issue where error alert and success alert are both triggered when product without required variant is added to shopping list

isaacchen-bc:
Jira: [B2B-2559](https://bigcommercecloud.atlassian.net/browse/B2B-2559)

## What/Why?
Fixed an issue where error alert and success alert are both triggered when product without required variant is added to shopping list.

The issue happened because the whether the success alert is triggered or not does not depend on whether the validation on product option is passed or not. 

## Rollout/Rollback
merge/revert.

## Testing
Before:

https://github.com/user-attachments/assets/d2f0a736-5771-4883-9178-61ab2e70923e

After: 

https://github.com/user-attachments/assets/ef3e445b-5d11-4f75-8305-b56caf917c8c

[B2B-2559]: https://bigcommercecloud.atlassian.net/browse/B2B-2559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/components/HeadlessController/index.tsx
+++ b/apps/storefront/src/components/HeadlessController/index.tsx
@@ -31,6 +31,7 @@ import { endMasquerade, startMasquerade } from '@/utils/masquerade';
 import { globalSnackbar } from '@/utils/b3Tip';
 
 import { getSku } from './getSku';
+import { ValidationError } from '@/utils';
 
 export interface FormattedQuoteItem
   extends Omit<QuoteItem['node'], 'optionList' | 'calculatedValue' | 'productsSearch'> {
@@ -257,7 +258,11 @@ export default function HeadlessController({ setOpenPage }: HeadlessControllerPr
               customerGroupId: customerRef.current.customerGroupId,
             })
               .then(() => displayAddedToShoppingListAlert(shoppingListId.toString()))
-              .catch(({ message }) => {
+              .catch((error) => {
+                const message =
+                  error instanceof ValidationError
+                    ? error.message
+                    : 'Something went wrong. Please try again.';
                 globalSnackbar.error(message, {
                   isClose: true,
                 });
--- a/apps/storefront/src/components/HeadlessController/index.tsx
+++ b/apps/storefront/src/components/HeadlessController/index.tsx
@@ -28,6 +28,7 @@ import { LineItem } from '@/utils/b3Product/b3Product';
 import createShoppingList from '@/utils/b3ShoppingList/b3ShoppingList';
 import { getCurrentCustomerInfo } from '@/utils/loginInfo';
 import { endMasquerade, startMasquerade } from '@/utils/masquerade';
+import { globalSnackbar } from '@/utils/b3Tip';
 
 import { getSku } from './getSku';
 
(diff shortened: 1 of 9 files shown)


apps/storefront/src/components/HeadlessController/index.tsx
```diff
@@ -254,7 +258,9 @@ export default function HeadlessController({ setOpenPage }: HeadlessControllerPr
               items: transformOptionSelectionsToAttributes(items),
               isB2BUser: isB2BUserRef.current,
               customerGroupId: customerRef.current.customerGroupId,
-            }).then(() => displayAddedToShoppingListAlert(shoppingListId.toString())),
+            })
+              .then(() => displayAddedToShoppingListAlert(shoppingListId.toString()))
+              .catch(addProductsToShoppingListErrorHandler),
```
isaacchen-bc, line 263: Can't really add tests to this because `addProducts` is just basically a function pull from `pages/PDP` with some modification that tailors to headless. 

Since it's quite straight forward and also the exact same logic is already covered by the newly added tests, will skip this one for now.

apps/storefront/src/pages/PDP/index.test.tsx
```diff
@@ -0,0 +1,568 @@
...
+    await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
+
+    expect(screen.getByText('Add to shopping list')).toBeInTheDocument();
+
+    await userEvent.click(screen.getByText('Test Shopping List 1'));
+    await userEvent.click(screen.getByText('OK'));
+
+    expect(globalSnackbar.error).toHaveBeenCalledWith(
+      'Something went wrong. Please try again.',
+      expect.anything(),
```
bc-akshatkaushik, line 565: [Suggestion]
It seems the second argument is always the same, maybe we can replace with what is the expected value
```suggestion
      {
        isClose: true,
      },
```

apps/storefront/src/pages/PDP/index.test.tsx
```diff
@@ -0,0 +1,568 @@
...
+
+    expect(screen.getByText('Add to shopping list')).toBeInTheDocument();
+
+    await userEvent.click(screen.getByText('Test Shopping List 1'));
+    await userEvent.click(screen.getByText('OK'));
+
+    expect(globalSnackbar.error).not.toHaveBeenCalled();
+    expect(globalSnackbar.success).toHaveBeenCalledWith(
+      'Products were added to your shopping list',
+      expect.anything(),
```
bc-akshatkaushik, line 407: Same as https://github.com/bigcommerce/b2b-buyer-portal/pull/337/files#r2106381690

bc-akshatkaushik (approved):
Minor suggestion otherwise LGTM

apps/storefront/src/pages/PDP/index.test.tsx
isaacchen-bc: updated

apps/storefront/src/pages/PDP/index.test.tsx
isaacchen-bc: This one is bit different, the second argument is a lot more complicated as you can see in apps/storefront/src/pages/PDP/useAddedToShoppingListAlert.tsx. Personally I don't think it's necessary to check the jsx thing in the second argument as we mainly just care about the message. So, I am leaning towards to keeping it as expect.anything().

PR #338, opened 2025-05-23
test(common): B2BCAT-36 Improve store mock + share local and global redux store

icatalina:
Jira: [B2BCAT-36](https://bigcommercecloud.atlassian.net/browse/B2BCAT-36)

## What/Why?
- Ensure the global store is restored between test runs to prevent state leak.

- Ensure the store setup as part of renderWithProviders is the same one consumers take from the global import. This is inline with the behaviour in production. The correct solution is to only use the store provided through context, but that requires some heavy refactoring in some cases.

- Add some test covering these two cases.

- Disable `parallelism: 2`.  This is really of no use in the test task. Instead of splitting the load, it is running the same task twice.

- Restrict imports of `lodash` and `msw`, prefer `lodash-es` and `test-utils`

- Allow importing dev dependencies from `__mocks__` files.

## Rollout/Rollback
Revert

## Testing
These changes only affect tests/ci, a successful build should prove they work.

### DIFF
--- a/apps/storefront/tests/setup-test-environment.ts
+++ b/apps/storefront/tests/setup-test-environment.ts
@@ -4,6 +4,8 @@ import { Environment } from '@/types';
 
 import '@testing-library/jest-dom/vitest';
 
+vi.mock('@/store');
+
 window.B3 = {
   setting: {
     channel_id: 1,
--- a/apps/storefront/tests/test-utils.tsx
+++ b/apps/storefront/tests/test-utils.tsx
@@ -6,8 +6,8 @@ import { render, RenderOptions } from '@testing-library/react';
 import { userEvent } from '@testing-library/user-event';
 import { Mock } from 'vitest';
 
-import { AppStore, RootState, setTimeFormat, setupStore, store as storeSingleton } from '@/store';
-import { setPermissionModules } from '@/store/slices/company';
+import { AppStore, RootState, setupStore } from '@/store';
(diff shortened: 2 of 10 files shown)


apps/storefront/tests/test-utils.tsx
```diff
@@ -33,20 +33,9 @@ export const renderWithProviders = (
...
-  // and NOT to the store passed in to the Provider context below
-  // until this is fixed, we need manually sync the time format to the store singleton
-  if (preloadedState.storeInfo?.timeFormat) {
-    storeSingleton.dispatch(setTimeFormat(preloadedState.storeInfo.timeFormat));
-  }
-
-  // As above, `validatePermissionWithComparisonType` reaches to the store singleton
-  if (preloadedState.company?.permissions) {
-    storeSingleton.dispatch(setPermissionModules(preloadedState.company.permissions));
-  }
```
icatalina, line 46: Since the global/local store is now shared, this is no longer necessary.

apps/storefront/tests/setup-test-environment.ts
```diff
@@ -4,6 +4,8 @@ import { Environment } from '@/types';
 
 import '@testing-library/jest-dom/vitest';
 
+vi.mock('@/store');
```
icatalina, line 7: Always mock the store by default, this ensures it is reset between test runs.

PR #339, opened 2025-05-23
POC - fix(headless): B2B-2974 masquerade callbacks

bc-victor:
POC

Jira: [B2B-2974](https://bigcommercecloud.atlassian.net/browse/B2B-2974)

## What/Why?
Adding start and end events for masquerade feature so merchants can introduce side effects to the application such as cart manipulation.


bc-victor:
Closing since we won't be adding this feature

PR #344, opened 2025-05-27
fix(quotes): B2B-2882 rendering modifier values in quotes

deov31:
## What/Why?
In quotes we are currently not rendering the modifier values of a product because we are relying on the optionLabel key to display the values. The modifiers are not being sent from the api with optionLabel, so instead we need to use the optionName that is part of the response for both modifiers and variants.

## Rollout/Rollback
Revert PR

## Testing
Before:
image

After:
image


deov31:
closing this PR since it's won't be needed

PR #349, opened 2025-05-29
Fix silent fail on quick order add to cart

thebigrick:

## What/Why?
The quick order was silently failing under unexpected circumstances.
Try to define a minimum quantity for a product ad use the quick order.
The expected result is to receive an error. The actual result is a silent failure.

## Testing
- Create a product with a minimum quantity
- Use the quick order to add this product using a quantity below the minimum
- Expect to see an error


bc-victor (approved):

thebigrick:
Hello @icatalina ,
forgot to close the issue. I provided a PR fixing it about 1 month ago.

PR #350, opened 2025-05-29
feat(dashboard): B2B-2993 challenge masquerading for cart removal

bc-victor:
Jira: [B2B-2993](https://bigcommercecloud.atlassian.net/browse/B2B-2993)

## What/Why?
Delete cart on masquerading as it is a possible security risk, read ticket.

## Rollout/Rollback
Revert

## Testing
To be added

### DIFF
--- a/apps/storefront/src/pages/Dashboard/index.tsx
+++ b/apps/storefront/src/pages/Dashboard/index.tsx
@@ -2,8 +2,10 @@ import { useEffect, useState } from 'react';
 import { useLocation } from 'react-router-dom';
 import { useB3Lang } from '@b3/lang';
 import { Box } from '@mui/material';
+import Cookies from 'js-cookie';
 
 import { usePageMask } from '@/components';
+import B3Dialog from '@/components/B3Dialog';
 import B3FilterSearch from '@/components/filter/B3FilterSearch';
 import B3Spin from '@/components/spin/B3Spin';
 import { B3PaginationTable, GetRequestList } from '@/components/table/B3PaginationTable';
@@ -11,15 +13,13 @@ import { TableColumnItem } from '@/components/table/B3Table';
 import { useSort } from '@/hooks';
 import { PageProps } from '@/pages/PageProps';
 import { superAdminCompanies } from '@/shared/service/b2b';
+import { deleteCart } from '@/shared/service/bc';
 import { setCartNumber, useAppSelector, useAppStore } from '@/store';
 import { endMasquerade, startMasquerade } from '@/utils/masquerade';
 
 import { DashboardCard } from './components/DashboardCard';
 import { ActionMenuCell } from './ActionMenuCell';
 import { CompanyNameCell } from './CompanyNameCell';
-import B3Dialog from '@/components/B3Dialog';
-import Cookies from 'js-cookie';
-import { deleteCart } from '@/shared/service/bc';
 
 interface ListItem {
   [key: string]: string;
@@ -138,7 +138,7 @@ function Dashboard(props: PageProps) {
   };
 
   const handleModalConfirm = () => {
-    const cartId = Cookies.get('cartId')
+    const cartId = Cookies.get('cartId');
 
     startActing(Number(tempCompanyId))
       .then(() => deleteCart(cartId as string))
@@ -147,7 +147,7 @@ function Dashboard(props: PageProps) {
         store.dispatch(setCartNumber(0));
         handleClose();
       });
-  }
+  };
 
   const columnItems: TableColumnItem<ListItem>[] = [
     {
--- a/apps/storefront/src/pages/Dashboard/index.tsx
+++ b/apps/storefront/src/pages/Dashboard/index.tsx
@@ -19,6 +19,7 @@ import { ActionMenuCell } from './ActionMenuCell';
 import { CompanyNameCell } from './CompanyNameCell';
 import B3Dialog from '@/components/B3Dialog';
 import Cookies from 'js-cookie';
+import { deleteCart } from '@/shared/service/bc';
 
 interface ListItem {
   [key: string]: string;
@@ -249,7 +250,9 @@ function Dashboard(props: PageProps) {
         loading={isRequestLoading}
         handleLeftClick={handleClose}
         handRightClick={() => {
+          const cartId = Cookies.get('cartId')
           startActing(Number(tempCompanyId))
+            .then(() => deleteCart(cartId as string))
             .then(() => Cookies.remove('cartId'))
             .finally(() => {
               store.dispatch(setCartNumber(0));
--- a/apps/storefront/src/pages/Dashboard/index.tsx
+++ b/apps/storefront/src/pages/Dashboard/index.tsx
@@ -11,12 +11,14 @@ import { TableColumnItem } from '@/components/table/B3Table';
 import { useSort } from '@/hooks';
 import { PageProps } from '@/pages/PageProps';
 import { superAdminCompanies } from '@/shared/service/b2b';
-import { useAppSelector, useAppStore } from '@/store';
+import { setCartNumber, useAppSelector, useAppStore } from '@/store';
 import { endMasquerade, startMasquerade } from '@/utils/masquerade';
 
 import { DashboardCard } from './components/DashboardCard';
 import { ActionMenuCell } from './ActionMenuCell';
 import { CompanyNameCell } from './CompanyNameCell';
+import B3Dialog from '@/components/B3Dialog';
+import Cookies from 'js-cookie';
 
 interface ListItem {
   [key: string]: string;
@@ -52,6 +54,8 @@ function Dashboard(props: PageProps) {
   const b3Lang = useB3Lang();
 
   const [isRequestLoading, setIsRequestLoading] = useState(false);
+  const [tempCompanyId, setTempCompanyId] = useState(1);
+  const [openModal, setOpenModal] = useState(false);
 
   const [filterData, setFilterData] = useState<ListItem>({
     q: '',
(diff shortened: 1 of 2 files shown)


apps/storefront/src/pages/Dashboard/components/ConfirmMasqueradeDialog.tsx
```diff
@@ -0,0 +1,64 @@
...
+import B3Dialog from '@/components/B3Dialog';
+
+interface ConfirmMasqueradeDialogProps {
+  isOpen: boolean;
+  isRequestLoading: boolean;
+  handleClose: () => void;
+  handleConfirm: () => void;
+}
+
+export function ConfirmMasqueradeDialog({
```
icatalina, line 13: This looks great. Thanks 🙇

packages/lang/locales/en.json
```diff
@@ -133,6 +133,11 @@
   "dashboard.selected": "Selected",
   "dashboard.masqueradeAction": "Masquerade",
   "dashboard.endMasqueradeAction": "End Masquerade",
+  "dashboard.startMasqueradeModal.title": "Start Masquerading",
+  "dashboard.startMasqueradeModal.message": "Starting masquerade will refresh your shopping cart. Do you want to continue?",
```
icatalina, line 137: not sure if "refresh" is the correct word here. When I see that I imagine same products but new information. @bc-sylvialin what do you think?

apps/storefront/src/pages/Dashboard/index.tsx
```diff
@@ -128,6 +141,18 @@ function Dashboard(props: PageProps) {
     });
   };
 
+  const onStartMasquerade = (companyId: number) => {
+    try {
+      const cartEntityId = z.string().parse(Cookies.get('cartId'));
```
icatalina, line 146: I feel like I explained myself poorly here. `Cookies.get` is already returning `string|undefined`, no need to parse that. I was expecting the value to be more volatile (`string|number|object|undefined|null`... which might grant the parsing)

apps/storefront/src/pages/Dashboard/index.tsx
```diff
@@ -128,6 +141,18 @@ function Dashboard(props: PageProps) {
...
+      setCartEntityId(cartEntityId);
```
icatalina, line 147: After some thought, we might want to read the cartId directly in the startActing method, this way we don't have to keep track of two variables.

```tsx
const cartEntityId = Cookies.get('cartId');

if (cartEntityId) {
  await deleteCart({ deleteCartInput: { cartEntityId } });
  Cookies.remove('cartId');
  store.dispatch(setCartNumber(0));
  // setCartEntityId(''); this should probably go in the `handleConfirm`, as it is only specific to the modal being open.
}
```

This separates the action of acting from the action of confirming.

`startActing` will always remove the existing cart, we only change whether we want confirmation or not.

apps/storefront/src/pages/Dashboard/index.tsx
```diff
@@ -52,6 +56,8 @@ function Dashboard(props: PageProps) {
   const b3Lang = useB3Lang();
 
   const [isRequestLoading, setIsRequestLoading] = useState(false);
+  const [tempCompanyId, setTempCompanyId] = useState(1);
+  const [cartEntityId, setCartEntityId] = useState<string>('');
```
icatalina, line 60: reading `cartId` in both places removes the need for `setCartEntityId`, we do have to configur `tempCompanyId` here. Can we give it a more explicit name:


```tsx
const [confirmMasquerade, setConfirmMasquerade] = useState<{ companyId: number } | undefined>(undefined);
```

apps/storefront/src/pages/Dashboard/index.tsx
```diff
@@ -224,6 +249,12 @@ function Dashboard(props: PageProps) {
           }}
         />
       </Box>
+      <ConfirmMasqueradeDialog
+        isOpen={Boolean(cartEntityId)}
+        isRequestLoading={isRequestLoading}
+        handleClose={() => setCartEntityId('')}
+        handleConfirm={() => startActing(tempCompanyId)}
```
icatalina, line 256:
```tsx
handleConfirm={
  async () => {
    if (confirmMasquerade) { // this `if` is only here to please TS 🤷 
      await startActing(confirmMasquerade.companyId);
      setConfirmMasquerade(undefined);
    }
  }
}
```

apps/storefront/src/pages/Dashboard/index.test.tsx
```diff
@@ -389,6 +390,15 @@ describe('when the user is associated with a company', () => {
...
       }),
     );
+    const deleteCartReturn = vi.fn().mockReturnValue({
+      data: {
+        cart: {
+          deleteCart: {
+            deletedCartEntityId: '1',
+          },
+        },
+      },
```
icatalina, line 400: if we don't use the response, which we don't seem to be using, we shouldn't include it in the mock.

We can ensure the request was correct by using `when`:

```tsx
when(deleteCartReturn)
  .calledWith(stringContainingAll('deletedCartEntityId: 1')) 
  .thenReturn({});
```

makes sure that setting the wrong `cartId` makes the test fail.

apps/storefront/src/pages/Dashboard/index.test.tsx
```diff
@@ -430,6 +443,7 @@ describe('when the user is associated with a company', () => {
 
     expect(beginMasquerade).toHaveBeenLastCalledWith(expect.stringContaining('companyId: 123'));
     expect(getAgentInfo).toHaveBeenLastCalledWith(expect.stringContaining('customerId: 789'));
+    expect(deleteCartReturn).toHaveBeenCalled();
```
icatalina, line 446: if we include the response only on `when`, and the test fails correctly with the wrong `id`, we don't need to include this expect here.

apps/storefront/src/pages/Dashboard/index.test.tsx
```diff
@@ -400,8 +410,10 @@ describe('when the user is associated with a company', () => {
       ),
       graphql.mutation('BeginMasquerade', ({ query }) => HttpResponse.json(beginMasquerade(query))),
       graphql.query('AgentInfo', ({ query }) => HttpResponse.json(getAgentInfo(query))),
+      graphql.mutation('deleteCart', ({ query }) => HttpResponse.json(deleteCartReturn(query))),
     );
 
+    Cookies.set('cartId', '1');
```
icatalina, line 416: can you check if the cookie stays set in the next test? it is a pretty common issue that things like this will spill into the next test. If so, make sure you add some reset in a `beforeEach` block or something on those lines.

icatalina (commented):
Hey @bc-victor!

I noticed this PR only changes the desktop view, do we want to show the same modal on mobile?

apps/storefront/src/pages/Dashboard/index.test.tsx
bc-victor: After looking into this the cookie persists so I added in a beforeEach block

bc-victor:
Screenshot 2025-06-03 at 5 43 48 p m
 
Mobile test

apps/storefront/src/pages/Dashboard/index.tsx
```diff
@@ -128,6 +143,24 @@ function Dashboard(props: PageProps) {
     });
   };
 
+  const onMasquerade = async ({
+    companyId,
+    type,
+  }: {
+    companyId: number;
+    type: 'start' | 'end';
+  }) => {
```
icatalina, line 152: 🤔 I'm not sure if we'd want to combine the two actions in `onMasquerade` and loosen the state's types, or go for a more strict version like:

```tsx
type ConfirmState = {
  type: 'start',
  companyId: number,
} | {
  type: 'end',
};

const [confirmMasquerade, setConfirmMasquerade] = useState<ConfirmState | undefined>(undefined);


const onStartMasquerade = (companyId: number) => {
  const cartId = Cookies.get('cartId');

  if (cartId) {
    setConfirmMasquerade({ type: 'start', companyId });
  } else {
    startActing(companyId);
  }
};

const onEndMasquerade = () => {
  const cartId = Cookies.get('cartId');

  if (cartId) {
    setConfirmMasquerade({ type: 'end' });
  } else {
    endActing(companyId);
  }
};
```


```tsx
onClick={() => onStartMasquerade(Number(companyId))}
// ...
onClick={() => onEndMasquerade())}
```

The current solution requires superfluous data to be sent to the `endMasquerade` action (eg. companyId)

apps/storefront/src/pages/Dashboard/index.tsx
```diff
@@ -128,6 +143,24 @@ function Dashboard(props: PageProps) {
...
+    const cartId = Cookies.get('cartId');
+
+    if (cartId && cartNumber > 0) {
```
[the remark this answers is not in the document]
bc-victor: cartNumber contains the count of the cart items. if there is a cart id set but it has no items, we do not need to challenge per the requirement

apps/storefront/src/pages/Dashboard/index.tsx
leeBigCommerce: sounds like a case for an `asserts` then?

```tsx
function assertDefined<T>(value?: T): asserts value is T {
  if (value === undefined) {
    throw new Error("Expected value to be defined, but got undefined");
  }
}
```

useful when the code forces us to explain to Typescript that a scenario is technically impossible e.g.
this modal being open and yet `confirmMasquerade` equal to `undefined`

definitely a technique of last resort, but it'd take some epic refactoring of this Dialogue component for Typescript to work out this scenario is impossible on it's own 🍹

bc-victor:
not sure why the mock fails when in the debugger the setOpenPage fn gets called :/ will check tomorrow again

Screenshot 2025-06-05 at 6 42 31 p m

icatalina (approved):

bc-victor:
After discussing with product, we'll only display the modal if the product count is bigger than zero, however, we'll always delete the cart from the cookie and API in order to avoid issues with attached consignments.

PR #358, opened 2025-06-04
test(shopping-lists): B2BCAT-38 headless controller tests for sls

bc-victor:
Jira: [B2BCAT-38](https://bigcommercecloud.atlassian.net/browse/B2BCAT-38)

## What/Why?
Initial tests for headless controller specifically getLists fn.

## Rollout/Rollback
Revert

## Testing
Test suite passes in the CI

[B2BCAT-38]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-38?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/components/HeadlessController/index.test.tsx
+++ b/apps/storefront/src/components/HeadlessController/index.test.tsx
@@ -1,112 +1,117 @@
-import { buildCompanyStateWith, builder, bulk, faker, graphql, HttpResponse, renderWithProviders, startMockServer, stringContainingAll } from 'tests/test-utils';
+import {
+  buildCompanyStateWith,
+  builder,
+  bulk,
+  faker,
+  graphql,
+  HttpResponse,
+  renderWithProviders,
+  startMockServer,
+  stringContainingAll,
+} from 'tests/test-utils';
 import { when } from 'vitest-when';
 
-import HeadlessController from '.';
 import { ShoppingListsItemsProps } from '@/pages/ShoppingLists/config';
+
+import HeadlessController from '.';
 
 const { server } = startMockServer();
 
-const buildShoppingListsNodeWith = builder<{node: ShoppingListsItemsProps}>(() => ({
-    node: {
-        id: faker.number.int({ max: 1000 }),
-        name: faker.word.words({ count: 1 }),
-        description: faker.word.words(),
-        status: faker.number.int({ multipleOf: 10, max: 50 }),
-        channelId: faker.number.int(),
-        customerInfo: {
-            firstName: faker.person.firstName(),
-            lastName: faker.person.lastName(),
-            userId: faker.number.int(),
-            email: faker.internet.email(),
-            role: faker.word.words()
-        },
-        products: {
-            totalCount: faker.number.int(),
-        },
-        updatedAt: faker.date.anytime().toString(),
-        approvedFlag: faker.datatype.boolean(),
-        isOwner: faker.datatype.boolean(),
-        companyInfo: {
-            companyId: faker.string.numeric(),
-            companyName: faker.company.name(),
-            companyAddress: faker.location.streetAddress(),
-            companyCountry: faker.location.country(),
-            companyState: faker.location.state(),
-            companyCity: faker.location.city(),
-            companyZipCode: faker.location.zipCode(),
-            phoneNumber: faker.string.numeric(),
-            bcId: faker.string.numeric(),
-        }
-    }
-}))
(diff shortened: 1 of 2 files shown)


apps/storefront/src/components/HeadlessController/index.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+            pageInfo: {
+              hasNextPage: false,
+              hasPreviousPage: false,
+            },
+            edges: bulk(buildShoppingListsNodeWith, 'WHATEVER_VALUES').times(1),
+          },
+        },
+    });
+
+    // not sure how this is used
```
bc-victor, line 95: @icatalina I'm still not sure how should I use `when` here

apps/storefront/src/components/HeadlessController/index.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    when(getShoppingLists)
+      .calledWith(stringContainingAll('offset: 0', 'first:50'))
+      .thenReturn({});
+
+    server.use(
+      graphql.operation(({ query }) => HttpResponse.json(getShoppingLists(query))),
```
bc-victor, line 101: using graphql.operation because when using graphql.query('customerShoppingLists', ...) the request failed

icatalina: you couldn't use query because `customerShoppingLists` is not a name, is the node. We need to add a name above the node:

image

apps/storefront/src/components/HeadlessController/index.test.tsx
icatalina: You use `when` to ensure your mock only responds under the correct conditions.

For example, say you're testing loading an order. You expect a query like:

```graphql
query GetOrder($orderId) {
  order(id: $orderId) {
    id,
    products,
    email,
  }
}
```

Let’s say `$orderId` is `42`.

If you mock it like this:

```tsx
const getOrder = vi.fn().mockReturnValue({
  id: 42,
  email: 'foo@email.com',
  products
});
```

Then **any** call to `getOrder` will return that order, regardless of the input. To ensure the request is correct, you’d need to assert on the call itself:

```tsx
expect(getOrder).toHaveBeenCalledWith(expect.stringContaining('id: 42'));
```

This works, but it's not ideal. We generally want to test outcomes, not internal calls. That’s where `when` comes in:

```tsx
when(getOrder)
  .calledWith(expect.stringContaining('id: 42'))
  .mockReturnValue({ id: 42, email: 'foo@email.com', products });
```

Now, your mock only responds when the request includes the correct ID. This lets you assert on the **output** of your logic, rather than how mocks were used:

```tsx
expect(controller.getList()).toContainEqual(expect.objectContaining({ id: 42 }));
```

No need to verify `getOrder` was called a certain way, you can stay focused on what your code *returns*.

apps/storefront/src/components/HeadlessController/index.test.tsx
bc-victor: makes sense, thanks! will add

apps/storefront/src/components/HeadlessController/index.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+        HttpResponse.json(getShoppingLists(query)),
+      ),
+    );
+
+    const mockSetOpenPage = vi.fn();
+    renderWithProviders(<HeadlessController setOpenPage={mockSetOpenPage} />);
+    const data = await window.b2b.utils.shoppingList.getLists();
+
+    expect(data).toBeDefined();
+    expect(data).toEqual(
```
icatalina, line 142: if not defined, the `toEqual` will fail.

```suggestion
    expect(data).toEqual(
```

apps/storefront/src/components/HeadlessController/index.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+        HttpResponse.json(getShoppingLists(query)),
+      ),
+    );
+
+    const mockSetOpenPage = vi.fn();
+    renderWithProviders(<HeadlessController setOpenPage={mockSetOpenPage} />, {
+      preloadedState: {
+        company: buildCompanyStateWith({
+          customer: {
+            role: 3,
```
icatalina, line 89: Use the `CustomerRole.SUPER_ADMIN` variable to make it more readable

icatalina (approved):

PR #363, opened 2025-06-10
refactor(order): B2BCAT-39 Use react-query for data fetching, simplify MyOrders component

icatalina:
Jira: [B2BCAT-39](https://bigcommercecloud.atlassian.net/browse/B2BCAT-39)

## What/Why?
- refactor(order): B2BCAT-39 Use react-query for improved data fetching and state management

	This removes the need to cache data, react-query diffs the query params and ensures new data is fetched whenever the params change.

- refactor(order): B2BCAT-39 Simplify order data fetching and remove unused variables
- refactor(orders): B2BCAT-39 Duplicate orders graphql fetcher
- refactor(order): B2BCAT-39 Update filter initialization for B2B and customer orders

## Rollout/Rollback
Revert

## Testing

https://github.com/user-attachments/assets/b0d8a65b-997a-49f2-93a2-ef33195d3136

[B2BCAT-39]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-39?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/pages/MyOrders/index.test.tsx
```diff
@@ -307,6 +307,46 @@ describe('when a personal customer', () => {
...
+      expect(createdOnHeader).toHaveAttribute('aria-sort', 'descending');
+
+      when(getOrders)
+        .calledWith(stringContainingAll('orderBy: "bcOrderId"'))
+        .thenReturn(
+          buildGetCustomerOrdersWith({
+            data: {
+              customerOrders: {
+                totalCount: 1,
+                edges: [buildCustomerOrderNodeWith({ node: { orderId: '66996' } })],
```
leeBigCommerce, line 332: might be a bit more intuitive to have two order nodes returned by each mock (but in different respective orders)

we would be mocking out BE responsibilities, but it gives readers a clearer sense of user expectations 🍹 

also, are people using order by ID as a stand in for ordering by created date? Or does it solve for some other use case?

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -94,55 +96,88 @@ function useData() {
...
-    isB2BUser,
-    orderSubViewPermission,
-    selectCompanyHierarchyId,
-    isEnabledCompanyHierarchy,
-    currentCompanyId,
-  } = useData();
+  const { role, isAgenting, companyId, isB2BUser, isEnabledCompanyHierarchy, selectedCompanyId } =
+    useData();
+
+  const [pagination, setPagination] = useState<{ offset: number; first: number }>({
```
leeBigCommerce, line 133: typescript will infer these types correctly from the argument alone 🍹

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -94,55 +96,88 @@ function useData() {
...
+  const [orderBy, setOrderBy] = useState<OrderBy>({
+    key: 'orderId',
+    dir: 'desc',
+  });
+
+  const handleSetOrderBy = (key: string) => {
+    setOrderBy((prev) => {
+      if (!isSortKey(key)) {
+        return prev;
+      }
```
leeBigCommerce, line 152: is there a use case where this branch would be hit?

If we're just doing it because of the lossy `string` type, we could use an `asserts` instead 🍹 

```tsx
  const handleSetOrderBy = (key: string) => {
    assertIsSortKey(key);
    // etc etc
```

apps/storefront/src/pages/order/config.ts
```diff
@@ -137,13 +117,31 @@ export const getFilterMoreData = (
   return filterCurrentMoreList;
 };
 
-export const getInitFilter = (
+export const getCustomerInitFilter = (): Partial<FilterSearchProps> => {
```
leeBigCommerce, line 120: does this need to be a function?

apps/storefront/src/pages/order/orders.ts
```diff
@@ -0,0 +1,149 @@
+import B3Request from '@/shared/service/request/b3Fetch';
+import { OrderStatusItem } from '@/types';
+import { convertArrayToGraphql } from '@/utils';
+
+export type CustomerOrderNode = {
+  node: {
+    orderId?: string;
```
leeBigCommerce, line 7: surprised to see the id as optional
as well as `money` and `totalIncTax` 🤔

apps/storefront/src/pages/order/orders.ts
```diff
@@ -0,0 +1,149 @@
...
+export interface GetCompanyOrders {
+  data: {
+    allOrders: {
+      totalCount: number;
+      edges: Array<CustomerOrderNode>;
+    };
+  };
+}
+
+const allOrders = (data: CustomFieldItems, fn: 'allOrders' | 'customerOrders') => `
```
leeBigCommerce, line 53: is the plan to turn these into regular queries (i.e. not pushing variables through string literals) in a different PR?

or maybe they're not currently getting in the way?

leeBigCommerce (approved):

apps/storefront/src/pages/order/orders.ts
icatalina: don't know why, but they are setup as optional in the schema.

```
  money?: Maybe<Scalars['JSONString']['output']>;
  orderId?: Maybe<Scalars['String']['output']>;
  totalIncTax?: Maybe<Scalars['Float']['output']>;
```

we can remove the optionality in the new schema, but it is there in the current one 🤷

apps/storefront/src/pages/MyOrders/index.test.tsx
icatalina: > might be a bit more intuitive to have two order nodes returned by each mock (but in different respective orders)

can do.

> also, are people using order by ID as a stand in for ordering by created date? Or does it solve for some other use case?

I don't know really, we offer both, you can sort by `orderId`, and `createdAt` 🤷

icatalina:
Hey @jp06 ! Thanks for the kind words. I hope the rebases aren't too bad with all the changes. Happy to take any feedback you might have. 🙇

PR #368, opened 2025-06-13
feat(common): B2B-3009 add translation for rejected company checkout error message

bc-amandalan:
Jira: [B2B-3010](https://bigcommercecloud.atlassian.net/browse/B2B-3010)

## What/Why?
We need to ensure it’s added to the translation service, allowing merchants to customize the wording. 

Also need to update the translation service

## Rollout/Rollback
Revert

## Testing
Before:
Screenshot 2025-06-17 at 2 42 52 PM

After:
Updating the value in the template and upload the template on merchant dashboard, the error message can be customised:
Screenshot 2025-06-17 at 2 58 03 PM

https://github.com/user-attachments/assets/2099384a-90f0-4442-b082-c14836e374bf

[B2B-3010]: https://bigcommercecloud.atlassian.net/browse/B2B-3010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/tests/hooks/useCartToQuote.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+      replace: vi.fn(),
+      reload: vi.fn(),
+      ancestorOrigins: {} as DOMStringList,
+      host: 'localhost',
+      hostname: 'localhost',
+      origin: 'http://localhost',
+      port: '',
+      protocol: 'http:',
+    };
+    (window as unknown as { location: Location }).location = mockLocation;
```
bc-akshatkaushik, line 30: [question]

Is there a better way to mock window location than casting it as `unknown`?

bc-amandalan: Good question. I updated the test with `Object.defineProperty()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty and now it doesn't require type casting:

PR #380, opened 2025-06-17
feat(catalyst): B2B-2626 adding styles for catalyst notifications

bc-victor:
Jira: [B2B-2626](https://bigcommercecloud.atlassian.net/browse/B2B-2626)

## What/Why?
Adding styles for Catalyst notifications.

## Rollout/Rollback
Revert

## Testing

Screenshot 2025-06-17 at 1 49 31 p m
Screenshot 2025-06-17 at 2 56 25 p m
Screenshot 2025-06-17 at 2 57 03 p m

### DIFF
--- a/apps/storefront/src/components/B3Tip.tsx
+++ b/apps/storefront/src/components/B3Tip.tsx
@@ -2,19 +2,19 @@ import { Alert, AlertTitle, Box, Snackbar } from '@mui/material';
 
 import useMobile from '@/hooks/useMobile';
 import { MsgsProps, TipMessagesProps } from '@/shared/dynamicallyVariable/context/config';
+import { platform } from '@/utils';
 
 interface B3TipProps extends TipMessagesProps {
   handleItemClose: (id: number | string) => void;
   handleAllClose: (id: string | number, reason: string) => void;
 }
 
-function MessageAlert({
-  msg,
-  onClose,
-}: {
+interface AlertProps {
   msg: MsgsProps;
   onClose: (id: string | number) => void;
-}) {
+}
+
+function MessageAlert({ msg, onClose }: AlertProps) {
   const Body = msg.jsx ? msg.jsx : () => <span>{msg.msg}</span>;
 
   return (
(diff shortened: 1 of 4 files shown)


apps/storefront/src/components/B3Tip.tsx
```diff
@@ -42,6 +42,69 @@ function MessageAlert({
   );
 }
 
+const CatalystColors = {
```
leeBigCommerce, line 45: what's the plan with all the Catalyst variants?

If there's going be lots of them, it might be worth us considering a specific folder to hold styled components which have both a standard and Catalyst versions

We could then make sure each variant adhered to the same interfaces so they can be easily interchanged in the main app

But if these are the rare exception or temporary solutions, that might be too much

apps/storefront/src/pages/PDP/useAddedToShoppingListAlert.tsx
```diff
@@ -1,8 +1,32 @@
...
 
 import { useAppSelector } from '@/store';
-import { globalSnackbar } from '@/utils';
+import { globalSnackbar, platform } from '@/utils';
+
+function CatalystButton({
+  goToShoppingDetail,
+  children,
+}: {
+  goToShoppingDetail: () => void | undefined;
```
leeBigCommerce, line 12: other than the onClick naming, this is the same as `B3AddToQuoteTip.tsx`

which again makes me wonder about needing a more robust system for picking the variants

apps/storefront/src/components/B3Tip.tsx
```diff
@@ -42,6 +42,69 @@ function MessageAlert({
...
+        padding: '12px 12px 12px 16px',
+        color: '#000',
+        fontWeight: 400,
+        letterSpacing: 'normal',
+        fontSize: '14px',
+        fontFamily:
+          'ui-sans-serif,system-ui,-apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica Neue,Arial,Noto Sans,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol,Noto Color Emoji',
+        textWrap: 'auto',
+
+        '& button[title="Close"]': {
```
leeBigCommerce, line 86: do we translate the button title? If so this might be a fragile selector

apps/storefront/src/components/B3Tip.tsx
```diff
@@ -42,6 +42,69 @@ function MessageAlert({
...
+          display: msg.isClose ? 'flex' : 'none',
+        },
+
+        '& .MuiAlert-message': {
```
leeBigCommerce, line 90: coupling ourselves to an implementation detail of material here

does Material consider its classnames part of an API that it promises not to break between releases?

leeBigCommerce (approved):

PR #386, opened 2025-06-23
feat(quick-order): B2B-3183 enable quick order for users representing child companies based on buyer permission

bc-amandalan:
Jira: [B2B-2960](https://bigcommercecloud.atlassian.net/browse/B2B-2960)

### Context 
Currently there is no permission gating the quick order page. So, even if an user does not have order view permission, the user can go to quick order page and see through the purchased products of the company. This is problematic as ideally this page should be gated behind order view permission (code: `get_orders`).

Moreover, when an user represents a child company, the user wouldn't see the quick order page because currently the permission logic uses `subsidiariesCompanyKey` in `apps/storefront/src/shared/routeList.ts` to filter out pages that shouldn't appear. And the quick order page route in `routeList` does not have `subsidiariesCompanyKey`.

##### 1. Quick order page needs to be gated by `get_orders`
- When user has  `get_orders`, the user can access the quick order page (What this PR does)
- If user does not have Purchasability permission (code: `purchase_enable`), the page only shows the purchased products table without any action buttons, and quick order pad does not show (The current code already does it)
- If user has `purchase_enable`, the page would show action buttons in the purchased products table and the quick order pad (The current code already does it)

##### 3. When user browses the purchases product table, we need to make sure the products are returned based on the permission levels and whether the user is representing the child company
- When `get_orders` level 1, see only the products from the user's previous orders
- When `get_orders` level 2, see the products from the previous orders of the user's company
- When `get_orders` level 3 and not representing child company, see the products from the previous orders of the user's company
- When `get_orders` level 3 and representing child company, see the products from the previous orders of the child company

### Solution

- To achieve no.1, in `apps/storefront/src/shared/routeList.ts` interface `RouteItem` has an optional field `permissionCodes` that controls the front-end gating of a page. The code would need to match the back-end one, in this case  `get_orders`
- To achieve no.2, we need to add `subsidiariesCompanyKey` to the quick order route in `apps/storefront/src/shared/routeList.ts`. When it combines with `permissionCodes`, it would check if the user has the permission with level 3. 
- To achieve no.3, we would need some changes in the back-end https://github.com/bigcommerce/b2b-edition-api/pull/5747. 

## Rollout/Rollback
Will need to merge https://github.com/bigcommerce/b2b-edition-api/pull/5747 first so that users can see the correct purchased products. 

If any issue happens, we should revert both this PR and https://github.com/bigcommerce/b2b-edition-api/pull/5747. 

## Testing
- Added tests to cover different situations for quick order page route
- Manual tests
#### Before:

User can use quick order page to see purchased products even if the user does not have order view

https://github.com/user-attachments/assets/4c1b752f-5ed1-41dc-8ac7-44efd6b33461

Quick order tab, Quick order Pad and Purchased products are not showing when representing a child company

https://github.com/user-attachments/assets/9fa7468a-906f-489f-9f30-138b8c1bc8d4

#### After (with b2b-edition-api checkout to https://github.com/bigcommerce/b2b-edition-api/pull/5747):

User without order view can't see quick order page

https://github.com/user-attachments/assets/38acf0a4-8598-408a-8bf2-4ac577ac3b9c

User with order view but without purchasability can see the page but it does not have table action buttons and quick order pad

https://github.com/user-attachments/assets/3d0cc40f-4c28-4861-9461-d7b136eda3ee

User with order view and purchasability can see the page which has table action buttons and quick order pad

https://github.com/user-attachments/assets/52016003-1d79-4e88-b2d8-9fa04b622cc5

When representing child company, user with order view permission level 2 (company) can't see the quick order page

https://github.com/user-attachments/assets/f5c3e61f-de86-4837-b460-6a0f508e9323

When representing child company, user with order view permission level 3 (company and subsidiaries) and purchasability permission level 1 (user) can see the quick order page but no action buttons or quick order pad

https://github.com/user-attachments/assets/7b474fa8-c11a-4f7d-8b72-04871cb92603

When representing child company, user with order view permission level 3 (company and subsidiaries) and purchasability permission level 3 (company and subsidiaries) can see the quick order page which has action buttons and quick order pad

https://github.com/user-attachments/assets/6e536d1a-99f0-4ca4-b6d4-221a15ad5f0e

[B2B-2960]: https://bigcommercecloud.atlassian.net/browse/B2B-2960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/components/layout/B3Nav.tsx
```diff
@@ -148,16 +148,22 @@ export default function B3Nav({ closeSidebar }: B3NavProps) {
     const subsidiariesPermission = getSubsidiariesPermission(routes);
 
     if (selectCompanyHierarchyId) {
-      routes = routes.filter((route) =>
-        route?.subsidiariesCompanyKey
+      routes = routes.filter((route) => {
+        if (route?.subsidiariesCompanyKey === 'quickOrderPad') {
+          return true;
+        }
+        return route?.subsidiariesCompanyKey
```
bc-akshatkaushik, line 155:
```suggestion
        if (route?.subsidiariesCompanyKey) {
            if (route.subsidiariesCompanyKey === 'quickOrderPad')
                return true;
            return subsidiariesPermission[route.subsidiariesCompanyKey];
        }
        return false;
```
💅 We can combine the checks for route and route.subsidiarycompany into one

bc-akshatkaushik: [question]

Similar to `subsidiariesPermission[route.subsidiariesCompanyKey]`, do we need to check for the order permissions to gate the `quickOrder` route?

apps/storefront/src/shared/routeList.ts
```diff
@@ -310,6 +313,10 @@ export const getAllowedRoutesWithoutComponent = (globalState: GlobalState): Buye
         });
       }
 
+      if (path === '/purchased-products' && hasPermission) {
+        return true;
+      }
+
```
bc-akshatkaushik, line 319: [question]

Why do we need this early return for `path === '/purchased-products'`, isn't it already covered by the `return hasPermission` on the next line?

[suggestion]

```suggestion
```

bc-akshatkaushik:
[question]

> Permission Integration: Implemented proper permission codes using quickOrderPermissionCodes

I can see the permission codes being introduced for order permissions but I don't see them being used from the changes in this PR, can you point me to where they are evaluated?
To get a better picture of the implementation

apps/storefront/src/components/layout/B3Nav.tsx
```diff
@@ -148,16 +148,22 @@ export default function B3Nav({ closeSidebar }: B3NavProps) {
...
+          : false;
+      });
     } else {
       routes = routes.filter((route) => {
         if (route?.subsidiariesCompanyKey === 'companyHierarchy') {
           return isEnabledCompanyHierarchy && subsidiariesPermission[route.subsidiariesCompanyKey];
         }
+        if (route?.subsidiariesCompanyKey === 'quickOrderPad') {
+          return true;
+        }
```
icatalina, line 166: not sure this does much 😅

apps/storefront/src/components/layout/B3Nav.tsx
```diff
@@ -148,16 +148,22 @@ export default function B3Nav({ closeSidebar }: B3NavProps) {
     const subsidiariesPermission = getSubsidiariesPermission(routes);
 
     if (selectCompanyHierarchyId) {
-      routes = routes.filter((route) =>
-        route?.subsidiariesCompanyKey
+      routes = routes.filter((route) => {
```
icatalina, line 151: 🤔 do we need to set `routes`? it looks like we could just return in both cases:

```suggestion
      return routes.filter((route) => {
```

bc-akshatkaushik (approved):
LGTM

bc-amandalan (commented):
Thank you so much Isaac, the changes look good to me. I think assume there were some communication with the product so we do not need to use two permissions to control the Nav bar 👏 👏 👏

PR #388, opened 2025-06-24
feat(common): B2B-3031 using microapps cdn for tier 1

deov31:
## What/Why?
I'm updating our after.sh deployment file to start using BC CDN to test stores of the prod tier1 list first before pushing the change for the whole prod environment.

## Rollout/Rollback
Revert 

## Testing
WIP


BC-EEspinosa (approved):
Looks good!

PR #392, opened 2025-06-25
feat(common): B2B-3031 using store hash as variable

deov31:
## What/Why?
Using store hashes from tier 1 list directly as variables in the after config script due to problems when reading the prod tier 1 json file.

## Rollout/Rollback
Revert

## Testing
WIP


BC-EEspinosa (approved):
LGTM

PR #393, opened 2025-06-26
fix(common): B2B-3031 fix after.sh script

BC-EEspinosa:
Jira: [B2B-3031](https://bigcommercecloud.atlassian.net/browse/B2B-3031)

## What/Why?
We removed  by mistake the creation for the payload required for non production environments in the after.sh script used by LaunchBay
I recovered such fragment of code

## Rollout/Rollback
Revert this PR

## Testing
Captura de pantalla 2025-06-25 a la(s) 7 36 58 p m

### DIFF
--- a/config/deploy/after.sh
+++ b/config/deploy/after.sh
@@ -43,6 +43,13 @@ if [[ $ENVIRONMENT =~ "production" ]]; then
 "revision": "${REVISION_TITLE}"
 }
 EOF
+else
+  tee deploy_revision_payload.json <<EOF >/dev/null
+{
+"deploy_all": true,
+"revision": "${REVISION_TITLE}"
+}
+EOF
 fi
 
 # create revision


deov31 (approved):
Thanks!

config/deploy/after.sh
```diff
@@ -43,6 +43,13 @@ if [[ $ENVIRONMENT =~ "production" ]]; then
 "revision": "${REVISION_TITLE}"
 }
 EOF
+else
+  tee deploy_revision_payload.json <<EOF >/dev/null
+{
+"deploy_all": true,
+"revision": "${REVISION_TITLE}"
+}
```
icatalina, line 51: 🗒️ No need to update it in this PR, but something to keep in mind in the future.

I know you fished this from history, but using `tee` with `>/dev/null` basically nulls the benefits of `tee`.

`tee` is like `cat`, but it also prints to screen, if we don't want to print the data to the log, we can use `cat`, if we do want the data to be printed we can remove `>/dev/null`.


```suggestion
cat > deploy_revision_payload.json <<EOF
{
"deploy_all": true,
"revision": "${REVISION_TITLE}"
}
```

same applies to all usages of `tee` in this file.

icatalina (approved):

PR #409, opened 2025-07-11
test(quick-order): B2BCAT-18 Add tests for QuickOrder + QuickOrder QuickAdd

icatalina:
⚠️ Use the different commits for review.

Jira: [B2BCAT-18](https://bigcommercecloud.atlassian.net/browse/B2BCAT-18)

## What/Why?
- Add an extra battery of tests for QuickOrder + Some test for QuickAdd in the QuickOrder Pad
- Temporarily added an extra test for a B2C version of the QuickAdd component
- Simplify `getVariantInfoBySkus` and remove the B2C version of the tests, as it is no longer applicable.

## Rollout/Rollback
Revert

## Testing
These are mostly test

[B2BCAT-18]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-18?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
```diff
@@ -179,7 +179,7 @@ export default function QuickAdd(props: AddToListContentProps) {
         return;
       }
 
-      if (isStock === '1' && quantity > Number(stock)) {
+      if (isStock === '1' && allQuantity > Number(stock)) {
```
icatalina, line 182: This is the only real behaviour change in this PR. If you look a few lines below, we use `allQuantity` to compute whether min/max have been reached. I'm guessing we need to do the same for checking if the stock has been exceeded. cc: @bc-micah

apps/storefront/src/pages/QuickOrder/index.quickadd.test.tsx
```diff
@@ -0,0 +1,768 @@
...
+    graphql.query('getCart', () => HttpResponse.json<GetCart>({ data: { site: { cart: null } } })),
+    graphql.mutation('createCartSimple', ({ variables }) =>
+      HttpResponse.json(createCartSimple(variables)),
+    ),
+  );
+
+  renderWithProviders(<QuickOrderPad />, { preloadedState });
+
+  const [skuInput, missingSkuInput] = screen.getAllByRole('textbox', { name: 'SKU#' });
+  const [qtyInput, missingQtyInput] = screen.getAllByRole('spinbutton', { name: 'Qty' });
```
leeBigCommerce, line 226: fun use of de-structoring 

`missingSkuInput` is a bit confusing as we're naming the input after a 'meta' quality of a value that we plan to enter into it

rather than the norm of naming an input based on its label/purpose

it's probably worth giving the special behaviour for "missing" products its own test

leeBigCommerce (approved):

bc-victor (approved):

PR #414, opened 2025-07-18
docs(account): B2BCAT-4 Add draft schema for account settings

icatalina:
Jira: [B2BCAT-4](https://bigcommercecloud.atlassian.net/browse/B2BCAT-4)

## What/Why?
- Add draft schema for account settings
- Remove unused dependencies from the rfc/graphql-schema package

## Rollout/Rollback
Revert

## Testing
n/a

[B2BCAT-4]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-4?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rfc/graphql-schema/additionalTypeDefs/byPage/accountSettings.ts
+++ b/rfc/graphql-schema/additionalTypeDefs/byPage/accountSettings.ts
@@ -8,26 +8,22 @@ export default /* GraphQL */ `
     companyRelationship: CompanyRelationship
   }
 
-  type CompanyFormFields {
-    user: [FormField!]! # This includes b2c customer fields + b2b extra fields
+  type ExtraFields {
+    companyUser: [FormField!]!
   }
 
-  extend type FormFields {
-    company: CompanyFormFields!
+  extend type Settings {
+    extraFields: ExtraFields!
   }
 
-  input UpdateCustomerInput {
-    firstName: String!
-    lastName: String!
-    email: String!
-    phone: String
-    company: String
-    formFields: CustomerFormFieldsInput
+  extend input UpdateCustomerInput {
+    extraFields: CustomerFormFieldsInput # This is only used for companyUser extra fields
   }
 
-  type CustomerMutations {
-    updateCustomer(input: UpdateCustomerInput!): UpdateCustomerResult
-  }
+  # This is already defined in storefront schema, added here for clarity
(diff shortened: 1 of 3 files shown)


rfc/graphql-schema/additionalTypeDefs/byPage/accountSettings.ts
```diff
@@ -0,0 +1,26 @@
+export default /* GraphQL */ `
+  type CompanyRelationship {
+    company: Company!
+    role: Role!
+  }
```
icatalina, line 5: thoughts on this pattern? this scopes role to the customer-company relationship. Open to other suggestions.

The alternatives I thought of were:

```gql
type Customer {
  company: Company # This is backwards incompatible, as we already have a company property
  role: Role # This must be optional, some customers are not b2b, therefore have no role.
  # Also, not sure if it is a common pattern right now, but role might be dependent on the current active company? 🤷 
}
```

bit weirder:

```gql
type Company {
  role: Role! # this is also problematic, only company objects under `customer` will/could include a role.
}

type Customer {
  company: Company # suffers the same issue as above, company exists already
}
```

rfc/graphql-schema/additionalTypeDefs/byPage/accountSettings.ts
```diff
@@ -0,0 +1,26 @@
...
+  input UpdateCustomerInput {
+    firstName: String!
+    lastName: String!
+    email: String!
+    phone: String
+    formFields: CustomerFormFieldsInput
+  }
+
+  type CustomerMutations {
+    updateCustomer(input: UpdateCustomerInput!): UpdateCustomerResult
```
icatalina, line 20: This is an existing mutation/input, I added it here for completeness. If we can fold `formFields` with `extraFields` we won't need to change the mutation, just send the fields to the correct storage.

rfc/graphql-schema/additionalTypeDefs/byPage/accountSettings.ts
bc-micah: I think the schema you defined w/ it being required fields make sense to me the most.

leeBigCommerce (approved):

PR #429, opened 2025-08-07
docs(invoices): B2BCAT-72 Add schemas for Invoices/InvoicePayment

icatalina:
Jira: [B2BCAT-72](https://bigcommercecloud.atlassian.net/browse/B2BCAT-72)

## What/Why?
Add draft schema for Invoice/InvoicePayments

## Rollout/Rollback
Revert

## Testing
N/A

[B2BCAT-72]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-72?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
+++ b/rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
@@ -1,0 +1,147 @@
+export default /* GraphQL */ `
+  enum InvoiceStatus {
+    OPEN
+    PAID
+    OVERDUE
+    CANCELLED
+  }
+
+  input InvoiceDateRangeFilterInput {
+    from: DateTime
+    to: DateTime
+  }
+
+  input InvoiceFiltersInput {
+    search: String
+    dateRange: InvoiceDateRangeFilterInput
+    status: InvoiceStatus
+  }
+
+  enum InvoicesSortInput {
+    INVOICE_NUMBER_A_TO_Z
+    INVOICE_NUMBER_Z_TO_A
+    ORDER_ID_A_TO_Z
+    ORDER_ID_Z_TO_A
+    CREATED_AT_NEWEST
+    CREATED_AT_OLDEST
+    DUE_DATE_NEWEST
+    DUE_DATE_OLDEST
+    ORIGINAL_BALANCE_HIGHEST
+    ORIGINAL_BALANCE_LOWEST
+    OPEN_BALANCE_HIGHEST
+    OPEN_BALANCE_LOWEST
+  }
+
+  type CreateCartFromInvoiceError implements Error {
+    message: String!
+  }
+
+  type CartCreateResult {
+    cart: Cart
+    errors: [CreateCartFromInvoiceError!]!
+  }
+
+  type GenerateInvoicePdfError implements Error {
+    message: String!
+  }
+
+  type GenerateInvoicePdfResult {
+    url: String
+    errors: [GenerateInvoicePdfError!]!
+  }
+
+  type InvoiceBalances {
+    open: Money!
+    original: Money!
+  }
+
+  type Invoice {
+    id: ID!
+    invoiceNumber: String!
+    createdAt: DateTime!
+    dueDate: DateTime!
+    balance: InvoiceBalances!
+    status: InvoiceStatus!
+    order: Order!
+    company: Company!
+  }
+
+  type InvoiceEdge {
+    node: Invoice!
+    cursor: String!
+  }
+
+  type InvoiceConnection {
+    edges: [InvoiceEdge!]!
+    pageInfo: PageInfo!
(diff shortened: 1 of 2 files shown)


rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
```diff
@@ -0,0 +1,143 @@
...
+    CANCELLED
+  }
+
+  type InvoiceStats {
+    openBalance: Float!
+    overdueBalance: Float!
+  }
+
+  input InvoiceDateRangeFilterInput {
+    from: DateTime!
```
leeBigCommerce, line 15: only from is mandatory? We don't want to offer "since forever"?

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
```diff
@@ -0,0 +1,143 @@
...
+    errors: [GenerateInvoicePdfError!]!
+  }
+
+  type Invoice {
+    id: ID!
+    invoiceNumber: String!
+    createdAt: DateTime!
+    dueDate: DateTime!
+    openBalance: Money!
+    originalBalance: Money!
```
leeBigCommerce, line 64: worth its own namespace (`balance.original`) or we confident that these two will be it for a while?

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
```diff
@@ -0,0 +1,143 @@
...
+  type InvoiceEdge {
+    node: Invoice!
+    cursor: String!
+  }
+
+  type InvoiceConnection {
+    edges: [InvoiceEdge!]!
+    pageInfo: PageInfo!
+    collectionInfo: CollectionInfo!
+    stats: InvoiceStats!
```
leeBigCommerce, line 79: `stats` will be an odd one out amongst all our `XConnection` types

what is it detailing, the total costs of all outstanding invoices?


we could avoid this with something like

```gql
type InvoiceTotals {
  openBalance: Float!
  overdueBalance: Float!
}

extend type Company {
  # etc etc
  invoiceTotals: InvoiceTotals!
}
```

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
```diff
@@ -0,0 +1,143 @@
...
+      after: String
+      first: Int
+      last: Int
+    ): InvoiceConnection!
+  }
+
+  input InvoicePaymentInput {
+    id: ID!
+    # If no amount is specified, the whole invoice amount will be used
+    amount: Float
```
leeBigCommerce, line 102: shame we can't have unions for inputs

leeBigCommerce (approved):

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
icatalina: I know 😔 

I thought of having two different mutations, one to pay in full, one to specify amounts, but it feels a bit extreme. What do you think?

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
icatalina: I do like the `balance` property, I moved both elements inside. I don't think we'll have anything else in the future, but 🤷

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
bc-micah: From should be optional - we will want to allow the user to ask for all their invoices like Lee pointed out. Since this is entirely w/ b2b domain - we have the ability to implement it - let's go with that.

For orders we will probably need to ask the orders team what is feasible and have a discussion around it.

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
leeBigCommerce: yeah, I'm not sure which is best. 

I guess splitting means there's fewer possible error types for the pay in full (as it doesn't need to validate the user's amount input) so it's a simpler signature for a reasonable proportion of mutations.

Would there be any BE benefit to splitting?

---

otherwise

maybe we can name it something more indicative of being less than the full amount?
`shortPayAmount`, `partialAmount`, 🤔 

---

also, should it be our `Money` type?

rfc/graphql-schema/additionalTypeDefs/byPage/invoices.ts
icatalina: made optional

PR #445, opened 2025-08-12
ci(common): B2BCAT-86 Only enable failOnConsole when in CI

icatalina:
Jira: [B2BCAT-86](https://bigcommercecloud.atlassian.net/browse/B2BCAT-86)

## What/Why?
Only enable `failOnConsole` when running in CI. Checking for console.logs is helpful in CI, but messy in dev. Disabling it for DEV. This allows for easier debugging in local env. and ensures `graphql.operation` can be used.

## Rollout/Rollback
Revert

### Before:
image

### After:
image


apps/storefront/vite.config.ts
```diff
@@ -3,18 +3,14 @@ import legacy from '@vitejs/plugin-legacy';
 import react from '@vitejs/plugin-react';
 import path from 'path';
 import { visualizer } from 'rollup-plugin-visualizer';
-import { defineConfig, loadEnv } from 'vite';
+import { defineConfig, loadEnv, UserConfig } from 'vite';
+import { ViteUserConfig } from 'vitest/config';
 
-export default defineConfig(({ mode }) => {
+export default defineConfig(({ mode }): UserConfig & Pick<ViteUserConfig, 'test'> => {
```
icatalina, line 9: Added types here for autocomplete. `defineConfig` does not include the `test` property in the config.

PR #476, opened 2025-08-26
docs(quotes): B2BCAT-66 Add schemas for Draft Quote

leeBigCommerce:
Jira: [B2BCAT-66](https://bigcommercecloud.atlassian.net/browse/B2BCAT-66)

## What/Why?

- Draft of the schema needed to drive the Draft Quote page

In this schema I have chosen to keep "Draft" as a status on a unified "Quote" type/entity, rather than give it its own

As such, mutations only applicable to Drafts appear under the `quote` namespace, but can technically only be applied to true Drafts. For example, `quote. updateLineItem` will need to return an error for all but "Draft" status quotes.

### Why?

If we had seperate `DraftQuote`, `OpenQuote`, `ExpiredQuote`, `OrderedQuote` and `ArchivedQuote` types they would exact same schemas (possibly with just `expirationDate` needing to be optional to suite `DraftQuotes`)

Keeping individual types would lead us to use different mutation namespaces. If we did, there would be ~20 mutations across all of them, with likely 4 exact duplicates within each (e.g. `addMessage`, `addAttachment`, `removeAttachment` and `createPDF`).

The full picture would be more explicit and business accurate, but it would also be much larger.

---

In contrast unifying the statuses within 1 type and, likewise, unifying the mutations under one namespace (`Quote`) presents as smaller/simpler.

However it is not as explicit and as such documentation and error messages will need to fill the gaps of a more ambiguous schema. For example, we will need to document that `addLineItem` only works when the target `Quote` is in draft and return specific errors if this is ignored.

The gamble here is that this 'simplified' is explicit enough to self document general use cases and that our view models for the different status are stable and unlikely to diverge in future. 

## Rollout/Rollback

- Revert

## Testing

- N/A draft schema only

[B2BCAT-66]: https://bigcommercecloud.atlassian.net/browse/B2BCAT-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
+++ b/rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
@@ -4,9 +4,65 @@ export default /* GraphQL */ `
     email: String!
   }
 
+  # CartSelectedOption clone STARTS
+  # Should we have an interface behind these?
+  # Or is this coincidental duplication?
+  # Can we imagine adding items to a cart and a draft quote diverging for business reasons?
+  interface QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+  }
+
+  type QuoteSelectedCheckboxOption implements QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+    value: String!
+    valueEntityId: Int!
+  }
+
+  type QuoteSelectedDateFieldOption implements QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+    date: DateTimeExtended!
+  }
+
+  type QuoteSelectedFileUploadOption implements QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+    fileName: String!
+  }
+
+  type QuoteSelectedMultiLineTextFieldOption implements QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+    text: String!
+  }
+
+  type QuoteSelectedMultipleChoiceOption implements QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+    value: String!
+    valueEntityId: Int!
+  }
+
+  type QuoteSelectedNumberFieldOption implements QuoteSelectedOption {
+    entityId: Int!
+    name: String!
+    number: Float!
+  }
+
+  type QuoteSelectedTextFieldOption {
+    entityId: Int!
+    name: String!
+    text: String!
+  }
+  # CartSelectedOption clone ENDS
+
   # could this reuse QuotePhysicalItem (or an interface that backs both)?
   type QuoteLineItem implements Node {
     id: ID!
+    configuredProductId: ID!
+    selectedOptions: [QuoteSelectedOption!]!
     sku: String
     name: String!
     quantity: Int!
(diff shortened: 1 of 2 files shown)


leeBigCommerce:
Once we're happy with the direction here, the other two `Quote` related schema drafts will need to be brought into line

rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
```diff
@@ -0,0 +1,344 @@
...
+
+  input QuoteLineItemInput {
+    quantity: Int!
+    productEntityId: ID!
+    variantEntityId: ID
+    selectedOptions: QuoteSelectedOptionsInput
+  }
+
+  input AddQuoteLineItemsInput {
+    quoteId: ID!
```
bc-micah, line 178: Let's make this `quoteId` optional because a user can be creating the draft quote on the first add.

rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
```diff
@@ -0,0 +1,344 @@
...
+    | AnotherAddQuoteLineItemsError
+
+  type QuoteMutations {
+    addLineItems(input: AddQuoteLineItemsInput): AddQuoteLineItemsResult!
+  }
+
+  input UpdateQuoteLineItemInput {
+    quoteId: ID!
+    lineItemId: ID!
+    lineItem: QuoteLineItemInput
```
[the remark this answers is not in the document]
leeBigCommerce: done:

- https://github.com/bigcommerce/b2b-buyer-portal/pull/476/commits/6ac449ce137bcd99c75273e221769deaf14d7672

rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
```diff
@@ -0,0 +1,344 @@
+export default /* GraphQL */ `
+  type SalesRep implements Node {
+    id: ID!
+    email: String!
+  }
+
+  # could this reuse QuotePhysicalItem (or an interface that backs both)?
+  type QuoteLineItem implements Node {
+    id: ID!
```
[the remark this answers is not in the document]
leeBigCommerce: done and done:

- https://github.com/bigcommerce/b2b-buyer-portal/pull/476/commits/6ac449ce137bcd99c75273e221769deaf14d7672

rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
bc-victor: we'll keep as required and add a `quote.create` mutation

rfc/graphql-schema/additionalTypeDefs/byPage/draftQuote.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  extend type QuoteMutations {
+    removeAttachment(
+      input: RemoveQuoteAttachmentInput!
+    ): RemoveAttachmentResult!
+  }
+
+  input CreateQuoteInput {
+      title: String
+      lineItems: [QuoteLineItemInput!]
```
icatalina, line 300: I don't think we'll need `lineItems` considering the UI we are proposing, but 🤷

icatalina (approved):
Looks good, thanks for picking this up @bc-victor.

Make sure you squash all the commits before merging please 🙇

PR #504, opened 2025-09-01
feat(orders): B2B-3250 allow download digital products files

deov31:
## What/Why?
Currently the buyer portal does not allow users to download their order's product digital files from the portal, so in order to allow users to download files of digital products from the buyer portal we need to get the downloadUrls provided and offer them a new UI to access the files.

The api currently does not return information related to the max amount limit of order's digital files so I'm addoing a couple of utils to find if the response blob is a type text/html to handle the display of the error for the users.

## Rollout/Rollback
Revert PR

## Testing

https://github.com/user-attachments/assets/68e48cc9-bb8c-4943-9c12-3d1ade323ab1

### DIFF
--- a/apps/storefront/src/components/B3ProductList.tsx
+++ b/apps/storefront/src/components/B3ProductList.tsx
@@ -1,6 +1,6 @@
 import { ChangeEvent, KeyboardEvent, ReactElement, useEffect, useState } from 'react';
 import styled from '@emotion/styled';
-import { Box, Checkbox, FormControlLabel, TextField, Typography } from '@mui/material';
+import { Box, Button, Checkbox, FormControlLabel, TextField, Typography } from '@mui/material';
 import noop from 'lodash-es/noop';
 
 import { PRODUCT_DEFAULT_IMAGE } from '@/constants';
@@ -124,6 +124,7 @@ interface ProductProps<T> {
   canToProduct?: boolean;
   textAlign?: string;
   type?: string;
+  getCurrentProductUrls?: (productId: number | undefined) => void;
 }
 
 export default function B3ProductList<T>(props: ProductProps<T>) {
@@ -142,6 +143,7 @@ export default function B3ProductList<T>(props: ProductProps<T>) {
     textAlign = 'left',
     money,
     type,
+    getCurrentProductUrls,
   } = props;
 
   const [list, setList] = useState<ProductItem[]>([]);
@@ -402,6 +404,20 @@ export default function B3ProductList<T>(props: ProductProps<T>) {
                 <Typography variant="body1" color="#616161">
                   {product.sku}
                 </Typography>
+                {product.type === 'digital' &&
+                  product.downloadFileUrls &&
+                  product.downloadFileUrls.length > 0 && (
+                    <Button
+                      sx={{
+                        m: '0 0 0 -8px',
+                        minWidth: 0,
+                      }}
+                      variant="text"
+                      onClick={() => getCurrentProductUrls?.(product.product_id)}
+                    >
+                      {b3Lang('orderDetail.digitalProducts.viewFiles')}
+                    </Button>
+                  )}
                 {(product.product_options || []).map((option) => (
                   <ProductOptionText
                     key={`${option.option_id}`}
--- a/apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
+++ b/apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
@@ -1,12 +1,15 @@
-import { useContext } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { Box, Card, CardContent, Stack, Typography } from '@mui/material';
 
 import { B3ProductList } from '@/components';
 import { useMobile } from '@/hooks';
 import { useB3Lang } from '@/lib/lang';
 
-import { OrderBillings } from '../../../types';
+import { Address, OrderProductItem } from '../../../types';
 import { OrderDetailsContext } from '../context/OrderDetailsContext';
+
+import DownloadDigitalProductsDialog from './DownloadDigitalProductsDialog/DownloadDigitalProductsDialog';
+import { getDigitalDownloadElements } from './getDigitalDownloadElements';
 
 type OrderBillingProps = {
   isCurrentCompany: boolean;
@@ -14,16 +17,53 @@ type OrderBillingProps = {
 
 export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
   const {
-    state: { billings = [], addressLabelPermission, orderId, money },
+    state: { billingAddress, digitalProducts = [], addressLabelPermission, orderId, money },
   } = useContext(OrderDetailsContext);
 
   const [isMobile] = useMobile();
 
   const b3Lang = useB3Lang();
 
-  const getFullName = (billing: OrderBillings) => {
-    const { billingAddress } = billing;
+  const [digitalProductsWithUrl, setDigitalProductsWithUrl] = useState<OrderProductItem[]>([]);
+  const [currentDigitalProduct, setCurrentDigitalProduct] = useState<OrderProductItem>();
 
+  useEffect(() => {
+    const getDigitalProductsInformation = async (id: number | string) => {
+      const elements = await getDigitalDownloadElements(id);
+
+      if (!elements.length) {
+        return;
+      }
+
+      const digitalProductsData = elements.map((item) => item.node);
+
+      const digitalProductsUrls = digitalProducts.map((product: OrderProductItem) => {
+        const fileUrls =
+          digitalProductsData.find(({ productEntityId }) => productEntityId === product.product_id)
+            ?.downloadFileUrls || [];
+
+        return {
+          ...product,
+          downloadFileUrls: fileUrls,
+        };
+      });
+
+      setDigitalProductsWithUrl(digitalProductsUrls);
+    };
+
+    if (orderId) {
+      getDigitalProductsInformation(orderId);
+    }
+  }, [digitalProducts, orderId]);
+
+  const getCurrentProductUrls = (productId: number | undefined) => {
+    const currentProduct = digitalProductsWithUrl?.find(
+      (product) => product.product_id === productId,
+    );
+    setCurrentDigitalProduct(currentProduct);
+  };
+
+  const getFullName = (billingAddress: Address) => {
     if (billingAddress) {
       const { first_name: firstName, last_name: lastName } = billingAddress;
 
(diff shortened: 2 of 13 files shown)


apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
```diff
@@ -105,13 +186,69 @@ export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
...
+            title={b3Lang('orderDetail.digitalProducts.filesToDownload')}
+            rightSizeBtn={undefined}
+            leftSizeBtn={b3Lang('orderDetail.digitalProducts.close')}
+            maxWidth="md"
+            loading={isDigitalProductLoading}
+            showRightBtn={false}
+          >
+            {currentDigitalProduct?.downloadFileUrls?.map((fileUrl: string, index: number) => (
+              <Box
+                key={uuid()}
```
bc-victor, line 210: The purpose of the key prop is to give React a stable identity for each item in a list. When you call key={uuid()}, you generate a brand new, random key for every single item every time your component re-renders.

This tells React that the entire list of old items was destroyed and replaced with a completely new list of items. As a result, React will destroy all the existing components in the list and create brand new instances for every item.

I think using `fileUrl` is safe as key since they will all be different.

apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
```diff
@@ -21,6 +40,56 @@ export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
 
   const b3Lang = useB3Lang();
 
+  const [isDigitalDownloadOpen, setIsDigitalDownloadOpen] = useState(false);
+  const [isDigitalProductLoading, setIsDigitalProductLoading] = useState(false);
```
icatalina, line 44: `isDigitalProductLoading` is really not doing much, the button won't be shown until the data is fetched. It can be removed.

If we get rid of `isDigitalProductLoading`, then we can remove the finally... and if we can remove the `finally`, we can probably remove the catch and just let it reap.

```tsx
  const getDigitalProductsInformation = useCallback(
    async (id: number | string) => {
      const elements = await getDigitalDownloadElements(id);

      if (!elements.length) {
        return;
      }

      const digitalProductsData = elements.map((item) => item.node);

      const digitalProducts = billings[0].digitalProducts.map((product: OrderProductItem) => {
        const fileUrls =
          digitalProductsData.find(({ productEntityId }) => productEntityId === product.product_id)
            ?.downloadFileUrls || [];```

File: apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx [hunk 3]
```diff
@@ -1,17 +1,36 @@
...
+type DigitalProductNode = {
+  downloadFileUrls: string[];
+  downloadPageUrl: string;
+  name: string;
+  productEntityId: number;
+};
+
+type DigitalProduct = {
+  node: DigitalProductNode;
+};
```
icatalina, line 32: If we move these data definitions to the fetch method, we don't need to manually annotate every time we read the array.

```tsx
type DigitalProductNode = {
  downloadFileUrls: string[];
  downloadPageUrl: string;
  name: string;
  productEntityId: number;
};

type DigitalProduct = {
  node: DigitalProductNode;
};

export const getDigitalDownloadElements = (orderId: number | string): Promise<DigitalProduct[]> =>
```

apps/storefront/src/shared/service/b2b/graphql/orders.ts
```diff
@@ -467,3 +515,11 @@ export const getOrdersCreatedByUser = (companyId: number, module: number) =>
   B3Request.graphqlB2B({
     query: getCreatedByUser(companyId, module, 'createdByUser'),
   });
+
+export const getDigitalDownloadElements = (orderId: number | string) =>
```
icatalina, line 519: Let's move this fetch inside the `OrderDetail/components/getDigitalDownloadElements.ts` folder, ideally each component should query the data it needs, without sharing the queries, it avoids bloated queries.

apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
```diff
@@ -21,6 +40,56 @@ export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
...
+    });
+
+    setDigitalProducts(digitalProducts);
+  }, [billings, orderId]);
+
+  useEffect(() => {
+    if (orderId) {
+      getDigitalProductsInformation();
+    }
+  }, [orderId, getDigitalProductsInformation]);
```
icatalina, line 83: we can embed all the code inside the `useEffect` and remove the need for the `useCallback` and the `useEffect`:

```tsx
useEffect(() => {
  const fetchOrderData = async () => {
    if (!orderId) return;

    const elements = await getDigitalDownloadElements(orderId);

    // ...
  };

	fetchOrderData();
}, [billings, orderId]);
```

apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
```diff
@@ -21,6 +40,56 @@ export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
...
+
+  const getCurrentDigitalProduct = (productId: number) => {
+    const current = digitalProducts.find(
+      (product: ProductItem) => product.product_id === productId,
```
icatalina, line 87: `digitalProducts` is already typed, no need to specify here:

```suggestion
      (product) => product.product_id === productId,
```

apps/storefront/src/components/B3ProductList.tsx
```diff
@@ -142,6 +145,7 @@ export default function B3ProductList<T>(props: ProductProps<T>) {
     textAlign = 'left',
     money,
     type,
+    getDigitalDownloadLinks = noop,
```
icatalina, line 148: Instead of adding a `noop` here, you can call the method conditonally: `getDigitalDownloadLinks?.()`

icatalina (changes requested):
Hey @deov31, great work, I added a bunch of comment which I think will help simplify the code. Let me know what you think.

I have a more general question though, It looks like the original code allows for more than one digital "consignment" (whatever that is), but the new code assumes there's only ever one. You can see this in the loop that renders the list of consignments:

```tsx
{billings.map((billingItem: OrderBillings) => (
```

Could this be an issue?

apps/storefront/src/pages/OrderDetail/components/DownloadDigitalProductsDialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+        return (
+          <Box
+            key={key}
+            sx={{
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'space-between',
+              padding: '8px 0',
+              margin: '15px 0 15px 0',
+              border: '1px solid #1976D2',
```
bc-micah, line 48:
```suggestion
              borderWidth: '1px',
              borderStyle: 'solid',
              borderColor: 'primary.main',
```

apps/storefront/src/pages/OrderDetail/components/DownloadDigitalProductsDialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+            <Box sx={{ display: 'flex' }}>
+              <InsertDriveFileOutlined
+                sx={{
+                  color: '#1976D2',
+                  fontSize: '40px',
+                }}
+              />
+              <Typography
+                sx={{
+                  color: '#1976D2',
```
bc-micah, line 62: let's use the theme values for consistency throughout the file.

apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
```diff
@@ -21,6 +27,47 @@ export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
...
+    const getDigitalProductsInformation = async (id: number | string) => {
+      const elements = await getDigitalDownloadElements(id);
+
+      if (!elements.length) {
+        return;
+      }
+
+      const digitalProductsData = elements.map((item) => item.node);
+
+      const digitalProducts = billings[0].digitalProducts.map((product: OrderProductItem) => {
```
bc-micah, line 46: Per our mobbing discussion:

- we should just introduce a new state variable called `billingAddress` and `digitalProducts` (`billings` being an array of 1 item that has these two values is unnecessary)
- combine usage of `convertB2BOrderDetails` and `convertBCOrderDetails` as they are the same function 😕

apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
```diff
@@ -105,13 +166,21 @@ export default function OrderBilling({ isCurrentCompany }: OrderBillingProps) {
...
+              getDigitalDownloadLinks={getCurrentDigitalProduct}
             />
           </CardContent>
+          <DownloadDigitalProductsDialog
+            isOpen={isDigitalDownloadOpen}
+            onClose={() => setIsDigitalDownloadOpen(false)}
+            product={currentDigitalProduct}
+            b3Lang={b3Lang}
+            handleDownloadDigitalFile={handleDownloadDigitalFile}
+          />
```
bc-micah, line 183:
```suggestion
          { currentDigitalProduct && (<DownloadDigitalProductsDialog
            isOpen={isDigitalDownloadOpen}
            onClose={() => setIsDigitalDownloadOpen(false)}
            product={currentDigitalProduct}
            b3Lang={b3Lang}
            handleDownloadDigitalFile={handleDownloadDigitalFile}
          />)}
```

apps/storefront/src/pages/OrderDetail/components/DownloadDigitalProductsDialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+interface DownloadDigitalProductsDialogProps {
+  isOpen: boolean;
+  onClose: () => void;
+  product?: {
+    downloadFileUrls?: string[];
+    product_id?: number;
+    name?: string;
+    sku?: string;
+  };
```
bc-micah, line 14:
```suggestion
  product: {
    downloadFileUrls?: string[];
    product_id?: number;
    name: string;
    sku: string;
  };
```

bc-micah (commented):
Looks great overall - some minor improvements to match mui theme guidlines and improving props.

There is a larger ask of refactoring the billing => `billingAddress` and `digitalProducts`. Looks like this is the only component that uses it so I think this is safe.

apps/storefront/src/pages/OrderDetail/components/OrderBilling.tsx
deov31: @bc-micah @bc-victor Waiting for currentDigitalProduct to be set up when clicking View Files button was causing a delay in the modal creating a bad experience for users. I changed the way we open the modal here, I deleted one of the state variables `isDigitalDownloadOpen` and used currentDigitalProduct to validate if open should be opened or not, so in the dialog the products prop can stay undefined.

PR #556, opened 2025-09-29
fix(quotes): B2B-3456 fix submit draft quote flow for users with no address permissions

deov31:
## What/Why?

- Currently users with a custom role an no permission addresses aren't able to submit a quote because the api call throws an error and breaks the flow of the code, so it never reaches the part where the customer info is updated in the state and that information is required to submit the quote.

## Rollout/Rollback
Revert 

## Testing

Before:

https://github.com/user-attachments/assets/251f522c-a1c7-4cd2-a29c-52be4b0085c0

After: 

https://github.com/user-attachments/assets/66badc22-c59d-4848-8791-dee1b4a490e8


icatalina (changes requested):
Please, avoid adding more conditionals to the fetch method. See my comment. 🙇

icatalina (approved):

PR #564, opened 2025-10-02
feat(quotes): B2B-3554 getting available to sell and unlimited backorder values from backend

deov31:
## What/Why?

- I'm using the new availableToSell and umlimitedBackorder keys from searchProducts to update the soft validations strings in the QuoteTable and shopping list errored panel. Validations cover the different scenarios such as Min and Max quantity values for each product, if is out of stock and display the value of stock in case the user is setting a bigger quantity of the current stock value.
- I also updated the current warning stock message in the QuoteTable to use the availableToSell key so is compatible with backordered products.

## Rollout/Rollback
Revert

## Testing
WIP


bc-alvaroolvera (approved):
Reviewed and tested on mobbing session, thanks for doing the fix on the quantity input

icatalina (approved):

PR #568, opened 2025-10-07
fix(orders): B2B-3691 use searchproduct information to create line items object

deov31:
## What/Why?
In our new backendValidation function we were not using the current functionality that transforms the lineItems object to the one required by the cart create query. So I'm reusing the implementation and using it in both front end and backend validation functions. Since the lineitems optionList was not prepared for the query it was causing errores when adding a purchased product to the cart.

## Rollout/Rollback
Revert

## Testing

https://github.com/user-attachments/assets/e3c619e4-79e7-43f6-b176-dae19ae3c92d

### DIFF
--- a/apps/storefront/src/pages/QuickOrder/components/QuickOrderFooter.tsx
+++ b/apps/storefront/src/pages/QuickOrder/components/QuickOrderFooter.tsx
@@ -16,6 +16,7 @@ import {
   searchProducts,
 } from '@/shared/service/b2b';
 import { activeCurrencyInfoSelector, rolePermissionSelector, useAppSelector } from '@/store';
+import { Product } from '@/types';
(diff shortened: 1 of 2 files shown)


apps/storefront/src/pages/QuickOrder/components/QuickOrderFooter.tsx
```diff
@@ -233,20 +231,13 @@ function QuickOrderFooter(props: QuickOrderFooterProps) {
     setIsRequestLoading(true);
     handleClose();
 
-    const productIds = uniq(checkedArr.map(({ node }) => Number(node.productId)));
-
-    if (productIds.length === 0) {
-      snackbar.error(b3Lang('purchasedProducts.footer.selectOneItemToAdd'));
```
deov31, line 239: This validation was deleted since we are already validating the length of checkedArr in the function handleOpenBtnList

bc-alvaroolvera (approved):

PR #575, opened 2025-10-13
fix(quotes): B2B-3665 fix: add missing addressId in quote when selecting from saved addresses

bc-veer:
fix(quotes): B2B-3665 fix: add missing addressId in quote when selecting from saved addresses

Jira: [B2B-3665](https://bigcommercecloud.atlassian.net/browse/B2B-3665)

## What/Why?

Missing address Id while submitted a quote.

**Case 1: Company with default shipping and billing address**  
The system correctly passes the addressId to the quote; no changes are needed.

**Case 2: Address selection from the address book**  
When the buyer selects an address via choose from saved, I will pass the addressId to the quote.

**Case 3: User-entered or modified address**  
If the user enters or modifies the address, the addressId will be 0. This functionality is also working correctly; no changes are needed.

## Rollout/Rollback
Revert

## Testing
Added tests to cover the all three scenarios
Screenshot 2025-10-13 at 3 38 46 pm

[B2B-3665]: https://bigcommercecloud.atlassian.net/browse/B2B-3665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/QuoteDraft/index.test.tsx
+++ b/apps/storefront/src/pages/QuoteDraft/index.test.tsx
@@ -16,6 +16,7 @@ import {
   startMockServer,
   stringContainingAll,
   userEvent,
+  waitFor,
   waitForElementToBeRemoved,
   within,
 } from 'tests/test-utils';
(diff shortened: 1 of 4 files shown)


apps/storefront/src/pages/QuoteDraft/index.tsx
```diff
@@ -498,6 +503,47 @@ function QuoteDraft({ setOpenPage }: PageProps) {
...
+   *
+   * @param address - Address with optional selectedAddress for comparison
+   * @returns Cloned address, potentially with reused addressId if no changes detected
+   *
+   * Cases:
+   * 1. No Selected Address: return cloned address
+   * 2. Selected Address copy exists and address unchanged: return cloned address with addressId
+   * 3. Selected Address copy exists but address changed: return cloned address without addressId
+   */
+  const cloneAddressWithId = (address: AddressWithSelected): ShippingAddress | BillingAddress => {
```
icatalina:
```suggestion
  const cloneAddressWithId = ({ selectedAddress, ...address }: AddressWithSelected): ShippingAddress | BillingAddress => {
```

seems like something like this would have been easier

apps/storefront/src/pages/QuoteDraft/index.test.tsx
```diff
@@ -3854,4 +3855,263 @@ describe('when the user is a B2B customer', () => {
...
+            shippingAddress,
+          },
+        },
+      },
+    });
+
+    const createQuoteMutation = vi.fn();
+
+    beforeEach(() => {
+      vi.clearAllMocks();
```
icatalina: clearAllMocks is performed automatically before each test by the vitest setup.

apps/storefront/src/pages/QuoteDraft/index.test.tsx
```diff
@@ -3854,4 +3855,263 @@ describe('when the user is a B2B customer', () => {
...
+
+      const billingFields = screen.getByRole('group', { name: 'Billing' });
+      await userEvent.click(within(billingFields).getByText('Choose from saved'));
+      await waitFor(() =>
+        expect(screen.getByRole('heading', { name: 'Choose from saved' })).toBeVisible(),
+      );
+      await userEvent.click(screen.getByRole('button', { name: 'Choose address' }));
+
+      await userEvent.click(screen.getByRole('button', { name: 'Save info' }));
+      await userEvent.click(await screen.findByRole('button', { name: 'Submit' }));
```
icatalina: I'm gonna imagine the submit button is available immediately, so no need to use `findBy`, `getBy` should suffice.

apps/storefront/src/pages/QuoteDraft/index.test.tsx
```diff
@@ -3854,4 +3855,263 @@ describe('when the user is a B2B customer', () => {
...
+          }),
+        ),
+        graphql.query('getQuoteExtraFields', () =>
+          HttpResponse.json({ data: { quoteExtraFieldsConfig: [] } }),
+        ),
+        graphql.query('GetQuoteInfoB2B', () =>
+          HttpResponse.json({ data: { quote: { id: '272989', quoteNumber: '911911' } } }),
+        ),
+        http.post('*/api/v2/extra-fields/quote/validate', () => HttpResponse.json({ code: 200 })),
+        graphql.mutation('CreateQuote', async ({ request }: { request: Request }) => {
```
icatalina: 🤔 I don't think you need these types:


```suggestion
        graphql.mutation('CreateQuote', async ({ request }) => {
```

apps/storefront/src/pages/QuoteDraft/index.test.tsx
```diff
@@ -3854,4 +3855,263 @@ describe('when the user is a B2B customer', () => {
...
+      );
+      await userEvent.click(screen.getByRole('button', { name: 'Choose address' }));
+
+      await userEvent.click(screen.getByRole('button', { name: 'Save info' }));
+      await userEvent.click(await screen.findByRole('button', { name: 'Submit' }));
+
+      await waitFor(() => expect(createQuoteMutation).toHaveBeenCalled());
+      const mutationData = createQuoteMutation.mock.calls[0][0];
+      expect(mutationData).toContain(BILLING_ADDRESS_ID);
+      expect(mutationData).not.toContain(SHIPPING_ADDRESS_ID);
```
icatalina: we are trying to move away from expecting the API mocks, the expectation should be around user facing features. eg. a success message/an address name in the UI, etc

These "expect" are normally setup with a `when()` clause, making sure the response is only correct in the case of a successful submission.

```tsx
when(createQuoteMutation)
  .calledWith(expect.stringContaining(BILLING_ADDRESS_ID)
  .thenReturn(buildQuoteCreateResponseWith({
      data: {
        quoteCreate: { quote: { id: 123, createdAt: '1245' } },
      },
    }));
```

icatalina (commented):
Hey @bc-veer , I had a chance to review this, I know it is merged, but I feel like the comments are valuable as something to keep in mind going forward. Please have a look and let me know what you think 🙇

apps/storefront/src/pages/QuoteDraft/index.test.tsx
bc-veer: we are fetching query out of request and if i remove the type then i am getting eslint issue.
I think it is good to type

icatalina: you can use `query` to fetch the query from the request. This is mistyped:

```tsx
graphql.mutation('CreateQuote', async ({ query }) => {
  createQuoteMutation(query);
```

apps/storefront/src/pages/QuoteDraft/index.test.tsx
icatalina: The when function is used to remove the need for expections on spies. If you only return the correct value for the correct input, your expects are about user-facing value.

As I said on my first comment, we are trying to avoid using `expect().toHaveBeenCalled()`, and instead set the stubs directly to return the correct data in the right scenario.

```tsx
// Arrange
const getName = vi.fn();

const getUser = (userId) => ({ 
  name: getName(userId),
})

when(getName)
  .calledWith(1)
  .thenReturn('John');

// Act
const user = getUser(1);

// Assert
exect(user).toEqual({ name: 'John' }); // This is what the end user cares about
```

```tsx
// Arrange
const getName = vi.fn(() => 'John');

const getUser = (userId) => ({ 
  name: getName(userId),
})

// Act
const user = getUser(1);

// Assert
exect(user).toEqual({ name: 'John' }); // This is what the end user cares about

// this is something we need to remember to do, and it is of no interest to the user or the product owner. It should be part of the arrange, not the expect.
expect(getName).toHaveBeenCalledWith(1);
```

apps/storefront/src/pages/QuoteDraft/index.test.tsx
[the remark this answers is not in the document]
icatalina: If you switch to using `when`, then your mocks are not always the same, you have 3 different scenarios:

```tsx
expect(mutationData).toContain(BILLING_ADDRESS_ID);
expect(mutationData).not.toContain(SHIPPING_ADDRESS_ID);
```
```tsx
expect(mutationData).toContain(BILLING_ADDRESS_ID);
expect(mutationData).toContain(SHIPPING_ADDRESS_ID);
```
```tsx
expect(mutationData).not.toContain(BILLING_ADDRESS_ID);
expect(mutationData).not.toContain(SHIPPING_ADDRESS_ID);
```

apps/storefront/src/pages/QuoteDraft/index.tsx
bc-veer: resolved in this [PR](https://github.com/bigcommerce/b2b-buyer-portal/pull/583)

apps/storefront/src/pages/QuoteDraft/index.test.tsx
bc-veer: Noted. Thank you for the detailed info.

PR #581, opened 2025-10-15
fix(quotes): B2B-3763 fix cant add np&oos product to quote

bc-victor:
Jira: [B2B-3763](https://bigcommercecloud.atlassian.net/browse/B2B-3763 )

## What/Why?

 "Add to Quote" form in the Draft Quote UI fails to add non-purchasable and out-of-stock (np&oos) product to a quote even if the setting is one.

Issue stemmed from the QuickAdd component L166 where we don't consider this setting in the following block:
```js
        if (purchasingDisabled === '1' && type !== 'shoppingList') {
```
## Rollout/Rollback
Revert this change

## Testing

https://github.com/user-attachments/assets/30b2fb2a-1c54-412b-b589-cff2ae2ae8d2

[B2B-3763]: https://bigcommercecloud.atlassian.net/browse/B2B-3763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/ShoppingListDetails/components/QuickAdd.tsx
+++ b/apps/storefront/src/pages/ShoppingListDetails/components/QuickAdd.tsx
@@ -21,7 +21,7 @@ interface AddToListContentProps {
   level?: number;
   buttonText?: string;
   buttonLoading?: boolean;
-  type?: string;
+  type: 'shoppingList' | 'quoteDraft';
 }
 
 export default function QuickAdd(props: AddToListContentProps) {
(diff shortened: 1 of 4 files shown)


apps/storefront/src/pages/ShoppingListDetails/components/QuickAdd.tsx
```diff
@@ -163,7 +166,12 @@ export default function QuickAdd(props: AddToListContentProps) {
...
-        if (purchasingDisabled === '1' && type !== 'shoppingList') {
+        if (purchasingDisabled === '1' && !['shoppingList', 'quoteDraft'].includes(type || '')) {
+          notPurchaseSku.push(sku);
+          return;
+        }
+
+        if (purchasingDisabled === '1' && type === 'quoteDraft' && !allowAddNonPurchasableProduct) {
           notPurchaseSku.push(sku);
           return;
         }
```
icatalina: After the changes in this PR, the only two possible values for `type` are `shoppingList` and `quoteDraft`. There's no case where the value is neither.

Image

Can we adjust the types to require `type: 'shoppingList' | 'quoteDraft'` and adjust this logic accordingly?

PR #582, opened 2025-10-15
fix(company-hierarchy): B2B-3216 fix company state persistence & permissions logic 

BC-KCarr:
Jira: [B2B-3216](https://bigcommercecloud.atlassian.net/browse/B2B-3216)

## What/Why?
We have reports where merchants that are representing another company within the same hierarchy and opening the store in a new tab will not show an indicator of the company being represented but continues to show the represented company's products and pricing.

After further inspection, it appears there is also a permissions issue where users can see different customer groups products after a couple clicks in the new tab.

The issue here was that company hierarchy representation state (`selectCompanyHierarchyId`) was stored in `sessionStorage` and lost when opening new tabs.

Updated by refetching the company hierarchy data anytime the app loads and then dispatch to redux

This was done by creating `storefrontConfigWithCompanyHierarchy` that fetches both storefront config and company hierarchy data in a single request

## Rollout/Rollback
revert and redeploy

### Before
Demonstrates the issue where company representation state is lost in new tabs and permission logic fails, causing users to see unauthorized products after navigation clicks

https://github.com/user-attachments/assets/07ea3b0a-0e56-4da1-90d5-4e06e4fc96c7

### After
Shows the fix working correctly with company representation state persists across tabs and permission logic properly restricts access to authorized products only

https://github.com/user-attachments/assets/1db7deed-0d9a-4051-8242-31997310fd9e

[B2B-3216]: https://bigcommercecloud.atlassian.net/browse/B2B-3216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/utils/initializeCompanyHierarchy.ts
```diff
@@ -0,0 +1,78 @@
...
+      role === CustomerRole.SENIOR_BUYER ||
+      role === CustomerRole.JUNIOR_BUYER ||
+      role === CustomerRole.CUSTOM_ROLE
+    ) {
+      const isEnabledAccountHierarchy = await getAccountHierarchyIsEnabled();
+
+      if (isEnabledAccountHierarchy) {
+        const [{ companySubsidiaries }, { userMasqueradingCompany }] = await Promise.all([
+          getCompanySubsidiaries(),
+          getUserMasqueradingCompany(),
```
bc-micah: can we combine this into one query?

bc-micah (commented):
looking better, the videos in JIRA looked like expected behavior.

I think we can simplify the implementation a bit more - i.e don't end the masquerading automatically as we talked about yesterday, and always show the data.

apps/storefront/src/utils/initializeCompanyHierarchy.ts
BC-KCarr: sure thing, good call 👍🏾

PR #585, opened 2025-10-20
feat(quotes): B2B-3746 fix backordering validation to include variant checking

BC-KCarr:
Jira: [B2B-3746](https://bigcommercecloud.atlassian.net/browse/B2B-3746)

## What/Why?
Products with variants and unlimited backorder limits were incorrectly showing stock level warnings in the draft quote UI, even when the variant had unlimited backorder enabled.

This is because the stock warning logic was only checking product-level unlimitedBackorder settings but not variant-level unlimited backorder settings when inventoryTracking was set to variant

## Rollout/Rollback
revert and redeploy

### Before

https://github.com/user-attachments/assets/6cfb1e4d-8710-4f6a-8c11-9fe2799a468e

### After

https://github.com/user-attachments/assets/aa116a1b-4ba6-42f5-8a59-fd5e23a86a89

[B2B-3746]: https://bigcommercecloud.atlassian.net/browse/B2B-3746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/quote/components/QuoteTable.tsx
+++ b/apps/storefront/src/pages/quote/components/QuoteTable.tsx
@@ -251,10 +251,10 @@ function QuoteTable(props: ShoppingDetailTableProps) {
               warningMessage = b3Lang('quoteDraft.quoteTable.unavailable.tip');
             }
           }
-        } else {
+        } else if (isInventoryTrackingEnabled) {
           let hasUnlimitedBackorder = product.unlimitedBackorder;
 
-          if (isInventoryTrackingEnabled && product.inventoryTracking === 'variant') {
+          if (product.inventoryTracking === 'variant') {
             const currentVariant = product.variants.find(
               (variant: CustomFieldItems) => variant.sku === row.variantSku,
             );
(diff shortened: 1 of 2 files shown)


apps/storefront/src/pages/quote/components/QuoteTable.tsx
```diff
@@ -251,13 +251,27 @@ function QuoteTable(props: ShoppingDetailTableProps) {
...
-          const productStock = isInventoryTrackingEnabled ? product.availableToSell : row.quantity;
-          if (isInventoryTrackingEnabled && productStock < row.quantity) {
-            warningMessage = b3Lang('quoteDraft.quoteTable.outOfStock.tip');
-            warningDetails = b3Lang('quoteDraft.quoteTable.oosNumber.tip', {
-              qty: productStock,
-            });
+        } else {
+          let hasUnlimitedBackorder = product.unlimitedBackorder;
+
+          if (isInventoryTrackingEnabled && product.inventoryTracking === 'variant') {
```
icatalina: can you lift the check for `isInventoryTrackingEnabled`, it looks like all the conditions in this if depend on it being `true`.

icatalina (changes requested):

PR #596, opened 2025-10-28
fix(common): B2B-3719 read hash at event time

bc-coolbeans:
Jira: [3719](https://bigcommercecloud.atlassian.net/browse/B2B-3719)

## What/Why?

We changed the hashchange handler in `apps/storefront/src/App.tsx`. The old code captured the hash at mount, so after ending masquerade it compared against a stale value and closed the overlay, causing a blank page. The new handler reads `window.location.hash` at event time and only closes when the current hash is empty or `#/`. This ensures the app frame stays open and the router correctly shows `/dashboard` after ending masquerade.

## Rollout/Rollback

Rollback

## Testing

Tested on integration and can no longer see blank page.

### DIFF
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -335,9 +335,12 @@ export default function App() {
   }, [isOpen]);
 
   useEffect(() => {
-    const { hash = '' } = window.location;
-
-    const handleHashChange = () => (!hash || hash === '#/') && setOpenPage({ isOpen: false });
+    const handleHashChange = () => {
+      const { hash } = window.location;
+      if (!hash || hash === '#/') {
+        setOpenPage({ isOpen: false });
+      }
+    };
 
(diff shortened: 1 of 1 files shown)


apps/storefront/src/App.tsx
```diff
@@ -335,9 +335,12 @@ export default function App() {
   }, [isOpen]);
 
   useEffect(() => {
-    const { hash = '' } = window.location;
-
-    const handleHashChange = () => (!hash || hash === '#/') && setOpenPage({ isOpen: false });
+    const handleHashChange = () => {
+      const { hash = '' } = window.location;
+      if (!hash || hash === '#/') {
```
icatalina: I know this is existing code, but I think we can omit the `= ''` since `hash` is guaranteed to be a string:

image

bc-veer: Sure. Thank you @icatalina

PR #602, opened 2025-10-28
fix(quotes): B2B-3814 Add variant-level available to sell checks

BC-MLombardi:
Jira: [B2B-3814](https://bigcommercecloud.atlassian.net/browse/B2B-3814)

## What/Why?

The stock warning messages in the products table were not properly accounting for products with variant-level inventory tracking. This PR fixes that by using the variant-level available to sell quantity instead of the product-level available to sell when variant-level inventory is enabled.

## Rollout/Rollback

Revert and redeploy.

## Testing

Before:

https://github.com/user-attachments/assets/c7295e19-df36-46e8-96e2-fe2593391c74

After:

https://github.com/user-attachments/assets/16b65e37-7251-45ae-afe3-d6be0a5461c2

[B2B-3814]: https://bigcommercecloud.atlassian.net/browse/B2B-3814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/QuoteDraft/index.test.tsx
+++ b/apps/storefront/src/pages/QuoteDraft/index.test.tsx
@@ -1331,431 +1331,607 @@ describe('when the user is a B2B customer', () => {
       'B2B-3318.move_stock_and_backorder_validation_to_backend': true,
     };
 
-    it('does not show stock warning in the product table', async () => {
-      server.use(
-        graphql.query('Countries', () => HttpResponse.json({ data: { countries: [fakeCountry] } })),
-        graphql.query('Addresses', () =>
-          HttpResponse.json({ data: { addresses: { totalCount: 0, edges: [] } } }),
-        ),
-        graphql.query('getQuoteExtraFields', () =>
-          HttpResponse.json({ data: { quoteExtraFieldsConfig: [] } }),
-        ),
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/QuoteDraft/index.test.tsx
```diff
@@ -1331,431 +1331,607 @@ describe('when the user is a B2B customer', () => {
...
-      renderWithProviders(<QuoteDraft setOpenPage={vi.fn()} />, {
-        preloadedState: {
-          ...preloadedState,
-          quoteInfo,
-          global: buildGlobalStateWith({
-            blockPendingQuoteNonPurchasableOOS: { isEnableProduct: true },
-            featureFlags,
-          }),
-        },
+        expect(within(productTable).queryAllByText('TBD')).toHaveLength(0);
```
icatalina: 🤔 should this be the opposite? check that the price is present?

I'm ok if we want ot make sure TBD is not present, but I feel we might want to ensure the prices are shown.


```tsx
it('shows the price when unlimited backorder is true', async () => {
```

PR #606, opened 2025-10-29
feat(company-hierarchy): B2B-3216 fetch company hierarchy info with storeconfig

BC-KCarr:
Jira: [B2B-3216](https://bigcommercecloud.atlassian.net/browse/B2B-3216)

## What/Why?
We have reports where merchants that are representing another company within the same hierarchy and opening the store in a new tab will not show an indicator of the company being represented but continues to show the represented company's products and pricing.

After further inspection, it appears there is also a permissions issue where users can see different customer groups products after a couple clicks in the new tab.

The issue here was that company hierarchy representation state `selectCompanyHierarchyId` was stored in sessionStorage and lost when opening new tabs.

Updated by refetching the company hierarchy data anytime the app loads and then dispatch to redux

This was done by creating `storefrontConfigWithCompanyHierarchy` that fetches both storefront config and company hierarchy data in a single request

## Rollout/Rollback
turning off the `B2B-3817.disable_masquerading_cleanup_on_login` LD feature flag should revert most of the logic added in this PR

### Before
company hierarch masquerading session does not continue when opening new tab

https://github.com/user-attachments/assets/07ea3b0a-0e56-4da1-90d5-4e06e4fc96c7

### After
company hierarchy masquerading session now continuing as expected when opening new tab

https://github.com/user-attachments/assets/dc7a173a-91c8-4e71-8fd5-1221b12a5c5c

user with no company hierarchy permissions are now able to load the app due to request not being made unless user has permissions

Screenshot 2025-10-29 at 9 38 03 AM

add to quote button no longer hidden for guest users 

Screenshot 2025-10-29 at 7 01 39 AM
 

[B2B-3216]: https://bigcommercecloud.atlassian.net/browse/B2B-3216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/shared/service/b2b/graphql/global.ts
```diff
@@ -171,6 +171,47 @@ const storefrontConfig = () => `{
...
+		parentCompanyName
+		channelFlag
+	}
+	userMasqueradingCompany @include(if: $includeCompanyHierarchy) {
+		companyId
+		companyName
+		bcId
+	}
+}`;
+
```
BC-KCarr: updated the combined query to conditionally request `companySubsidiaries` and `userMasqueradingCompany` only if user has permissions to company hierarchy info

apps/storefront/src/shared/service/b2b/graphql/global.ts
```diff
@@ -171,6 +171,47 @@ const storefrontConfig = () => `{
...
+			recentlyViewed,
+			returns,
+			shoppingLists,
+			tradeProfessionalApplication,
+			userManagement,
+			wishLists,
+		}
+		configId,
+	}
+	companySubsidiaries @include(if: $includeCompanyHierarchy) {
```
icatalina: very nice, didn't know this was a thing 👍

apps/storefront/src/components/layout/B3CompanyHierarchy.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+      renderWithProviders(<B3CompanyHierarchy />, {
+        preloadedState,
+        initialGlobalContext: {},
+      });
+
+      const dropdownButton = screen.getByRole('button');
+      expect(dropdownButton).toBeInTheDocument();
+
+      expect(dropdownButton).toHaveTextContent('Parent Company');
```
icatalina:
```tsx
const dropdownButton = screen.getByRole('button', { name: 'Parent Company' });
expect(dropdownButton).toBeInTheDocument();
```

if that doesn't work, then simply omit the `toBeInTheDocument`:


```tsx
const dropdownButton = screen.getByRole('button');

expect(dropdownButton).toHaveTextContent('Parent Company');
```

`getBy*` will blow up if the element is not in the document, so there's no need to ensure it is in. We add the `.toBeInTheDocument` when we don't need to make any other check on the element to avoid leaving the variable "unused".

icatalina (approved):

PR #616, opened 2025-11-06
fix(quotes): B2B-3925 load quotes details with backend validations

deov31:
## What/Why?

When loading a quote with the backend validations feature flag enabled we are still using front end validations in the QuoteDetail component to know if there's any out of stock or non purchasable error in the product, so we need to differentiate using the feature flag when is required to use front end or backend validations.

For products with a warning status, we should be able to checkout the quote and see no errors in the page.
For products with a error status, we should see a snackbar for each failed product and proceed to checkout button should be disable.

## Rollout/Rollback
Revert

## Testing

https://github.com/user-attachments/assets/36f0da33-6f1a-466c-aed6-8d41b5e404c0

https://github.com/user-attachments/assets/7c024b5a-5c69-4fc0-9f04-915febb48000

https://github.com/user-attachments/assets/e64f216a-7352-42da-be72-ec5591c2d23d

### DIFF
--- a/apps/storefront/src/pages/QuoteDetail/index.test.tsx
+++ b/apps/storefront/src/pages/QuoteDetail/index.test.tsx
@@ -123,71 +123,10 @@ const buildProductSearchWith = builder<ProductSearch>(() => ({
   isPriceHidden: faker.datatype.boolean(),
 }));
 
-const buildProductSearchDataWith = builder<B2BProducts>(() => ({
+const buildProductSearchResponseWith = builder<B2BProducts>(() => ({
   data: {
     productsSearch: Array.from({ length: faker.number.int({ min: 0, max: 10 }) }, () => ({
-      id: faker.number.int(),
-      name: faker.commerce.productName(),
-      sku: faker.string.uuid(),
-      costPrice: faker.commerce.price(),
-      inventoryLevel: faker.number.int(),
-      inventoryTracking: faker.helpers.arrayElement(['none', 'simple', 'complex']),
-      availability: faker.helpers.arrayElement(['available', 'unavailable']),
-      orderQuantityMinimum: faker.number.int(),
-      orderQuantityMaximum: faker.number.int(),
-      variants: Array.from({ length: faker.number.int({ min: 0, max: 10 }) }, () => ({
-        variant_id: faker.number.int(),
-        product_id: faker.number.int(),
-        sku: faker.string.uuid(),
-        option_values: Array.from({ length: faker.number.int({ min: 0, max: 10 }) }, () => ({
-          id: faker.number.int(),
-          label: faker.commerce.productAdjective(),
-          option_id: faker.number.int(),
-          option_display_name: faker.commerce.productAdjective(),
-        })),
-        calculated_price: parseFloat(faker.commerce.price()),
-        image_url: faker.image.url(),
-        has_price_list: faker.datatype.boolean(),
-        bulk_prices: [],
-        purchasing_disabled: faker.datatype.boolean(),
-        cost_price: parseFloat(faker.commerce.price()),
-        inventory_level: faker.number.int(),
-        bc_calculated_price: {
-          as_entered: parseFloat(faker.commerce.price()),
-          tax_inclusive: parseFloat(faker.commerce.price()),
-          tax_exclusive: parseFloat(faker.commerce.price()),
-          entered_inclusive: faker.datatype.boolean(),
-        },
-      })),
-      currencyCode: faker.finance.currencyCode(),
-      imageUrl: faker.image.url(),
-      modifiers: [],
-      options: Array.from({ length: faker.number.int({ min: 0, max: 10 }) }, () => ({
-        option_id: faker.number.int(),
-        display_name: faker.commerce.productAdjective(),
-        sort_order: faker.number.int(),
-        is_required: faker.datatype.boolean(),
-      })),
-      optionsV3: Array.from({ length: faker.number.int({ min: 0, max: 10 }) }, () => ({
-        id: faker.number.int(),
-        product_id: faker.number.int(),
-        name: faker.commerce.productAdjective(),
-        display_name: faker.commerce.productAdjective(),
-        type: faker.helpers.arrayElement(['rectangles', 'swatch', 'dropdown']),
-        sort_order: faker.number.int(),
-        option_values: Array.from({ length: faker.number.int({ min: 0, max: 10 }) }, () => ({
-          id: faker.number.int(),
-          label: faker.commerce.productAdjective(),
-          sort_order: faker.number.int(),
-          value_data: null,
-          is_default: faker.datatype.boolean(),
-        })),
-        config: [],
-      })),
-      channelId: [],
-      productUrl: faker.internet.url(),
-      taxClassId: faker.number.int(),
-      isPriceHidden: faker.datatype.boolean(),
+      ...buildProductSearchWith('WHATEVER_VALUES'),
     })),
   },
 }));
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -228,7 +238,25 @@ function QuoteDetail() {
...
+      errors.forEach((error) => {
+        if (error.type === 'validation') {
+          snackbar.error(error.message);
+        }
+      });
+
+      setShouldHideCheckoutButton(true);
+    }
+
+    setIsHideQuoteCheckout(false);
```
bc-micah: these seem to be very similar - any reason we couldn't rely on the existing variable?

deov31: i added a new state variable because the current `isHideQuoteCheckout` blocks the checkout to be performed if there's any error in a product but doesn't actually disabled to button or anything. If i use the same one i would end up disabling the button even when only using front end validations, and i didn't want change anything from the current behavior. I agree that it does feels duplicated but is easier to manage some states separately for the different flows.

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -577,8 +633,16 @@ function QuoteDetail() {
...
   const isEnableProductShowCheckout = () => {
     if (isEnableProduct) {
-      if (isHandleApprove && isHideQuoteCheckout) return true;
-      if (!isHideQuoteCheckout) return true;
+      if (isMoveStockAndBackorderValidationToBackend) {
+        if (
+          (isHandleApprove && shouldHideCheckoutButtonWhenBackendValidations) ||
+          !shouldHideCheckoutButtonWhenBackendValidations
+        )
+          return true;
```
icatalina: this should be equivalent to:


```tsx
if (isHandleApprove || !shouldHideCheckoutButtonWhenBackendValidations) {
  return true;
}
```

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -192,8 +206,20 @@ function QuoteDetail() {
 
   const [quoteCheckoutLoading, setQuoteCheckoutLoading] = useState<boolean>(false);
 
+  const [
+    shouldHideCheckoutButtonWhenBackendValidations,
+    setShouldHideCheckoutButtonWhenBackendValidations,
+  ] = useState<boolean>(false);
+
+  const [isValidationLoading, setIsValidationLoading] = useState<boolean>(false);
```
icatalina: These two are probably better of as:

```tsx
const [isButtonDisabled, setIsButtonDisabled] = useState(false);
```

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -269,28 +351,40 @@ function QuoteDetail() {
...
-          }),
-        );
-
-      if (nonPurchasable)
-        snackbar.error(
-          b3Lang('quoteDetail.message.nonPurchasable', {
-            ProductName: nonPurchasable,
-          }),
-        );
+    if (!isMoveStockAndBackorderValidationToBackend) {
```
icatalina: follow the same pattern as with `validateQuoteProducts`, two methods (even if one is simply a `() => true`

apps/storefront/src/utils/validateProducts.ts
```diff
@@ -58,7 +82,7 @@ export const validateProducts = async (products: CustomFieldItems[]): Promise<Va
   settledResults.forEach((result, index) => {
     // Network or unexpected error
     if (result.status === 'rejected') {
-      const { productName } = products[index].node;
+      const { productName } = products[index];
```
BC-MLombardi: This should be `productsList` shouldn't it? `products` is the original data with the `node` property

```suggestion
      const { productName } = productsList[index];
```

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -139,6 +139,62 @@ function useData() {
...
+        alignItems: 'flex-end',
+        flexDirection: 'column',
+      }
+    : {
+        alignItems: 'center',
+      };
+};
+
+function Footer({ children, isAgenting }: { children: React.ReactNode; isAgenting: boolean }) {
+  const [isMobile] = useMobile();
```
BC-MLombardi: `isMobile` can be passed down as a prop here.

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -228,7 +290,21 @@ function QuoteDetail() {
     });
   }, [isB2BUser, quoteDetail, selectCompanyHierarchyId, purchasabilityPermission]);
 
-  useEffect(() => {
+  const quoteDetailBackendEndValidations = async () => {
```
BC-MLombardi:
```suggestion
  const quoteDetailBackendValidations = async () => {
```

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -269,11 +345,31 @@ function QuoteDetail() {
       oos: oosErrorList,
       nonPurchasable: nonPurchasableErrorList,
     });
+  };
+
+  const validateQuoteProducts = isMoveStockAndBackorderValidationToBackend
+    ? quoteDetailBackendEndValidations
+    : quoteDetailFrontEndValidations;
+
+  const proceedingCheckoutFnBackendEndFlow = () => {
```
BC-MLombardi:
```suggestion
  const proceedingCheckoutFnBackendFlow = () => {
```

apps/storefront/src/pages/QuoteDetail/index.tsx
deov31: https://github.com/bigcommerce/b2b-buyer-portal/pull/616#discussion_r2511135458

apps/storefront/src/pages/QuoteDetail/index.test.tsx
```diff
@@ -298,4 +305,265 @@ describe('when the user is a B2B customer', () => {
...
+        },
+      },
+    });
+
+    await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
+
+    expect(await screen.findByRole('button', { name: /PROCEED TO CHECKOUT/i })).toBeInTheDocument();
+    await userEvent.click(await screen.findByRole('button', { name: /PROCEED TO CHECKOUT/i }));
+    expect(
+      await screen.findAllByText('A product with the id of 123 does not have sufficient stock'),
```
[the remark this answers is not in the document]
icatalina: yep, that makes sense, it is just to make sure it is more obvious for the next person reading the test.

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -358,10 +358,10 @@ function QuoteDetail() {
         snackbar.error(error.message);
       });
 
-      return false;
+      return true;
     }
 
-    return true;
+    return false;
```
deov31: hey @icatalina, i need to change this because in this line https://github.com/bigcommerce/b2b-buyer-portal/pull/616/files#diff-518d8baa6ceb5f14406981623bf714117821b4adf07267757822901cd30e8a64R28 if i return true here, it will stop the process to checkout. But also it doesn't have much sense returning true when we have errors, should I rename this function? Any recommendation here?

deov31: I made a change to the naming of these "proceedToCheckout..." function to `hasQuoteValidationErrorsXFlow` since it seems to make more sense with the return value we use to stop the checkout flow, let me know what you think.

PR #618, opened 2025-11-07
fix(common): B2B-3939 Limit vitests resources

icatalina:

## What/Why?
Limit vitests resources
Vitests seems to be exhausting resources in the machine. Strictly limiting them so tests pass.

## Rollout/Rollback
Revert

## Testing
Hopefully this PR passes all the tests and lets me merge to master 😅

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,8 @@ jobs:
   test:
     executor: node
     resource_class: medium+
+    environment:
+      MAX_WORKERS: 3 # Number of CPUs in resource_class
     steps:
       - ci/pre-setup
       - install-dependencies
@@ -67,7 +69,7 @@ jobs:
 
   lint:
     executor: node
-    resource_class: medium+
+    resource_class: medium
     steps:
       - ci/pre-setup
       - install-dependencies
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/QuickOrder/index.main.test.tsx
```diff
@@ -3933,7 +3943,9 @@ describe('When backend validation feature flag is on', () => {
         expect(screen.getByText('ERROR-SKU-123')).toBeInTheDocument();
       });
 
-      const addToCartButton = screen.getByRole('button', { name: /Add 1 products to cart/i });
+      const addToCartButton = await screen.findByRole('button', {
```
icatalina: This selector was failing on one of the runs, replaced with `findByRole` to see if it is more stable.

.circleci/config.yml
```diff
@@ -67,7 +69,7 @@ jobs:
 
   lint:
     executor: node
-    resource_class: medium+
+    resource_class: medium
```
icatalina: These jobs don't need a medium+, and this one is cheaper.

.circleci/config.yml
```diff
@@ -45,6 +45,8 @@ jobs:
   test:
     executor: node
     resource_class: medium+
+    environment:
+      MAX_WORKERS: 3 # Number of CPUs in resource_class
```
icatalina: `medium+` provides 3 CPUs, using an ENV variable so the number of CPUs is defined close to the resource_class

bc-micah (approved):

PR #624, opened 2025-11-11
fix(quick-order): B2B-3941 Treat validation warnings as errors when adding to cart

BC-MLombardi:
Jira: [B2B-3941](https://bigcommercecloud.atlassian.net/browse/B2B-3941)

## What/Why?

When the non-purchasable & out of stock setting is enabled, validation warnings are treated as successful. As a result, when trying to add multiple products to a cart, the request to the cart API included non-purchasable/out of stock products, causing it to fail and preventing any requested products from being added.

This PR fixes this by modifying the `validateProducts` function to accept an optional `destination` string. When the string is set to `quote` (the default), we continue to treat warnings as successful (which is the desired behavior for quotes). When it's set to `cart`, we treat warnings as errors, ensuring that non-purchasable/out of stock products are filtered out from cart API request.

## Rollout/Rollback

Revert and redeploy.

## Testing

Added new unit tests.

Before:

https://github.com/user-attachments/assets/45532cda-2396-4c72-a8f0-1c535c070ee2

After:

https://github.com/user-attachments/assets/49a8e002-4223-46d5-90cd-1d37a270c997

Products with validation warnings still get added to quotes:

https://github.com/user-attachments/assets/77d333d7-5afa-4f64-bfe3-21582aaa71c0

[B2B-3941]: https://bigcommercecloud.atlassian.net/browse/B2B-3941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
+++ b/apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
@@ -335,7 +335,9 @@ export default function QuickAdd(props: AddToListContentProps) {
 
     const productsToValidate = mapCatalogToValidationPayload(variantInfoList, skuValue);
 
-    const { validProducts, errors } = await validateProducts(productsToValidate);
+    // When adding products to a cart, warnings need to be treated as errors.
+    // We are doing this in the frontend for now, but it will be moved to the backend in the future.
+    const { validProducts, errors } = await validateProducts(productsToValidate, 'cart');
 
     const productItems = mergeValidatedWithCatalog(validProducts, variantInfoList);
 
--- a/apps/storefront/src/pages/QuickOrder/index.main.test.tsx
+++ b/apps/storefront/src/pages/QuickOrder/index.main.test.tsx
@@ -2843,6 +2843,178 @@ describe('When backend validation feature flag is on', () => {
     });
   });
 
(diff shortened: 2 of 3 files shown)


apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
```diff
@@ -335,7 +335,9 @@ export default function QuickAdd(props: AddToListContentProps) {
 
     const productsToValidate = mapCatalogToValidationPayload(variantInfoList, skuValue);
 
-    const { validProducts, errors } = await validateProducts(productsToValidate);
+    // When adding products to a cart, warnings need to be treated as errors.
+    // We are doing this in the frontend for now, but it will be moved to the backend in the future.
+    const { validProducts, errors } = await validateProducts(productsToValidate, 'cart');
```
deov31: Question, we can also add products to the cart from the purchased products table, do we need to also pass the "cart" prop to this function?

BC-KCarr (approved):
💚 Tested LGTM. 

As @deov31 mentioned, we should decide whether to follow up with the same frontend patch when adding multiple products to the cart from the purchased products table

BC-MLombardi:
Closing in favor of #628

PR #627, opened 2025-11-12
refactor(quick-order): B2B-3941 Separate add to cart functions

BC-MLombardi:
Jira: [B2B-3941](https://bigcommercecloud.atlassian.net/browse/B2B-3941)

## What/Why?

This PR separates the quick order add to cart functions into two:
- Adding a single product to a cart (which is handled via search in the quick order pad)
- Adding multiple products to a cart (handled by the quick add/csv upload in the quick order pad)

This allowed me to move the function that adds multiple products to a cart into the `QuickAdd` component, which will enable me to call the new shared validate/add to cart function easily once it's created.

## Rollout/Rollback

Revert and redeploy.

## Testing

https://github.com/user-attachments/assets/be4d62f2-9797-479c-8ebe-362a568ca18d

[B2B-3941]: https://bigcommercecloud.atlassian.net/browse/B2B-3941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
+++ b/apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
@@ -5,11 +5,14 @@ import { Box, Grid, Typography } from '@mui/material';
 import CustomButton from '@/components/button/CustomButton';
 import B3ControlTextField from '@/components/form/B3ControlTextField';
 import B3Spin from '@/components/spin/B3Spin';
+import { CART_URL } from '@/constants';
 import { useBlockPendingAccountViewPrice, useFeatureFlags } from '@/hooks';
 import { useB3Lang } from '@/lib/lang';
 import { getVariantInfoBySkus } from '@/shared/service/b2b';
 import { useAppSelector } from '@/store';
 import { snackbar } from '@/utils';
+import b3TriggerCartNumber from '@/utils/b3TriggerCartNumber';
+import { createOrUpdateExistingCart } from '@/utils/cartUtils';
 import { validateProducts, ValidationError } from '@/utils/validateProducts';
 
 import { SimpleObject } from '../../../types';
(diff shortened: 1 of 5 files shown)


apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
```diff
@@ -344,6 +342,27 @@ export default function QuickAdd(props: AddToListContentProps) {
...
+          onClick: () => {
+            if (window.b2b.callbacks.dispatchEvent('on-click-cart-button')) {
+              window.location.href = CART_URL;
+            }
+          },
+        },
+      });
+    }
+
+    b3TriggerCartNumber();
```
BC-KCarr: Should this be called even when the cart operation fails?

apps/storefront/src/pages/QuickOrder/components/SearchProduct.tsx
```diff
@@ -105,18 +105,18 @@ export default function SearchProduct({ addToList }: SearchProductProps) {
...
     } finally {
       setIsLoading(false);
     }
   };
 
-  const handleProductListAddToList = async (products: CustomFieldItems[]) => {
-    await handleAddToListClick(products);
+  const handleProductListAddToList = async (product: CustomFieldItems) => {
+    await handleAddToListClick(product);
   };
```
icatalina: do we need this indirection? can't we just pass `handleAddToListClick` directly?

icatalina (approved):

apps/storefront/src/pages/QuickOrder/components/QuickAdd.tsx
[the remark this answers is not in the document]
BC-MLombardi: As long as at least one product gets added to the cart, we need to call it.

I'll do more investigating after I create the new shared function, but I'll leave this as for now.

PR #639, opened 2025-11-15
fix(quotes): B2B-3964 showing TBD only for products with warning or erros

deov31:
## What/Why?
The quote detail summary is currently showing TBD for all prices even after submission of the quote since we are relying only in the front end flow validations to pass the "showPrice" prop to the quote summary.

When having a purchasable product we should show TBD only on Shipping, only after revision we should see the value of the shipping.

When having out of stock/non purchasable products, we should display TBD on all prices and only render the prices after revision of the quote. 

## Rollout/Rollback
Revert

## Testing

https://github.com/user-attachments/assets/493d2ba0-260c-498f-830f-d176e2a03c33

### DIFF
--- a/apps/storefront/src/pages/QuoteDetail/index.tsx
+++ b/apps/storefront/src/pages/QuoteDetail/index.tsx
@@ -226,10 +226,11 @@ function QuoteDetail() {
   const [quoteDetail, setQuoteDetail] = useState<any>({});
   const [productList, setProductList] = useState<any>([]);
   const [fileList, setFileList] = useState<any>([]);
-  const [isHandleApprove, setHandleApprove] = useState<boolean>(false);
+  const [quoteReviewedBySalesRep, setQuoteWasReviewedBySalesRep] = useState<boolean>(false);
 
   const [isHideQuoteCheckout, setIsHideQuoteCheckout] = useState<boolean>(true);
   const [quoteValidationErrors, setQuoteValidationErrors] = useState<ValidatedProductError[]>([]);
+  const [quoteHasWarnings, setQuoteHasWarnings] = useState<boolean>(true);
 
   const [quoteSummary, setQuoteSummary] = useState<any>({
     originalSubtotal: 0,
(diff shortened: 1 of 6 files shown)


apps/storefront/src/pages/quote/components/QuoteDetailSummary.tsx
```diff
@@ -191,8 +200,10 @@ export default function QuoteDetailSummary({
                     margin: '4px 0',
                   }}
                 >
-                  <Typography>{shippingAndTax.taxText}</Typography>
-                  <Typography>{showPrice(shippingAndTax.taxVal)}</Typography>
+                  <Typography id="quote-summary-tax">{shippingAndTax.taxText}</Typography>
+                  <Typography aria-labelledby="quote-summary-tax">
```
bc-veer: As part of A11Y best practices, we should avoid using `aria-labelledby` on non-editable or non-interactive elements. Instead, content should be expressed using native semantics, such as proper heading tags, labels, or descriptive text directly in the DOM.

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -250,13 +251,18 @@ function QuoteDetail() {
...
+  const [shouldHidePrices, setShouldHidePrices] = useState<boolean>(true);
+
   const location = useLocation();
 
   const featureFlags = useFeatureFlags();
 
   const isMoveStockAndBackorderValidationToBackend =
     featureFlags['B2B-3318.move_stock_and_backorder_validation_to_backend'];
 
+  const isAutoQuotingEnabled =
```
bc-veer: Since this lookup runs on every render, consider wrapping it in a useMemo—especially if quoteConfig is large or the component re-renders often. Memoizing will avoid recalculating the find logic unnecessarily.

apps/storefront/src/pages/quote/components/QuoteDetailSummary.tsx
bc-micah: Good point - the use of `aria-labeledby` was my suggestion, but seems like it should also be only used for interactive elements (similar to actual labels).

The main goal of this change was trying to help remove arbritrary `title=` and `data-test-id` as we want to try make the test use more of the visible text/surface area as much as possible. So maybe we can try selecting the row based on the text using a matcher, and then verifying the row also has TBD using a matcher?

```javascript
// get the summary element
const summaryElement = screen.getByTestId('quote-summary');
const withinSummary = within(parentElement);
withinSummary.getByRole('row', { name: /Estimated Tax/ }).toHaveTextContent(/TBD/);
```

icatalina (approved):

PR #644, opened 2025-11-19
fix(quotes): B2B-3474 clean sku for pdp add to quote

bc-victor:
Jira: [B2B-3474](https://bigcommercecloud.atlassian.net/browse/B2B-3474)

## What/Why?

Fixing a bug where items with an SKU string including an "&" symbol could not be added to quote. This fix replaces `innerHtml` for `textContent` in order to correctly read strings without html encodings.

## Rollout/Rollback

Feature is behind ff `B2B-3474.get_sku_from_pdp_with_text_content`.

## Testing

Added unit tests for this fix.

Before:

https://github.com/user-attachments/assets/bc49dc55-a624-45f6-bc74-a8285b1c9eca

After

https://github.com/user-attachments/assets/28576c45-5010-4ccd-8f1b-9be16128ee3b

[B2B-3474]: https://bigcommercecloud.atlassian.net/browse/B2B-3474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/hooks/dom/utils.test.ts
+++ b/apps/storefront/src/hooks/dom/utils.test.ts
@@ -756,5 +756,52 @@ describe('addProductFromProductPageToQuote', () => {
       expect(globalSnackbar.success).toHaveBeenCalled();
       expect(addQuoteDraftProduce).toHaveBeenCalled();
     });
+
+    it('correctly retrieves the SKU of a product with special characters and adds it to cart', async () => {
+      createDOM({ productId: 123, qty: 1, sku: 'A&B' });
+
+      when(searchProduct)
+        .calledWith(expect.stringContaining('productIds: [123]'))
+        .thenReturn(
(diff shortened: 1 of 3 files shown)


apps/storefront/src/hooks/dom/utils.test.ts
```diff
@@ -756,5 +756,50 @@ describe('addProductFromProductPageToQuote', () => {
       expect(globalSnackbar.success).toHaveBeenCalled();
       expect(addQuoteDraftProduce).toHaveBeenCalled();
     });
+
+    it('correctly decodes HTML entities in SKU when matching variants', async () => {
```
icatalina: the decoding is an implementation detail, this is more like handling SKUs with especial characters in them.

PR #651, opened 2025-11-20
fix(quick-order): B2B-3978 passing withModifiers to bulk upload csv

bc-victor:
Jira: [B2B-3978](https://bigcommercecloud.atlassian.net/browse/B2B-3978)

## What/Why?

Fixing a bug where passing an SKU of a product with required modifiers would break the CSV upload in quick order form by not rendering the error report due to an unknown error. The reason being not passing the `withModifiers` flag. 

This flag is used by the api to validate modifiers for every single product. 

## Rollout/Rollback

This fix is under the feature flag `B2B-3978.pass_with_modifiers_to_product_upload`

## Testing

Before:
Error has no product name, no error report snackbar is shown.
https://github.com/user-attachments/assets/4cc267d2-f489-409a-8ccd-a05fe28a4d28

After:

https://github.com/user-attachments/assets/2531eeaa-9a4f-4b88-bf8c-8556eebdbb1c

[B2B-3978]: https://bigcommercecloud.atlassian.net/browse/B2B-3978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/pages/QuickOrder/components/QuickOrderPad.tsx
+++ b/apps/storefront/src/pages/QuickOrder/components/QuickOrderPad.tsx
@@ -32,6 +32,8 @@ export default function QuickOrderPad() {
   const featureFlags = useFeatureFlags();
   const backendValidationEnabled =
     featureFlags['B2B-3318.move_stock_and_backorder_validation_to_backend'] ?? false;
+  const passWithModifiersToProductUpload =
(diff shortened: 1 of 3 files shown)


icatalina (approved):
do we need some test to ensure this behaviour stays correct?

PR #652, opened 2025-11-21
fix(login): B2B-3957 Added restrictions for InActive and Pending company users login

bc-veer:
Jira: [B2B-3957](https://bigcommercecloud.atlassian.net/browse/B2B-3957)

## What/Why?

Restrict Access and Masquerade Actions for Users of Inactive and pending Companies.
- Implement restrictions for users belonging to inactive companies from bc customers page
- Users cannot log in if their company is inactive.
- Users cannot start `masquerade` or `switch to child company`.

## Rollout/Rollback
Revert

## Testing
Added unit tests to validate the usecase

[B2B-3957]: https://bigcommercecloud.atlassian.net/browse/B2B-3957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -194,7 +194,7 @@ export default function App() {
       if (!customerId) {
         const info = await getCurrentCustomerInfo().catch((error) => {
           if (isCompanyError(error)) {
-            gotoPage(`/login?loginFlag=${error.reason}&showTip=false`);
+            gotoPage(`/login?loginFlag=${error.reason}`);
           }
         });
         if (info) {
--- a/apps/storefront/src/pages/CompanyHierarchy/components/HierarchyDialog.tsx
+++ b/apps/storefront/src/pages/CompanyHierarchy/components/HierarchyDialog.tsx
@@ -19,9 +19,9 @@ import {
 import { buildHierarchy, flattenBuildHierarchyCompanies } from '@/utils/b3Company';
 import { snackbar } from '@/utils/b3Tip';
 import { deleteCartData } from '@/utils/cartUtils';
-import { CompanyStatusKeyType, isCompanyError } from '@/utils/companyUtils';
+import { CompanyStatusKey, isCompanyError } from '@/utils/companyUtils';
 
-const COMPANY_STATUS_MAPPINGS: Record<CompanyStatusKeyType, string> = {
+const COMPANY_STATUS_MAPPINGS: Record<CompanyStatusKey, string> = {
   pendingApprovalToViewPrices:
     'global.statusNotifications.willGainAccessToBusinessFeatProductsAndPricingAfterApproval',
   pendingApprovalToOrder:
--- a/apps/storefront/src/pages/Login/index.test.tsx
+++ b/apps/storefront/src/pages/Login/index.test.tsx
@@ -23,10 +23,8 @@ vi.mock('@/utils/loginInfo');
 const { server } = startMockServer();
 
 describe('LoginPage', () => {
-  let snackbarErrorSpy: ReturnType<typeof vi.spyOn>;
-
   beforeEach(() => {
-    snackbarErrorSpy = vi.spyOn(snackbar, 'error').mockImplementation(() => {});
+    vi.spyOn(snackbar, 'error');
   });
 
   describe('successful login and redirects', () => {
@@ -146,7 +144,7 @@ describe('LoginPage', () => {
       await renderLoginAndSubmit();
 
       await waitFor(() => {
-        expect(snackbarErrorSpy).toHaveBeenCalledWith(
+        expect(snackbar.error).toHaveBeenCalledWith(
           "Your email address or password is incorrect. Please try again. If you've forgotten your sign in details, just click the 'Forgot your password?' link below.",
         );
       });
@@ -175,13 +173,13 @@ describe('LoginPage', () => {
       await renderLoginAndSubmit();
 
       await waitFor(() => {
-        expect(snackbarErrorSpy).toHaveBeenCalledWith(
+        expect(snackbar.error).toHaveBeenCalledWith(
           'Your business account is pending approval. You will gain access to business account features, products, and pricing after account approval.',
         );
       });
 
       await waitFor(() => {
-        expect(logoutMock).toHaveBeenCalledWith(false);
+        expect(logoutMock).toHaveBeenCalledWith({ showLogoutBanner: false });
       });
     });
 
@@ -208,12 +206,12 @@ describe('LoginPage', () => {
       await renderLoginAndSubmit();
 
       await waitFor(() => {
-        expect(snackbarErrorSpy).toHaveBeenCalledWith(
+        expect(snackbar.error).toHaveBeenCalledWith(
           'Your business account is pending approval. Products, pricing, and ordering will be enabled after account approval.',
         );
       });
       await waitFor(() => {
-        expect(logoutMock).toHaveBeenCalledWith(false);
+        expect(logoutMock).toHaveBeenCalledWith({ showLogoutBanner: false });
       });
     });
 
@@ -240,13 +238,13 @@ describe('LoginPage', () => {
       await renderLoginAndSubmit();
 
       await waitFor(() => {
-        expect(snackbarErrorSpy).toHaveBeenCalledWith(
+        expect(snackbar.error).toHaveBeenCalledWith(
           'Your business account is pending approval. You will gain access to business account features after account approval.',
         );
       });
 
       await waitFor(() => {
-        expect(logoutMock).toHaveBeenCalledWith(false);
+        expect(logoutMock).toHaveBeenCalledWith({ showLogoutBanner: false });
       });
     });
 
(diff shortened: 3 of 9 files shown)


icatalina (commented):
hey @bc-veer , sorry for the delay, have a look at the comments, see if we can reduce the coupling a bit 🙇

apps/storefront/src/pages/CompanyHierarchy/components/HierarchyDialog.tsx
```diff
@@ -65,7 +65,7 @@ function HierarchyDialog({
       if (companyId === Number(currentCompanyId)) {
         await endUserMasqueradingCompany();
       } else if (companyId) {
-        await startUserMasqueradingCompany(Number(companyId));
+        await startUserMasqueradingCompany(Number(companyId), true);
```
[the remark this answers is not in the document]
bc-veer: Sure. Thank you @icatalina

apps/storefront/src/utils/loginInfo.ts
```diff
@@ -196,7 +196,14 @@ const loginWithCurrentCustomerJWT = async () => {
 
   if (!currentCustomerJWT || prevCurrentCustomerJWT === currentCustomerJWT) return undefined;
 
-  const data = await getB2BToken(currentCustomerJWT, channelId);
+  let data;
+  try {
+    data = await getB2BToken(currentCustomerJWT, channelId);
+  } catch (error) {
+    b2bLogger.error('Failed to get B2B token:', error);
```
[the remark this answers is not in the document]
bc-veer: Noted. Thank you

apps/storefront/src/utils/companyUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const getTranslationKeyByMessage = (message: string): string | undefined => {
+  const mapping = COMPANY_STATUS_MAPPINGS.find((item) => item.message === message);
+  return mapping?.key;
+};
+
+export const getTranslationKeyByFlag = (flag: string): string | undefined => {
+  const mapping = COMPANY_STATUS_MAPPINGS.find((item) => item.flag === flag);
+  return mapping?.key;
+};
```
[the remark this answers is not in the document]
icatalina: Yes, definitely, which is why coupling all these sections to the `utils` files seems problematic. Let me know if you need a sounding board to talk about the solution...

apps/storefront/src/pages/CompanyHierarchy/components/HierarchyDialog.tsx
```diff
@@ -93,7 +93,17 @@ function HierarchyDialog({
...
       );
     } catch (error) {
-      b2bLogger.error(error);
+      if (error instanceof Error) {
+        const i18nKey = getTranslationKeyByMessage(error.message);
+        if (i18nKey) {
+          // Delay the snackbar error to avoid race conditions during dialog transition and subsequent navigation to the company hierarchy page.
+          setTimeout(() => {
+            snackbar.error(b3Lang(i18nKey));
+          }, 300);
```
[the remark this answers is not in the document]
bc-veer: **Previous behaviour:**
When a user switched companies from the container header dropdown, the app immediately navigated them to the Company Hierarchy page. During this page transition, the snackbar was removed from the DOM, so the user never got a chance to read the actual error message.

**What’s changed now:**
I updated the logic so that we don’t navigate the user until the masquerading request succeeds.
This also fixes the related issue where the user was navigated even if they cancelled the masquerading confirmation dialog.

This ensures the user remains on the same page long enough to see any snackbar errors, and navigation only happens on a successful company masquerade.

apps/storefront/src/utils/loginInfo.ts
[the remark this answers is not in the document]
bc-veer: Noted. Thank you

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -123,12 +119,21 @@ function Login(props: PageProps) {
...
+          }
+        }
         setLoading(false);
       } finally {
         setLoading(false);
       }
     })();
-  }, [b3Lang, isLoggedIn, logout, searchParams]);
+    // Ignore b3Lang and logout, as they are functions that do not depend on any reactive value
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
icatalina: is this giving you trouble? disabling this eslint rule is problematic, as it can lead to serious bugs.

`b3lang` should be stable, if `useLogout` isn't, try wrapping it in a `useCallback`...

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -123,12 +119,21 @@ function Login(props: PageProps) {
           await logout();
         }
 
+        if (loginFlag && companyStatusMappingKeys.includes(loginFlag)) {
+          await logout(false);
+          const i18nKey = getTranslationKeyByFlag(loginFlag);
+          if (i18nKey) {
+            snackbar.error(b3Lang(i18nKey));
+          }
+        }
```
icatalina: this logic is coupling the concept of a traditional `loginFlag` to `companyStatus` flags. The `loginFlag` parameter now represents both. This is done without modifying the official definition of a `LoginFlagType`. Doing it like so can lead to confusion and maintainability issues in the future.

apps/storefront/src/pages/Login/index.test.tsx
```diff
@@ -1,10 +1,15 @@
...
+  screen,
+  startMockServer,
+  waitFor,
+} from 'tests/test-utils';
 
-import { dispatchEvent } from '@/hooks/useB2BCallback';
-import { b2bLogin, bcLogin, customerLoginAPI } from '@/shared/service/bc';
 import { snackbar } from '@/utils';
-import { getCurrentCustomerInfo } from '@/utils/loginInfo';
+import * as loginInfo from '@/utils/loginInfo';
```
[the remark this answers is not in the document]
bc-veer: Earlier, instead of mocking complete loginInfo, i've modified to mock only particular function `getCurrentCustomerInfo`

```
vi.spyOn(loginInfo, 'getCurrentCustomerInfo').mockResolvedValue({
        userType: 5,
        role: 2,
        companyRoleName: 'Junior Buyer',
      });
```

Now, i've modified the change with loginInfo mock.

apps/storefront/src/pages/Login/index.tsx
bc-veer: Got your point. Updated LoginFlagType with company related login snackbar

apps/storefront/src/App.tsx
```diff
@@ -195,9 +196,16 @@ export default function App() {
       };
 
       if (!customerId) {
-        const info = await getCurrentCustomerInfo();
-        if (info) {
-          userInfo.role = info?.role;
+        try {
```
bc-coolbeans: Right now any non‑`isCompanyError` failure from `getCurrentCustomerInfo` is silently swallowed and the flow just continues. Could we either rethrow / fail the login flow, or at least log/track those errors so we don’t end up with a partially initialised `userInfo` and no visibility into why `getCurrentCustomerInfo` failed?

apps/storefront/src/pages/CompanyHierarchy/components/HierarchyDialog.tsx
```diff
@@ -93,14 +96,43 @@ function HierarchyDialog({
...
+      }
+      isMasquerading.current = false;
     } finally {
       setLoading(false);
-
       handleClose();
     }
   };
 
+  const onExited = () => {
```
bc-coolbeans: This `onExited` handler is doing quite a lot: 
- checking current company, 
- inspecting window.location.hash, masquerade status, permission resolution, and then navigation. 

Could we split this into a couple of smaller helpers (e.g. `redirectIfLeavingOwnCompanyShoppingList`, `redirectIfNoSubsidiaryPermission`) and avoid calling `Number(currentCompanyId)` multiple times so the business rules are easier to read and test?

apps/storefront/src/App.tsx
```diff
@@ -191,7 +192,11 @@ export default function App() {
       };
 
       if (!customerId) {
-        const info = await getCurrentCustomerInfo();
+        const info = await getCurrentCustomerInfo().catch((error) => {
+          if (isCompanyError(error)) {
+            gotoPage(`/login?loginFlag=${error.reason}`);
```
icatalina: I changed this so we can remove the duplication but keep the fixed message, rather than the snackbar, this is more accessible and harder to miss.

apps/storefront/src/shared/service/bc/graphql/login.ts
```diff
@@ -59,15 +60,15 @@ const getB2bLogin = `mutation Login($loginData: UserLoginType!) {
...
+
+export const b2bLogin = (variables: LoginData): Promise<UserLoginResult> =>
+  B3Request.graphqlB2B<UserLoginResult>(
     {
       query: getB2bLogin,
       variables,
     },
-    customMessage,
-  );
+    true,
```
icatalina: no need to have a parameter, nothing uses it.

apps/storefront/src/pages/CompanyHierarchy/components/HierarchyDialog.tsx
bc-coolbeans: we should be trying to address tech debt as we complete tickets rather than doing them separately. It can be easily forgotten

PR #677, opened 2025-12-16
feat(quotes): B2B-4029 quote uuid added to session storage

bc-gauravsobti:
Jira: [B2B-4029](https://bigcommercecloud.atlassian.net/browse/B2B-4029)

## What/Why?
- there was need to pass uuid from url to quotes detail api in checkout repo
- uuid is accessed from url and saved to session storage so that it can be accessed from checkout repo
- `getKey` has been added to B3Storage because storage keys exist only as types and they needed to be accesed in the test

## Testing

With feature flag turned on

https://github.com/user-attachments/assets/09fc51ba-a18a-4acf-85b7-6755ef63de4d

[B2B-4029]: https://bigcommercecloud.atlassian.net/browse/B2B-4029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

With feature flag turned off

https://github.com/user-attachments/assets/60dfc361-d261-4dd4-9ec6-cef71c016130

---

> [!NOTE]
> Adds UUID propagation and storage to support downstream checkout usage.
> 
> - Parse `uuid` from URL via `useSearchParams` in `QuoteDetail`, include in `getQuote` payload and pass to `handleQuoteCheckout`
> - Update `handleQuoteCheckout` to forward `quoteUuid` and call `setQuoteToStorage(quoteId, date, quoteUuid)`
> - Extend `setQuoteToStorage` to persist `quoteCheckoutUuid` alongside existing keys
> - Add tests in `QuoteDetail/index.test.tsx` verifying sessionStorage values when proceeding to checkout for quotes with and without `uuid`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7ea05b6f55981b9e2426afeef0ade8130fe0f964. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/QuoteDetail/index.test.tsx
+++ b/apps/storefront/src/pages/QuoteDetail/index.test.tsx
@@ -1082,4 +1082,150 @@ describe('when the user is a B2B customer', () => {
 
     expect(screen.queryByRole('alert')).not.toBeInTheDocument();
   });
+
+  describe('when "Proceed to checkout" button is clicked', () => {
+    beforeEach(() => {
+      server.use(
+        graphql.query('SearchProducts', () =>
+          HttpResponse.json(
+            buildProductSearchResponseWith({
+              data: {
+                productsSearch: [buildProductSearchWith({ id: 123 })],
+              },
+            }),
+          ),
+        ),
+        graphql.query('getQuoteExtraFields', () =>
+          HttpResponse.json(buildQuoteExtraFieldsWith('WHATEVER_VALUES')),
+        ),
+        graphql.query('ValidateProduct', () =>
+          HttpResponse.json({
+            data: {
+              validateProduct: {
+                responseType: 'SUCCESS',
+                message: 'Product is valid',
+              },
+            },
+          }),
+        ),
+        graphql.query('getStorefrontProductSettings', () =>
+          HttpResponse.json({
+            data: {
+              storefrontProductSettings: {
+                hidePriceFromGuests: false,
+              },
+            },
+          }),
+        ),
+        graphql.mutation('CheckoutQuote', () =>
+          HttpResponse.json({
+            data: {
+              quoteCheckout: {
+                quoteCheckout: {
+                  checkoutUrl:
+                    'https://my-store/cart.php?action=loadInCheckout&id=123&token=1234567889&isFromQuote=Y',
+                  cartId: '123',
+                  cartUrl: 'https://my-store/cart.php?action=load&id=123&token=1234567889',
+                },
+              },
+            },
+          }),
+        ),
+      );
+
+      vitest.mocked(useParams).mockReturnValue({ id: '272989' });
+    });
+    describe('given the quote has uuid', () => {
+      it('sets all the session storage for checkout', async () => {
+        const id = '272989';
+        const uuid = 'lsd-g';
+        const dateString = Date.now().toString();
+        const quote = buildQuoteWith({
+          data: {
+            quote: {
+              id,
+              uuid,
+              quoteNumber: '911911',
+              status: 2,
+              allowCheckout: true,
+              productsList: [buildQuoteProductWith({ productId: '123' })],
+            },
+          },
+        });
+        server.use(graphql.query('GetQuoteInfoB2B', () => HttpResponse.json(quote)));
+        renderWithProviders(<QuoteDetail />, {
+          preloadedState: {
+            ...preloadedState,
+            company: {
+              ...preloadedState.company,
+              permissions: [
+                { code: 'purchase_enable', permissionLevel: 1 },
+                { code: 'checkout_with_quote', permissionLevel: 1 },
+              ],
+            },
+            global: {
+              ...preloadedState.global,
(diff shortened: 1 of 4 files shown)


apps/storefront/src/utils/b3checkout.ts
```diff
@@ -28,8 +30,9 @@ export const attemptCheckoutLoginAndRedirect = async (
   }
 };
 
-export const setQuoteToStorage = (quoteId: string, date: any) => {
+export const setQuoteToStorage = (quoteId: string, date: any, quoteUuid: string | null) => {
   sessionStorage.setItem('isNewStorefront', JSON.stringify(true));
   sessionStorage.setItem('quoteCheckoutId', quoteId);
   sessionStorage.setItem('quoteDate', date?.toString());
+  B3SStorage.set('quoteCheckoutUuid', quoteUuid);
 };
```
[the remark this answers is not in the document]
bc-gauravsobti: My thought process was that if there is a common mechanism (B3SStorage) to store the values in storage, that should be used. I don't know why in some part of the code sessionStorage is being used directly and the other B3SStorage is being used.

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -935,6 +939,7 @@ function QuoteDetail() {
                     role,
                     location,
                     quoteId: quoteDetail.id,
+                    quoteUuid: uuid,
```
icatalina: 🤔 instead of reading uuid from the url, shouldn't it be read from the `quoteDetail` object?

apps/storefront/src/utils/b3checkout.ts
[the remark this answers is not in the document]
icatalina: If checkout is not using `B3SStorage`, it might be worth using the same pattern for all data related to checkout. As Cursor correctly pointed out, there could be integration issues.

As you mentioned, we should use the common mechanism to write/read data from session storage `B3SStorage` when reading/writing _from within buyer portal_. In this case however, this is really an integration point with checkout, and we are writing data for it. It is probably better to make them all consistent and use `sessionStorage` for the new one. We can, in the future, consider moving them all to `B3SSTorage`, or even better, remove our dependency from session storage altogether.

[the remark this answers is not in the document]
bc-gauravsobti: Thank you, yes it is better to be consistent in the checkout regarding storage keys right now. In future, we can make sure that all the keys use B3SStorage. I'll make the change.

apps/storefront/src/pages/QuoteDetail/index.tsx
bc-gauravsobti: Screenshot 2025-12-18 at 12 03 37 pm
Screenshot 2025-12-18 at 12 02 42 pm

- I was not sure why in one place directly id is being used from the url and in another it is being used from quoteDetail. 
- Since this task is towards eventually not passing the id to the quoteDetail api, uuid will be used. So I thought it is better to use uuid from url, which later can be used to get quoteDetail.

apps/storefront/src/pages/QuoteDetail/index.test.tsx
```diff
@@ -957,129 +956,122 @@ describe('when the user is a B2B customer', () => {
...
+            featureFlags: {
+              'B2B-3318.move_stock_and_backorder_validation_to_backend': true,
+            },
           },
-        }),
-      },
+        },
+      });
+      await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
+      const checkoutButton = await screen.findByRole('button', { name: /PROCEED TO CHECKOUT/i });
```
bc-veer: use `getByRole` as button will be available immediately after the loading removed and also avoid regex for name

apps/storefront/src/utils/b3checkout.ts
```diff
@@ -28,8 +28,9 @@ export const attemptCheckoutLoginAndRedirect = async (
   }
 };
 
-export const setQuoteToStorage = (quoteId: string, date: any) => {
+export const setQuoteToStorage = (quoteId: string, date: any, quoteUuid: string | undefined) => {
```
icatalina: 🤔 looking at the `searchParams.get('uuid')`, I see we are defaulting to `null`, but here it is `undefined`...

SunilP-WassDigital:
Hello,

Can you please tell me if i did not upgrade 
https://b2bedition.noticeable.news/publications/action-required-for-custom-b2b-buyer-portal-checkout-or-quote-api-users-effective-march-2-2026 changes, what will happen ? if any thing related to quote does not work.

@bc-gauravsobti

bc-gauravsobti:
@SunilP-WassDigital Can you please test if the new quotes that you create have the uuid? The old quotes created before the upgrade won't have uuid.

PR #692, opened 2025-12-24
fix(checkout): Moved Up Product Validation Api Call with product list…

bc-gauravsobti:
… to not display TBD after loading

## What/Why?

TBD was visible for around a second after loading. 
TBD depends on `validateQuoteProduct` api call which is made after the loading is done. So the `validateQuoteProduct` Api was moved up along with `getQuote` and `products` api calls which are responsible for loading. 

`validateQuoteProduct` depends on `quoteReviewedBySalesRep`(derived state from quote) and `productList`.
Previously both `quoteReviewedBySalesRep` and `productList` were first set in state and then the `useEffect` having the both as deps was used to call the validateQuoteProduct api. 

Now `validateQuoteProduct` is moved along with `getQuote` and `products` api call, we needed `quoteReviewedBySalesRep` and `productList` immediately and not after the state update. So `quoteReviewedBySalesRep` and `productList` are passed as params to `validateQuoteProduct`.

Since `quoteReviewedBySalesRep` is a derived state, it doesn't need to be in the state, hence removed it from the state.

## Testing
In Quote Summary, values should be visible right after loading is done, it shouldn't show TBD.
There were tests to check the value of quote summary, Updated those tests and added delay of 1.2 seconds to the api response to mimick the original api response time. Made sure to synchronously find and expect elements to be in the DOM right after loading, previously the elements were being searched for async and waited to be present in the document.

Before:

https://github.com/user-attachments/assets/ac686431-28cd-42d4-bbd3-ef8205b9624d

After:

https://github.com/user-attachments/assets/2397c0c3-eaec-4637-9a40-dcce0542ab79

---

> [!NOTE]
> Prevents temporary `TBD` in quote summary by validating products before rendering and removing unnecessary state.
> 
> - Move `validateQuoteProducts` into `getQuoteDetail` and pass `productList` and derived `quoteReviewedBySalesRep` directly (no `useEffect` dependency, no stored state)
> - Update backend/frontend validation functions to accept parameters; set `shouldHidePrices`/warnings based on results for immediate price display
> - Derive `quoteReviewedBySalesRep` from `quoteDetail` (`salesRep`/`salesRepEmail`) instead of state
> - Tests: add `msw` `delay`, mock `ValidateProduct` with async delay, and switch to synchronous `getBy*` assertions to verify prices appear right after loading
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9c283d1d7132bc3b6f7957190db571915153d164. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/QuoteDetail/index.test.tsx
+++ b/apps/storefront/src/pages/QuoteDetail/index.test.tsx
@@ -5,6 +5,7 @@ import {
   buildGlobalStateWith,
   buildQuoteWith,
   buildStoreInfoStateWith,
+  delay,
   faker,
   getUnixTime,
   graphql,
@@ -644,16 +645,22 @@ describe('when the user is a B2B customer', () => {
       graphql.query('getQuoteExtraFields', () =>
         HttpResponse.json(buildQuoteExtraFieldsWith('WHATEVER_VALUES')),
       ),
-      graphql.query('ValidateProduct', () =>
-        HttpResponse.json({
+      graphql.query('ValidateProduct', async () => {
+        /* 
+          adding a delay to make sure we are mimicking the scenario where validateProduct api takes time
+          and product error is visible immediately after loading
+        */
+        await delay(200);
+
+        return HttpResponse.json({
           data: {
             validateProduct: {
               responseType: 'ERROR',
               message: 'A product with the id of 123 does not have sufficient stock',
             },
           },
-        }),
-      ),
+        });
+      }),
     );
 
     vitest.mocked(useParams).mockReturnValue({ id: '272989' });
@@ -675,21 +682,17 @@ describe('when the user is a B2B customer', () => {
 
     await waitForElementToBeRemoved(() => screen.queryByText(/loading/i));
     expect(
-      await screen.findByText('A product with the id of 123 does not have sufficient stock'),
+      screen.getByText('A product with the id of 123 does not have sufficient stock'),
     ).toBeInTheDocument();
     const summaryElement = screen.getByTestId('quote-summary');
     const withinSummary = within(summaryElement);
-    expect(await withinSummary.findByRole('row', { name: /Original subtotal/ })).toHaveTextContent(
-      /TBD/,
-    );
-    expect(await withinSummary.findByRole('row', { name: /Quoted subtotal/ })).toHaveTextContent(
-      /TBD/,
-    );
-    expect(await withinSummary.findByRole('row', { name: /Shipping/ })).toHaveTextContent(/TBD/);
-    expect(await withinSummary.findByRole('row', { name: /Grand total/ })).toHaveTextContent(/TBD/);
+    expect(withinSummary.getByRole('row', { name: /Original subtotal/ })).toHaveTextContent(/TBD/);
+    expect(withinSummary.getByRole('row', { name: /Quoted subtotal/ })).toHaveTextContent(/TBD/);
+    expect(withinSummary.getByRole('row', { name: /Shipping/ })).toHaveTextContent(/TBD/);
+    expect(withinSummary.getByRole('row', { name: /Grand total/ })).toHaveTextContent(/TBD/);
   });
 
-  it('renders prices in quote summary if product has no errors', async () => {
+  it('renders prices immediately after loading in quote summary if product has no errors', async () => {
     const quote = buildQuoteWith({
       data: {
         quote: {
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -375,7 +382,10 @@ function QuoteDetail() {
     setQuoteValidationErrors(error);
   };
 
-  const quoteDetailFrontendValidations = () => {
+  const quoteDetailFrontendValidations = (
+    productList: ProductInfoProps[],
```
RobinOng: Same here as well: https://github.com/bigcommerce/b2b-buyer-portal/pull/692/files#r2644789579

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -350,7 +354,10 @@ function QuoteDetail() {
     });
   }, [isB2BUser, quoteDetail, selectCompanyHierarchyId, purchasabilityPermission]);
 
-  const quoteDetailBackendValidations = async () => {
+  const quoteDetailBackendValidations = async (
+    productList: ProductInfoProps[],
```
RobinOng: Same here as well: https://github.com/bigcommerce/b2b-buyer-portal/pull/692/files#r2644789579

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -316,6 +315,11 @@ function QuoteDetail() {
   const isAutoQuotingEnabled =
     quoteConfig.find((item) => item.key === 'quote_auto_quoting')?.value === '1';
 
+  const quoteReviewedBySalesRep =
```
RobinOng: Similar comment here as well where we have `quoteReviewedBySalesRep` here and https://github.com/bigcommerce/b2b-buyer-portal/pull/692/files#diff-fd74bd54643fb77de34e6c6a628aba958ee4737317758aed8d32c5b5fa3bae49R552

bc-gauravsobti: Done, thank you.

apps/storefront/src/pages/QuoteDetail/index.tsx
```diff
@@ -540,7 +544,16 @@ function QuoteDetail() {
...
+
+      const productListResponse = productsWithMoreInfo ?? [];
+      setProductList(productListResponse);
+
+      const { salesRep, salesRepEmail } = quote;
+      const quoteReviewedBySalesRepResponse = !!salesRep || !!salesRepEmail;
+
+      await Promise.resolve(
+        validateQuoteProducts(productListResponse, quoteReviewedBySalesRepResponse),
+      );
```
RobinOng:
```suggestion
      const productListResponse = productsWithMoreInfo ?? [];
      setProductList(productListResponse);
      
      if (productListResponse.length) {
        const { salesRep, salesRepEmail } = quote;
        const quoteReviewedBySalesRepResponse = Boolean(salesRep || salesRepEmail);

        await Promise.resolve(
          validateQuoteProducts(productListResponse, quoteReviewedBySalesRepResponse),
        );
      }
```

apps/storefront/src/pages/QuoteDetail/index.test.tsx
```diff
@@ -644,16 +645,18 @@ describe('when the user is a B2B customer', () => {
       graphql.query('getQuoteExtraFields', () =>
         HttpResponse.json(buildQuoteExtraFieldsWith('WHATEVER_VALUES')),
       ),
-      graphql.query('ValidateProduct', () =>
-        HttpResponse.json({
+      graphql.query('ValidateProduct', async () => {
+        await delay(1200);
```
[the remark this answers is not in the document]
bc-gauravsobti: If no delay is added: 
  - All apis in test immediately give response
  - Even without the code change in this PR, In tests, if we use synchronous search queries (eg. getByRole) instead of async (eg. findByRole) to get the exact values in QuotesSummary instead of TBD, test will pass when they should fail. Hence, they don't mimick our usecase. 
    It can be any delay, I just used 1200 to match closely to what is happening in prod.

RobinOng (approved):

bc-akshatkaushik (approved):
LGTM

PR #841, opened 2026-04-24
B2b 4614 migrate My Orders filter / search / sort state to the SF unified GraphQL 

bc-gauravsobti:
Jira: [B2B-4614](https://bigcommercecloud.atlassian.net/browse/B2B-4614)

## What/Why?

This PR migrates My Orders filter / search / sort state to the SF unified GraphQL `OrdersFiltersInput` + `OrdersSortInput` shape, gated behind `B2B-4613.buyer_portal_unified_sf_gql_orders`. It's the first half of a two-step migration — the filter state shape is unified now; the data source itself swaps to `customer.orders` in ticket 2a.

A temporary `adaptUnifiedToLegacyFilterParams` bridge maps the unified filter shape back to the legacy `FilterSearchProps` / `orderBy` strings the existing `getB2BAllOrders` / `getBCAllOrders` fetcher consumes, so the wire shape is identical with the flag on or off.

### Commit-by-commit

**1. `feat(orders): B2B-4614 add useCustomerOrdersFilterState for the unified customer.orders query`**
Adds the new hook holding search / status / date-range / sort state in the unified `OrdersFiltersInput` + `OrdersSortInput` shape. Includes the customLabel → systemLabel translation so storefront-renamed order statuses still send canonical codes to the API. Pure addition — not wired into the page yet, no user-facing change.

**2. `refactor(orders): B2B-4614 extract legacy filter state into useLegacyOrdersFilterState`**
Pulls the existing legacy filter state out of `Order.tsx` into a sibling hook with the same shape as commit 1, and extracts a pure `translateFilterMoreData` helper into `config.ts`. Also renames `getOrdersCreatedByUser` → `getCreatedByUserForOrders` for clarity. Pure refactor — page behaves identically. Sets up the symmetry the FF gate needs.

**3. `feat(orders): B2B-4614 wire unified filter state into My Orders behind the FF`**
Connects the new filter hook to the page under `B2B-4613.buyer_portal_unified_sf_gql_orders`. The temporary `adaptUnifiedToLegacyFilterParams` adapter maps the unified filter shape to the legacy params, so filter / search / sort changes go through the new state but still hit the legacy data path.

**4. `test(orders): B2B-4614 run My Orders tests under both legacy and unified filter paths`**
Wraps the personal / company / masquerade `/orders` describe blocks in `describe.each` so every scenario runs twice — once with the FF off, once with it on. No new test cases, just a new axis. Locks in the adapter's "wire-shape unchanged" guarantee.

## Rollout/Rollback

**Rollout:** Behind feature flag `B2B-4613.buyer_portal_unified_sf_gql_orders` (registered in PR #838). Flag defaults to off.

**Success signal:** GraphQL request shape on `/orders` is identical with the flag off vs on (validated by the parameterized tests in commit 4). If a store sees no behavior change after we flip the flag, the unified state shape is working end-to-end.

**Rollback:** Flip the FF off. No code revert, no migration. Because the adapter produces wire-identical GraphQL, the off-state matches `main` exactly.

## Testing

**Automated:**
- `MyOrders/index.test.tsx` — every existing scenario (personal customer, company customer, masquerading super admin) now runs under both `ffOn: false` (legacy filter path) and `ffOn: true` (unified filter path). Same GraphQL assertions hold both ways.
- `adaptUnifiedToLegacyParams.test.ts` — unit tests covering all field mappings, sort key/direction translation, B2B `isShowMy` flag, default values for missing fields.

[B2B-4614]: https://bigcommercecloud.atlassian.net/browse/B2B-4614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: refactors `Order.tsx` filter/sort/search state and query-key logic and introduces a new unified code path (behind `B2B-4613...`) that could change request params or caching behavior when enabled.
> 
> **Overview**
> Migrates **My Orders** filter/search/sort state to the unified SF GraphQL shapes (`OrdersFiltersInput` + `OrdersSortInput`) behind `B2B-4613.buyer_portal_unified_sf_gql_orders`, while keeping the legacy fetch path intact via a temporary `adaptUnifiedToLegacyFilterParams` bridge.
> 
> Refactors `Order.tsx` to use new `useCustomerOrdersFilterState`/`useLegacyOrdersFilterState` hooks, extracts translation of “more filters” metadata into `translateFilterMoreData`, and renames `getOrdersCreatedByUser` to `getCreatedByUserForOrders`.
> 
> Expands test coverage to assert unified-path filter behavior (search, status/date ranges, customLabel→systemLabel mapping) and adds unit tests validating the adapter’s field/sort conversions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df6f8bc2fae307c26d2088e009560da35a5fd961. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/MyOrders/index.test.tsx
+++ b/apps/storefront/src/pages/MyOrders/index.test.tsx
@@ -3,7 +3,6 @@ import {
   buildB2BFeaturesStateWith,
   buildCompanyStateWith,
   builder,
-  buildGlobalStateWith,
   buildStoreInfoStateWith,
   bulk,
   faker,
@@ -83,17 +82,7 @@ afterEach(() => {
   vi.useRealTimers();
 });
 
-// Run each /orders scenario under both FF states. Bridge produces wire-identical
-// GraphQL, so all assertions hold either way; drift on any role cut fails one test.
-const FF_KEY = 'B2B-4613.buyer_portal_unified_sf_gql_orders';
-const ordersFilterFFStates = [
-  { ffOn: false, label: 'legacy b2b api path' },
-  { ffOn: true, label: 'unified storefront api path' },
-];
-const buildFFGlobalState = (ffOn: boolean) =>
-  buildGlobalStateWith({ featureFlags: { [FF_KEY]: ffOn } });
-
-describe.each(ordersFilterFFStates)('when a personal customer ($label)', ({ ffOn }) => {
+describe('when a personal customer', () => {
   const preloadedState = {
     company: buildCompanyStateWith({
       customer: {
@@ -101,7 +90,6 @@ describe.each(ordersFilterFFStates)('when a personal customer ($label)', ({ ffOn
       },
     }),
     storeInfo: buildStoreInfoStateWith({ timeFormat: { display: 'j F Y' } }),
-    global: buildFFGlobalState(ffOn),
   };
 
   beforeEach(() => {
@@ -829,7 +817,7 @@ const buildCompanyOrderStatusesWith = builder<CompanyOrderStatuses>(() => ({
   },
 }));
 
-describe.each(ordersFilterFFStates)('when a company customer ($label)', ({ ffOn }) => {
+describe('when a company customer', () => {
   const preloadedState = {
     company: buildCompanyStateWith({
       customer: {
@@ -841,7 +829,6 @@ describe.each(ordersFilterFFStates)('when a company customer ($label)', ({ ffOn 
       },
     }),
     storeInfo: buildStoreInfoStateWith({ timeFormat: { display: 'j F Y' } }),
-    global: buildFFGlobalState(ffOn),
   };
 
   beforeEach(() => {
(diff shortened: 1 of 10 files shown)


apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
```diff
@@ -0,0 +1,121 @@
+import { useEffect, useMemo, useState } from 'react';
+
+import { OrdersFiltersInput, OrdersSortInput } from '@/shared/service/bc/graphql/orders';
+
+import { getCustomerOrdersInitFilter, packDateRange } from './unifiedApiFilters';
+
+export type SortableColumnKey = 'orderId' | 'poNumber' | 'totalIncTax' | 'status' | 'createdAt';
```
bc-gauravsobti: Why does this not have placedBy sortable key?

apps/storefront/src/pages/order/adaptUnifiedToLegacyFilterParams.ts
```diff
@@ -0,0 +1,41 @@
...
+  filters,
+  activeSort,
+  isB2BUser,
+}: AdaptUnifiedToLegacyParamsArgs): {
+  filterData: Partial<FilterSearchProps>;
+  orderBy: string;
+} => ({
+  filterData: {
+    q: filters.search ?? '',
+    statusCode: filters.status,
```
bc-gauravsobti: Can statusCode be undefined because status can be undefined here.

apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
bc-gauravsobti: So placedBy is for CompanyOrders, make sure to mention that in PR, that this scope will be covered in 3a.

bc-gauravsobti:
bugbot run

apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
```diff
@@ -0,0 +1,129 @@
...
+    }));
+  };
+
+  const handleCompanyIdsChange = (companyIds: number[]) => {
+    const isAll = companyIds.length === 0 || companyIds.includes(-1);
+    setFilters((prev) => ({
+      ...prev,
+      companyIds: isAll ? undefined : companyIds.map(String),
+    }));
+  };
```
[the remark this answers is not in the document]
bc-gauravsobti: Fixed

apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
```diff
@@ -0,0 +1,129 @@
...
+  );
+
+  const handleSearchChange = (key: string, value: string) => {
+    if (key !== 'search') return;
+    setFilters((prev) => ({ ...prev, search: value || undefined }));
+  };
+
+  const handleFilterChange = (value: AppliedFilters) => {
+    const raw = normalizeString(value.orderStatus);
+    const status = raw
```
bc-jeffteng: [Edge cases]
When `orderStatuses` is still empty (the window between mount and `initFilter()` resolving), `raw` is the UI dropdown label (e.g. "Awaiting Fulfillment") rather than the system code (`AWAITING_FULFILLMENT`). The `??` raw fallback passes the display label directly to the API — silently. The API will return zero results or mismatched results with no error surface to the user.

apps/storefront/src/pages/MyOrders/index.test.tsx
```diff
@@ -81,14 +83,25 @@ afterEach(() => {
...
+// GraphQL, so all assertions hold either way; drift on any role cut fails one test.
+const FF_KEY = 'B2B-4613.buyer_portal_unified_sf_gql_orders';
+const ordersFilterFFStates = [
+  { ffOn: false, label: 'legacy b2b api path' },
+  { ffOn: true, label: 'unified storefront api path' },
+];
+const buildFFGlobalState = (ffOn: boolean) =>
+  buildGlobalStateWith({ featureFlags: { [FF_KEY]: ffOn } });
+
+describe.each(ordersFilterFFStates)('when a personal customer ($label)', ({ ffOn }) => {
```
bc-jeffteng: The `describe.each` parameterisation is a good structural choice for catching regressions, but the action tests (sort changes, filter changes) use `mockReturnValue` without argument matching — so under `ffOn: true`, these pass even if `useCustomerOrdersFilterState`'s handlers are broken. Worth adding at least one `when(...).calledWith(stringContainingAll(...))` assertion in a sort or filter test so the `ffOn: true` run actually validates the unified handler produces the right query string, not just that the component doesn't crash. 
For example, the "sorts orders by different columns" test could assert the `orderBy` string changes under both flag states, similar to how the masquerade test already uses `stringContainingAll('orderBy: "-bcOrderId"')`.

apps/storefront/src/pages/order/adaptUnifiedToLegacyFilterParams.ts
```diff
@@ -0,0 +1,41 @@
...
+export const adaptUnifiedToLegacyFilterParams = ({
+  filters,
+  activeSort,
+  isB2BUser,
+}: AdaptUnifiedToLegacyFilterParamsArgs): {
+  filterData: Partial<FilterSearchProps>;
+  orderBy: string;
+} => ({
+  filterData: {
+    q: filters.search ?? '',
```
bc-gary, line 31: **[Suggestion]** — `statusCode` inconsistency: `undefined` vs `''`

`filters.status` can be `undefined`, but the legacy path always provides a string (`''` when empty). While `orders.ts:57` handles this via `"${data?.statusCode || ''}"`, matching the legacy behavior explicitly is safer and addresses the earlier comment about this:

```ts
statusCode: filters.status ?? '',
```

apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
bc-gary: **[Suggestion]** — Guard against empty `orderStatuses` during status translation

When `orderStatuses` is still empty (between mount and `initFilter()` resolving), the `?? raw` fallback sends the display label (e.g. "Awaiting Fulfillment") directly as the API status code. Practically unlikely since the dropdown has no options until statuses load, but the new code could improve on the legacy pattern:

```ts
const status = raw
  ? (orderStatuses.find((s) => s.customLabel === raw || s.systemLabel === raw)?.systemLabel ??
    (orderStatuses.length > 0 ? raw : undefined))
  : undefined;
```

apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
```diff
@@ -0,0 +1,129 @@
...
+  const reset = () => {
+    setFilters(getCustomerOrdersInitFilter(companyId));
+  };
+
+  return {
+    filters,
+    sortBy,
+    activeSort,
+    handleSearchChange,
+    handleFilterChange,
```
bc-gary, line 126: **[Suggestion]** — `reset()` doesn't reset `activeSort`

`reset` clears filters but leaves `activeSort` at whatever state it was in. Not consumed in this PR, but when wired up later this will be a subtle bug — filters reset but sort direction persists.

```ts
const reset = () => {
  setFilters(getCustomerOrdersInitFilter(companyId));
  setActiveSort(DEFAULT_SORT);
};
```

bc-gauravsobti: Screenshot 2026-04-28 at 1 35 19 pm
reset comes from this option, it doesn't have any relation with sorting as per view.

bc-gauravsobti: I dug deeper in this, we actually don't need reset filter because the filters state gets changed only on clicking apply button which is already handled via `handleFilterChange`
So, I will go ahead and remove reset function as well because it is dead code.

apps/storefront/src/pages/order/useCustomerOrdersFilterState.ts
bc-gauravsobti: Took a different direction — returning `undefined` whenever the lookup misses, whether the list is empty or just doesn't contain the `filter`. Sending a display label as the API status code is wrong either way, so I'd rather drop the filter than send something the backend will reject. This diverges from the legacy hook's `|| currentStatus` fallback, but that path only triggers on a stale value the backend wouldn't accept anyway. Happy to switch to your exact suggestion if you'd prefer parity with legacy under the FF.

bc-gauravsobti: @bc-jeffteng Please check this [comment](https://github.com/bigcommerce/b2b-buyer-portal/pull/841#discussion_r3151446675).

apps/storefront/src/pages/MyOrders/index.test.tsx
bc-gauravsobti: @bc-jeffteng sort related tests are the only culprit here, they will be handled in the sort PR 2c.

bc-gauravsobti:
Squashed the commits and rebased with main.

bc-gary (approved):
Thanks for the work @bc-gauravsobti 👏

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -113,93 +95,76 @@ function Order({ isCompanyOrder = false }: OrderProps) {
...
+  const legacyFilterState = useLegacyOrdersFilterState({
+    isB2BUser,
+    isCompanyOrder,
+    selectedCompanyId,
+    orderStatuses: getOrderStatuses,
+  });
+  const customerFilterState = useCustomerOrdersFilterState({
+    companyId: selectedCompanyId,
+    orderStatuses: getOrderStatuses,
+  });
```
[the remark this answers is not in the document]
bc-gauravsobti: Looking into this.

bc-jeffteng (approved):
LGTM, great work.

apps/storefront/src/pages/order/Order.tsx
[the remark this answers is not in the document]
bc-gauravsobti: Resolved, using featureflag inside the hooks and only letting useEffect run when it is on the right path.

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -113,95 +95,80 @@ function Order({ isCompanyOrder = false }: OrderProps) {
...
-        isAgenting,
-        createdByUsers,
-        orderStatuses,
-      );
-
       setOrderStatuses(orderStatuses);
 
-      const filterInfoWithTranslatedLabel = filterInfo.map((element) => {
-        const translatedElement = element;
-        translatedElement.label = b3Lang(element.idLang);
```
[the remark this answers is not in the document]
bc-gauravsobti: Fixed, divided fetchList into legacy and unified orders. Making sure right params are being passed, also making sure right `queryKey` is used by `useQuery`.

apps/storefront/src/pages/MyOrders/index.test.tsx
```diff
@@ -1578,19 +1579,19 @@ describe('when a company customer', () => {
...
+    },
+  }),
+  storeInfo: buildStoreInfoStateWith({ timeFormat: { display: 'j F Y' } }),
+  b2bFeatures: buildB2BFeaturesStateWith({
+    masqueradeCompany: {
+      id: 433,
+      isAgenting: true,
+    },
+  }),
+};
```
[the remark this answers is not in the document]
bc-gauravsobti: Fixed

bc-gauravsobti:
## Review feedback follow-ups

- Gate each filter hook's `useEffect` with the same FF check the caller uses, so the dormant path doesn't fire `setFilterData` / `setFilters` and trigger spurious re-renders (Cursor bot — low).
- Split `fetchList` into `fetchUnifiedOrders` and `fetchLegacyOrders` so each path's `queryKey` matches what its fetcher actually sends — no more cache-vs-fetch drift (Cursor bot — medium).
- Add filter behavior tests to `unified-orders.test.tsx` (search, status + date, date-only, custom-status → systemLabel resolution) for both B2C and B2B. All 81 MyOrders tests pass as well.

Last 2 commits will need an another look. Thank you guys. 

cc: @bc-gary @bc-jeffteng

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -113,93 +95,76 @@ function Order({ isCompanyOrder = false }: OrderProps) {
...
+    handleSetOrderBy,
+  } = isUnifiedOrdersNonCompanyOrderPath ? customerFilterState : legacyFilterState;
+
+  const { filterData, orderBy } = isUnifiedOrdersNonCompanyOrderPath
+    ? adaptUnifiedToLegacyFilterParams({
+        filters: customerFilterState.filters,
+        activeSort: customerFilterState.activeSort,
+        isB2BUser,
+      })
+    : { filterData: legacyFilterState.filterData, orderBy: legacyFilterState.orderBy };
```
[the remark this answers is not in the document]
bc-gauravsobti: No need

bc-gary:
Looked through the last two commits — all three follow-ups are addressed:

- **Hook gating** — both hooks now check the FF internally and skip their effect on the dormant path. Clean.
- **fetchList split** — `fetchUnifiedOrders` / `fetchLegacyOrders` with separate query keys (`orderList:unified` vs `orderList:legacy`) eliminates the cache-vs-fetch drift. Filters and sort are now passed explicitly as params instead of read from the closure.
- **Filter tests** — solid coverage across B2C and B2B (search, status+date, date-only, custom label→systemLabel). The `calledWith(expect.objectContaining(...))` assertions verify the actual GraphQL variables, which is what we needed.

LGTM — good to merge.

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -113,103 +99,92 @@ function Order({ isCompanyOrder = false }: OrderProps) {
...
+  const fetchUnifiedOrders = async ({
+    first,
+    filters,
+    sortBy,
+  }: {
+    first: number;
+    filters: OrdersFiltersInput;
+    sortBy: OrdersSortInput;
+  }): Promise<{ edges: ListItem[]; totalCount: number }> => {
+    const result = await getCustomerOrders({ first, filters, sortBy });
```
bc-jeffteng, line 177: I think there is still one pagination gap on the unified path. The `B3Table` updates its `offset` for page changes, but the `getCustomerOrders` only receives the `first`, `filter`, and `sort`. Since the SF query use `after/before` cursor rather than offset-based, page 2 can refetch the same first page unless we track the cursor or change the pagination contract for this path.

PR #843, opened 2026-04-28
[B2B-4615] Implement cursor-based pagination for My Orders

bc-gary:
Jira: [B2B-4615](https://bigcommercecloud.atlassian.net/browse/B2B-4615)

## What/Why?

Replaces offset-based pagination (`first`/`offset`) with Relay-style cursor pagination (`first`/`after`, `last`/`before`) on the My Orders unified SF GQL path. All changes are gated behind the existing `B2B-4613.buyer_portal_unified_sf_gql_orders` feature flag — the legacy path is completely unchanged.

### Why cursor pagination?

The SF GQL `customer.orders` query uses a Relay connection model (`OrdersConnection`) with `pageInfo` and cursors. The legacy offset-based approach doesn't forward cursors, so page 2+ always returned the same first page of results. This PR fixes that by tracking cursors and passing `after`/`before` on each page navigation.

### What changed

**`useCustomerOrdersFilterState.ts`** — Added cursor pagination state (`after`/`before` cursors, `pageSize`, `pageInfo`, `currentPage`). Exposes `paginationVariables` (computed `first/after` or `last/before`), `handlePageChange`, `handlePageSizeChange`, `updatePageInfo`. All filter/sort/search handlers reset cursors to page 1 via `resetPagination()`. `handlePageChange` includes a dedup guard — if the cursor being set matches what's already in state (fetch in-flight), the call is skipped to prevent `currentPage` desync from rapid clicks.

**`Order.tsx`** — `fetchUnifiedOrders` reads `paginationVariables` from the hook and passes cursors to `getCustomerOrders()`. Returns `pageInfo` from the response, which is synced into the hook via a `useEffect` calling `updatePageInfo`. `handleUnifiedPaginationChange` bridges B3Table's offset-based page events to cursor operations. `cursorPageInfo` is passed to B3Table to control next/prev button disabled state. `goToDetail` passes `totalCount: -1` so DetailPagination knows it's the unified path.

**`B3Table.tsx`** — `labelDisplayedRows` shows `{from}–{to}` (range only) when `count === -1`, instead of displaying "-1" as the total. New `cursorPageInfo` prop uses `slotProps` on `TablePagination` to properly disable next/prev buttons based on `hasNextPage`/`hasPreviousPage` — without this, MUI's unknown-count mode keeps "next" always enabled.

[...]

PR #846, opened 2026-04-30
feat(orders): B2B-4824 Migrate order detail header with unified SF GQL

bc-lulu-yang:
Jira: [B2B-4824](https://bigcommercecloud.atlassian.net/browse/B2B-4824)

## What/Why?
Replace the dual B2B/B2C order detail fetch with a single unified SF GQL call (`getOrderDetail`), and update `DetailPagination` component to use the unified SF GQL orders query (`getCustomerOrders`) for the My Orders navigation. Both changes are gated behind the existing B2B-4613 feature flag, and the legacy fetch paths are fully preserved when the flag is off.

A new `convertOrderDetail` adapter maps the SF GQL Order type onto the `OrderDetailsState` fields required for the order detail header (order ID, status label, PO number). Unit tests cover both the unified and legacy paths, including boundary cases for pagination prev/next behaviour.

## Rollout/Rollback
Feature flag added to cover the changes. And revert this PR if needed.

## Testing
Order header in legacy works as expected (order ID, status label)
Screenshot 2026-04-30 at 3 57 03 pm

[B2B-4824]: https://bigcommercecloud.atlassian.net/browse/B2B-4824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the order-detail data fetching path (behind a feature flag) and alters detail-page navigation state, which could cause missing header fields or broken back/next behavior if the new GraphQL response shape differs.
> 
> **Overview**
> Adds a feature-flagged *unified* order-detail fetch: when `B2B-4613.buyer_portal_unified_sf_gql_orders` is on, `OrderDetail` now calls `getOrderDetail` (SF GraphQL) and maps the result via a new `convertOrderDetail` adapter for the header fields (order id, status label, PO number); legacy B2B/B2C REST fetching remains when the flag is off.
> 
> Updates order-list navigation state so the unified orders listing routes to order detail with `totalCount: -1` and unified filter/sort state, and updates `DetailPagination` to treat `totalCount === -1` as the unified path and hide prev/next navigation (and adjust `aria-labelledby`) until cursor-based detail navigation is implemented.
> 
> Adds a new unified-path test (`index.unified.test.tsx`) that mocks the `GetOrderDetail` GraphQL query and verifies header rendering and back navigation under the flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4705154a1d0449792da7e47cd64217d440ff611. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
```diff
@@ -73,56 +89,76 @@ export function DetailPagination({ onChange, color }: DetailPageProps) {
...
-    if (listIndex === totalCount - 1) {
-      flag = 'toRight';
-      leftId = list[list.length - 2]?.node.orderId || 0;
-    } else if (listIndex === 0) {
-      flag = 'toLeft';
-      rightId = list[1]?.node.orderId || 0;
+    let rightId: number | string = '';
+    let leftId: number | string = '';
+
+    if (isUnifiedOrdersNonCompanyOrderPath) {
```
bc-gauravsobti: Better to keep legacy and current logic in separate functions. fetchList can contain unified orders logic and the old one can be names fetchLegacyList

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -207,14 +207,18 @@ function Order({ isCompanyOrder = false }: OrderProps) {
...
         currentIndex: index,
-        searchParams: {
-          ...filterData,
-          orderBy,
-        },
         totalCount: allTotal,
         isCompanyOrder,
-        beginDateAt: filterData?.beginDateAt,
-        endDateAt: filterData?.endDateAt,
+        ...(isUnifiedOrdersNonCompanyOrderPath
```
bc-gauravsobti: Let's keep old and new logic in separate functions. `legacyGoToDetail`, `goToDetail`

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
```diff
@@ -0,0 +1,285 @@
...
+const unifiedLocationState = {
+  isCompanyOrder: false,
+  totalCount: 3,
+  unifiedCustomerFilters: { status: 'Pending' },
+  unifiedCustomerSortBy: OrdersSortInput.CREATED_AT_NEWEST,
+  searchParams: { orderBy: '-createdAt', offset: 0, first: 10 },
+};
+
+describe('DetailPagination', () => {
+  describe('with B2B-4613 flag ON, non-company order — unified SF GQL path', () => {
```
bc-gauravsobti: we can't reference ticket name in test name, I think having it like this `with unified orders flag ON` would give more context

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
```diff
@@ -0,0 +1,285 @@
...
+          searchParams: { orderBy: '-createdAt', offset: 0, first: 3 },
+        },
+        b2bUserFlagOnState,
+      );
+
+      await waitFor(() => expect(getAllOrders).toHaveBeenCalled());
+    });
+  });
+
+  describe('with B2B-4613 flag OFF — legacy path', () => {
```
bc-gauravsobti: Same here: flag name change

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
```diff
@@ -0,0 +1,285 @@
...
+    it('calls GetAllOrders (B2B legacy) when the unified flag is disabled', async () => {
```
bc-gauravsobti: We already have in describe that the flag is off, we can remove it from here

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
```diff
@@ -0,0 +1,285 @@
...
+          totalCount: 3,
+          searchParams: { orderBy: '-createdAt', offset: 0, first: 3 },
+        },
+        b2bUserFlagOffState,
+      );
+
+      await waitFor(() => expect(getAllOrders).toHaveBeenCalled());
+    });
+
+    it('calls GetCustomerOrders (B2C legacy) when the unified flag is disabled', async () => {
```
bc-gauravsobti: [Question]
Why is this test helpful?
I thought only having tests which check whether next and previous are working would help knowing that the behaviour is working properly.

apps/storefront/src/pages/OrderDetail/index.tsx
```diff
@@ -111,29 +116,44 @@ function OrderDetail() {
...
-
-            setIsCurrentCompany(Number(companyInfo.companyId) === Number(currentCompanyId));
-
-            const data = convertB2BOrderDetails(newOrder, b3Lang);
-            dispatch({
-              type: 'all',
-              payload: data,
-            });
-            setPreOrderId(orderId);
+          if (isUnifiedOrders) {
```
bc-gauravsobti: Let's keep different useEffects: one for legacy, one for new unified orders
Pros of this approach:
- less dependency in one useEffect
- No if else logic
- Easy to delete the useEffect later on for the legacy way to fetch orders

bc-gauravsobti (changes requested):

apps/storefront/src/pages/order/Order.tsx
bc-jeffteng, line 1: I think the unified path breaks the route-state handoff into `DetailPagination`.

In `Order.tsx`, `goToDetail()` still passes `totalCount: allTotal`, but `allTotal` is only updated by the legacy fetch path via `setAllTotal(totalCount)`. The unified fetch returns `totalCount: -1` and never updates `allTotal`, so when the FF is on we navigate to detail with `totalCount: 0`.

`DetailPagination` then reads that as `state.totalCount || 0` and disables both arrows because both buttons use `totalCount <= 1`. That means the direct component tests pass because they inject `totalCount: 3`, but the real `Order.tsx -> DetailPagination` flow can render “Order N of 0” with prev/next disabled.

There’s a related route-state issue too: unified navigation omits `searchParams`, so `DetailPagination` falls back to a fresh `{ offset: 0 }` object and uses it as an effect dependency. Once pagination is enabled, that can reset `listIndex` back to `currentIndex + 0` instead of preserving the absolute list offset.

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
bc-gary: **[Blocker]** `totalCount: -1` breaks both the display and button disabled state on the unified path.

On `main`, B2B-4613 sets `allTotal = -1`. This gets passed as `totalCount` here. Two issues:

1. **Display** (line 209): renders "Order 2 of -1"
2. **Buttons** (lines 216/224): `totalCount <= 1` is always `true` for `-1`, so both prev/next are permanently disabled even though the unified fetch correctly populates `rightLeftSide`.

**Recommended fix:** On the unified path, drive disabled state from `arrived`/`loading` only (not `totalCount`), and hide "of Y" in the counter display (pending product confirmation on whether to show position-only or hide entirely).

```tsx
const isPrevDisabled = isUnifiedOrdersNonCompanyOrderPath
  ? arrived === 'toLeft' || loading
  : totalCount <= 1 || arrived === 'toLeft' || loading;

const isNextDisabled = isUnifiedOrdersNonCompanyOrderPath
  ? arrived === 'toRight' || loading
  : totalCount <= 1 || arrived === 'toRight' || loading;
```

For the display, hide "of Y" when on the unified path — consistent with B3Table's range-only label pattern when count is unknown (from B2B-4615). **Note: hiding "of Y" needs product confirmation** — check with product whether showing "Order 2" without total is acceptable, or if a different treatment is preferred.

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
```diff
@@ -73,56 +89,76 @@ export function DetailPagination({ onChange, color }: DetailPageProps) {
...
+      // Cursor-based: fetch listIndex + 2 items from the start so that
+      // edges[listIndex - 1] (prev) and edges[listIndex + 1] (next) are in the result.
+      const result = await getCustomerOrders({
+        first: listIndex + 2,
+        filters: unifiedCustomerFilters,
+        sortBy: unifiedCustomerSortBy ?? OrdersSortInput.CREATED_AT_NEWEST,
```
bc-gary: **[Question]** Should this be capped to prevent fetching hundreds of orders?

`first: listIndex + 2` means if the user is viewing order #200, this fetches 202 orders over GraphQL. Consider adding a ceiling — e.g. `first: Math.min(listIndex + 2, 50)` — and hiding prev/next when beyond that cap.

This would prevent unbounded fetches for users deep in their order list while B2B-4629 (4f) implements the proper cursor-based neighbor lookup.

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
```diff
@@ -73,56 +89,76 @@ export function DetailPagination({ onChange, color }: DetailPageProps) {
...
     beginDateAt = state?.beginDateAt || null;
     endDateAt = state?.endDateAt || null;
-    searchParams = state?.searchParams || {
-      offset: 0,
-    };
+    isCompanyOrder = state?.isCompanyOrder ?? false;
+    searchParams = state?.searchParams || { offset: 0 };
+    unifiedCustomerFilters = state?.unifiedCustomerFilters;
+    unifiedCustomerSortBy = state?.unifiedCustomerSortBy;
   }
```
bc-gary, line 79: **[Required]** This PR will need to rebase on PR #843 (B2B-4615) after it merges.

#843 modifies both `DetailPagination.tsx` and `Order.tsx` — it adds `isUnifiedPath` guard, `return null` for the unified path, renames `pagination` to `legacyPagination`, and adds `cursorPageInfo` to B3Table.

After rebase:
- Replace the `return null` with this unified fetch implementation
- Use the `isUnifiedOrdersNonCompanyOrderPath` pattern already established by #843
- Adopt #843's `Order.tsx` as-is and add `unifiedCustomerFilters`/`unifiedCustomerSortBy` to `goToDetail`

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
```diff
@@ -73,56 +89,76 @@ export function DetailPagination({ onChange, color }: DetailPageProps) {
...
-    setRightLeftSide({
-      leftId,
-      rightId,
-    });
-
+    setRightLeftSide({ leftId, rightId });
     setArrived(flag);
-
     setLoading(false);
   };
```
bc-gary, line 131: **[Required]** +1 to Gaurav's feedback — split `fetchList` into `fetchUnifiedList` / `fetchLegacyList`.

The 60-line `if/else` block mixes two completely different fetch strategies (cursor-based SF GQL vs offset-based legacy). Separate functions make each path independently readable and make legacy removal in B2B-4630 a clean delete.

```tsx
const fetchUnifiedList = async () => { /* getCustomerOrders path */ };
const fetchLegacyList = async () => { /* getB2BAllOrders / getBCAllOrders path */ };

const fetchList = async () => {
  setLoading(true);
  const result = isUnifiedOrdersNonCompanyOrderPath
    ? await fetchUnifiedList()
    : await fetchLegacyList();
  setRightLeftSide(result);
  setArrived(result.flag);
  setLoading(false);
};
```

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-gary: **[Required]** +1 to Gaurav — drop ticket references from test names. Use `'with unified orders flag ON'` / `'with unified orders flag OFF'` instead of ticket numbers. Ticket numbers go stale; the flag name gives context.

apps/storefront/src/pages/OrderDetail/index.tsx
```diff
@@ -111,29 +116,44 @@ function OrderDetail() {
...
+              setPreOrderId(orderId);
+            }
+          } else {
+            const order = isB2BUser ? await getB2BOrderDetails(id) : await getBCOrderDetails(id);
+
+            if (order) {
+              const { products, companyInfo } = order;
+
+              const newOrder = {
+                ...order,
```
bc-gary: **[Suggestion]** The TODO references B2B-4824 but this *is* B2B-4824. Update to reference B2B-4825 (Task 2 — cross-company banner) so it's tracked against the right ticket.

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+      const { onChange } = renderDetailPagination(
+        { ...unifiedLocationState, currentIndex: 1 },
+        flagOnState,
+      );
+
+      const nav = await screen.findByRole('navigation', { name: 'Order 2 of 3' });
+      const [prev] = within(nav).getAllByRole('button');
+      await waitFor(() => expect(prev).toBeEnabled());
+
```
bc-gary: **[Suggestion]** B2B-4824 AC says tests should include "honours the orderBy parameter". No current test asserts that `getCustomerOrders` receives the `unifiedCustomerSortBy` value from location state. Consider adding one:

```tsx
it('forwards the sort order to getCustomerOrders', async () => {
  const getOrders = vi.fn().mockReturnValue(buildCustomerOrdersResponse([101, 102, 103]));
  server.use(graphql.query('GetCustomerOrders', ({ variables }) =>
    HttpResponse.json(getOrders(variables))
  ));

  renderDetailPagination(
    { ...unifiedLocationState, unifiedCustomerSortBy: OrdersSortInput.ID_A_TO_Z },
    flagOnState,
  );

  await waitFor(() => expect(getOrders).toHaveBeenCalledWith(
    expect.objectContaining({ sortBy: 'ID_A_TO_Z' })
  ));
});
```

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
bc-gary: @bc-sylvialin Can you comment on this. This is one of the limitation of cursor pagination similar to the one on Orders list page.

bc-lulu-yang: Hi @bc-sylvialin , just added a screenshot here. Below is the legacy order detail.  `Order 1 of 2` top right of the page is what are thinking to hide for unified orders, since we can't get the `totalCount` due to the limitation of cursor pagination

Screenshot 2026-05-01 at 11 14 39 am

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-lulu-yang: Updated

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-lulu-yang: updated

apps/storefront/src/pages/order/Order.tsx
bc-lulu-yang: updated

apps/storefront/src/pages/OrderDetail/index.tsx
bc-lulu-yang: updated

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-lulu-yang: updated

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
bc-lulu-yang: confirmed with product. Now hide the `Order X of Y` , updated

bc-lulu-yang: Thanks @bc-gauravsobti, I've reverted the changes related to the `DetailPagination` component. I had a quick catch up with @bc-gary,  we have a separated card to handle the logic of `prev/next` - `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse).  So weather we need to fetch or not for the unified order path, will be decided in that ticket.

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-lulu-yang: I've deleted this test file, since the logic of `prev/next` button will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse).

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-lulu-yang: Same as above.  I've deleted this test file, since the logic of `prev/next` button will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse)

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
bc-lulu-yang: I've deleted this test file, since the logic of `prev/next` button will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse)

apps/storefront/src/pages/order/Order.tsx
bc-lulu-yang: Same as above.  Had a discussion with Gary, all the logic of `prev/next` button in  `DetailPagination` will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse)

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
bc-lulu-yang: As we discussed, all the logic of `prev/next` button in  `DetailPagination` will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse)

apps/storefront/src/pages/OrderDetail/components/DetailPagination.test.tsx
bc-lulu-yang: As we discussed, all the logic of `prev/next` button in  `DetailPagination` will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse)

apps/storefront/src/pages/OrderDetail/components/DetailPagination.tsx
bc-lulu-yang: As we discussed, all the logic of `prev/next` button in  `DetailPagination` will be handled in `4f` [B2B-4629](https://bigcommercecloud.atlassian.net/browse)

apps/storefront/src/pages/OrderDetail/index.tsx
bc-lulu-yang: I've updated the logic about `setIsCurrentCompany` with a comment reference B2B-4825 to double confirm. I will add this in the ticket of B2B-4825

bc-gary (approved):
LGTM. Nice work @bc-lulu-yang 👏

apps/storefront/src/pages/OrderDetail/index.unified.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+  }),
+  global: buildGlobalStateWith({
+    featureFlags: {
+      'B2B-4613.buyer_portal_unified_sf_gql_orders': true,
+    },
+  }),
+  storeInfo: buildStoreInfoStateWith({ timeFormat: { display: 'j F Y' } }),
+};
+
+describe('Order detail path with B2B-4613 unified SF GQL flag ON', () => {
```
bc-gauravsobti: No ticket reference in test please.

bc-gauravsobti (changes requested):
Just one small change. Everything else looks good.

bc-gauravsobti (approved):
LGTM.

PR #853, opened 2026-05-06
fix(register): B2B-4861 Buyer Portal: Add null check for stateRequired flag and remove FF

bc-vasu:
Jira: [B2B-4861](https://bigcommercecloud.atlassian.net/browse/B2B-4861)

## What/Why?

Removed rolled-out feature flag B2B-4481.use_grpc_geo_for_state_required_flag and all its call sites.

Added null check for stateRequired in getIsStateRequired — falls back to stateList.length > 0 only when the flag is absent, correctly honoring false.

stateRequired is now always requested in the Countries GraphQL query.

Applied getIsStateRequired to the Address List form (previously bypassed the flag).

## Rollout/Rollback

Rollback or Revert

## Testing

https://github.com/user-attachments/assets/b1644bbf-3b31-49e0-9c1e-9108d78bd5d2

https://github.com/user-attachments/assets/943905d8-0896-4fa8-a96b-1bf9ff6c5c1f

https://github.com/user-attachments/assets/448e113d-91b0-4580-856c-3bbbee5f05d9

After(Addresses)
Before (Addresses)
registeredbctob2b

[B2B-4861]: https://bigcommercecloud.atlassian.net/browse/B2B-4861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes how `state` field required-ness is computed across registration and address forms and updates the countries GraphQL query shape, which can affect form validation/UX if backend data differs or is missing.
> 
> **Overview**
> Removes the rolled-out `B2B-4481.use_grpc_geo_for_state_required_flag` and updates all call sites to fetch countries without flag-gated behavior.
> 
> Introduces a shared `getIsStateRequired` helper that treats `country.stateRequired` as authoritative when present (including `false`), falling back to `states.length > 0` when absent, and applies this consistently to registration and address book forms so `state` required-ness matches backend data.
> 
> Updates the countries GraphQL query/response typing to always request `stateRequired` and removes the old `getIsStateRequired` implementation from `Registered/config.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e89e61b6d20c2bf79e62757b119952307705479f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/hooks/useGetCountry.ts
+++ b/apps/storefront/src/hooks/useGetCountry.ts
@@ -1,7 +1,6 @@
 import { useContext, useEffect } from 'react';
 import { Control, FieldValues, UseFormGetValues, UseFormSetValue, useWatch } from 'react-hook-form';
 
-import { useFeatureFlag } from '@/hooks/useFeatureFlag';
 import { GlobalContext } from '@/shared/global';
 import { Country, State } from '@/shared/global/context/config';
 import { getB2BCountries } from '@/shared/service/b2b';
@@ -12,14 +11,10 @@ const useSetCountry = () => {
     dispatch,
   } = useContext(GlobalContext);
 
-  const grpcGeoForStateRequiredFlag = useFeatureFlag(
-    'B2B-4481.use_grpc_geo_for_state_required_flag',
-  );
-
   useEffect(() => {
     const init = async () => {
       if (countriesList && !countriesList.length) {
-        const { countries } = await getB2BCountries(grpcGeoForStateRequiredFlag);
+        const { countries } = await getB2BCountries();
 
         dispatch({
           type: 'common',
--- a/apps/storefront/src/hooks/useGetCountry.ts
+++ b/apps/storefront/src/hooks/useGetCountry.ts
@@ -1,6 +1,7 @@
 import { useContext, useEffect } from 'react';
 import { Control, FieldValues, UseFormGetValues, UseFormSetValue, useWatch } from 'react-hook-form';
 
+import { getIsStateRequired } from '@/pages/Registered/config';
 import { GlobalContext } from '@/shared/global';
 import { Country, State } from '@/shared/global/context/config';
 import { getB2BCountries } from '@/shared/service/b2b';
@@ -89,8 +90,11 @@ const useGetCountry = ({
   // Populate state array when the user change selected country
   useEffect(() => {
     if (!countryCode || !countriesList?.length) return;
-    const stateList =
-      countriesList.find((country: Country) => country.countryCode === countryCode)?.states || [];
+    const selectedCountry = countriesList.find(
+      (country: Country) => country.countryCode === countryCode,
+    );
+    const stateList = selectedCountry?.states || [];
+    const isStateRequired = getIsStateRequired(selectedCountry, stateList);
     const stateFieldIndex = addresses.findIndex(
       (formFields: FormFieldsProps) => formFields.name === 'state',
     );
@@ -103,10 +107,10 @@ const useGetCountry = ({
                 ...addressField,
                 fieldType: 'dropdown',
                 options: stateList,
-                required: true,
+                required: isStateRequired,
               };
             }
-            return { ...addressField, fieldType: 'text', options: [], required: false };
+            return { ...addressField, fieldType: 'text', options: [], required: isStateRequired };
           }
           return addressField;
         }),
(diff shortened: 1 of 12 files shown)


apps/storefront/src/pages/AddressList/components/AddressForm.tsx
```diff
@@ -487,11 +488,11 @@ function AddressForm(
         if (stateList.length > 0) {
           stateFields.fieldType = 'dropdown';
           stateFields.options = stateList;
-          stateFields.required = true;
+          stateFields.required = isStateRequired;
```
bc-gary, line 491: [Required] `handleBackFillData` (lines 450-458 in this file) bypasses the new `getIsStateRequired` logic.

The `handleCountryChange` effect now correctly uses `getIsStateRequired`, but `handleBackFillData` — called when opening the edit address form — still hardcodes `field.required = true` when states exist and `field.required = false` when they don't.

This means for a country where `stateRequired` is explicitly `false` but states exist (e.g. optional provinces), editing an existing address will incorrectly mark state as required on initial form load, creating an inconsistency with the country-change handler.

Suggested: apply `getIsStateRequired` in `handleBackFillData` the same way it's applied here.

apps/storefront/src/hooks/useGetCountry.ts
```diff
@@ -12,14 +11,10 @@ const useSetCountry = () => {
     dispatch,
   } = useContext(GlobalContext);
 
-  const grpcGeoForStateRequiredFlag = useFeatureFlag(
-    'B2B-4481.use_grpc_geo_for_state_required_flag',
-  );
-
   useEffect(() => {
     const init = async () => {
```
bc-gary, line 16: [Required] The `useGetCountry` hook (lines 106-114 in this file) also hardcodes `required: true` when states exist and `required: false` when they don't, bypassing the new `getIsStateRequired` utility.

This hook is used by `QuoteAddress.tsx`, so the Quote flow will still have incorrect state-required behavior for countries where `stateRequired === false` but states exist.

Suggested: import and apply `getIsStateRequired` inside `useGetCountry` as well.

apps/storefront/src/pages/AddressList/components/AddressForm.tsx
```diff
@@ -476,9 +477,9 @@ function AddressForm(
 
   useEffect(() => {
     const handleCountryChange = (countryCode: string) => {
-      const stateList =
-        countries.find((country: CountryProps) => country.countryCode === countryCode)?.states ||
-        [];
+      const selectedCountry = countries.find((country) => country.countryCode === countryCode);
+      const stateList = selectedCountry?.states || [];
```
bc-gary, line 481: [Suggestion] The `as Country` cast here works because `getIsStateRequired` only accesses `country?.stateRequired`, but `CountryProps` and `Country` have incompatible `id` and `states` types that this cast hides.

Consider either widening `getIsStateRequired`'s parameter to accept `{ stateRequired?: boolean } | undefined`, or using the `CountryProps` type directly (which now includes `stateRequired`). This would eliminate the need for the cast.

bc-gary (approved):

apps/storefront/src/hooks/useGetCountry.ts
```diff
@@ -1,7 +1,7 @@
 import { useContext, useEffect } from 'react';
 import { Control, FieldValues, UseFormGetValues, UseFormSetValue, useWatch } from 'react-hook-form';
 
-import { useFeatureFlag } from '@/hooks/useFeatureFlag';
+import { getIsStateRequired } from '@/pages/Registered/config';
```
bc-veer: [Required]
Avoid using Pages inside Hooks. Usually, the dependency direction should be the other way around (Pages using Hooks).

apps/storefront/src/pages/Registered/config.ts
```diff
@@ -48,11 +48,12 @@ export interface State {
 }
 
 export function getIsStateRequired(
```
bc-veer: [Required]
Since this logic is shared across the Buyer Portal, can we move it to utils?

apps/storefront/src/pages/Registered/config.ts
bc-veer, line 41: Since the GraphQL query now always requests `stateRequired`, the field could be typed as boolean
  (non-optional) in Country.

bc-veer (approved):

PR #858, opened 2026-05-08
feat(orders): static local order data

bc-jeffteng:
Jira: [B2B-4859](https://bigcommercecloud.atlassian.net/browse/B2B-4859)

## What/Why?

Adds an opt-in, dev-only static order data path for unified order services. When `DEV` and `VITE_USE_STATIC_ORDER_DATA=true`, order services load ignored JSON from `apps/storefront/public/mock-data/orders.json`; otherwise they call Storefront GraphQL as before.

## Rollout/Rollback

Run `yarn generate:static-orders` from `apps/storefront/` to refresh the local JSON. The generated file is ignored and can be deleted anytime. Turn the flag off or remove the file to fall back to GraphQL.

## Testing

Run from `apps/storefront/`:

```bash
yarn generate:static-orders
yarn test src/shared/service/bc/graphql/orders.test.ts --run
yarn test src/pages/MyOrders/unified-orders.test.tsx src/pages/CompanyOrderList/unified-company-orders.test.tsx src/pages/OrderDetail/index.unified.test.tsx --run
yarn tsc --noEmit
yarn lint:eslint
yarn lint:dependencies
```

[B2B-4859]: https://bigcommercecloud.atlassian.net/browse/B2B-4859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/storefront/src/mocks/README.md
```diff
@@ -0,0 +1,223 @@
+# B3Request GraphQL Mock Router
```
bc-akshatkaushik, line 1: [required]

We generally don't add READMEs to specific parts of an application, docs should live in the docs folder or root README

Secondly I am not sure we should document this in the public repo as it will be accessible to everyone, again leaking information of our plans to add new GQL APIs which are only relevant within the org not outside it

bc-akshatkaushik:
[question]

What is the benefit of using this over static data from an assumed API response? [KISS]

i.e.

`var x = fetchOrders().then()` can be mocked as a API call to return data as per the expected schema as per this PR (if I understand it correctly)

VS

`var x = {mockResponse: ...}`

yarn.lock
```diff
@@ -2323,6 +2323,24 @@
...
+
+"@graphql-tools/mock@^9.1.7":
+  version "9.1.7"
+  resolved "https://registry.yarnpkg.com/@graphql-tools/mock/-/mock-9.1.7.tgz#fcf5bff779b4a6cb58d1c5413327fe8bd6039830"
+  integrity sha512-JpsmamDwd6PO1Oj437T+pdM5tpV41XrUD3GJsdrAVEtDRljdQvbxVxmt6tXruqMHj6XAOPoS1Ny1e/G3HA7sgg==
+  dependencies:
+    "@graphql-tools/schema" "^10.0.33"
+    "@graphql-tools/utils" "^11.1.0"
+    fast-json-stable-stringify "^2.1.0"
+    tslib "^2.4.0"
```
bc-gary, line 2342: [Required] **Stale `@graphql-tools/mock` and transitive dependencies**

`@graphql-tools/mock@^9.1.7` (and its transitive deps `@graphql-tools/schema@^10.0.33`, `@graphql-tools/merge@^9.1.9`, `@graphql-tools/utils@^11.1.0`) appear in the lock file but no `package.json` references `@graphql-tools/mock` as a direct dependency. These were likely installed during earlier development iterations — the "trim mock branch scope" commit removed the code that used them, but the lock file wasn't regenerated.

Run `yarn install` and commit the cleaned lock file to remove the phantom entries.

apps/storefront/package.json
```diff
@@ -6,6 +6,7 @@
   "prettier": "@bigcommerce/eslint-config/prettier",
   "scripts": {
     "dev": "vite",
+    "dev:mock": "VITE_USE_MOCK_API=true vite",
```
bc-gary, line 9: [Required] **PR description references `yarn build:assert-no-mock-bundle` but the script doesn't exist**

The PR body claims "The mock layer is additionally verified by `yarn build:assert-no-mock-bundle`" and lists it as a testing step, but neither this diff nor the existing `package.json` on `main` defines this script. If this guard is part of the production-safety story, it should be included in this PR. If it was intentionally deferred, the PR description should be updated to avoid confusion.

apps/storefront/src/mocks/resolvers/orders.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  return sorted.sort((left, right) => right.orderedAt.utc.localeCompare(left.orderedAt.utc));
+}
+
+export async function executeGetCustomerOrders({
+  variables,
+}: OperationExecutionContext = {}): Promise<SuccessfulGetCustomerOrdersResponse> {
+  const normalizedVariables = normalizeVariables(variables);
+  const filteredOrders = orderStore
```
bc-gary, line 97: [Suggestion] **`sortOrders` silently falls back to newest-first for 8 of 10 sort enum values**

Only `CREATED_AT_OLDEST` is explicitly handled — everything else (`ID_A_TO_Z`, `STATUS_A_TO_Z`, `REFERENCE_Z_TO_A`, etc.) silently produces `CREATED_AT_NEWEST` ordering. A developer testing sort behavior with one of those values would get unexpected results with no indication why.

Consider adding a `logMockDecision` call for unimplemented sort values so it's visible in the dev console, e.g.:
```ts
default:
  logMockDecision(`sortOrders: ${sortBy} not implemented, falling back to CREATED_AT_NEWEST`);
```

apps/storefront/src/mocks/requestRouter.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    operationName: 'GetCustomerOrders',
+    transport: 'sf-proxy',
+  });
+  const originalExecute = customerOrdersOperation?.execute;
+
+  afterEach(() => {
+    if (customerOrdersOperation && originalExecute) {
+      customerOrdersOperation.execute = originalExecute;
+    }
+  });
```
bc-gary, line 30: [Suggestion] **Test mutates shared module-level registry state**

Directly reassigning `customerOrdersOperation.execute` on a shared module-level registry object is fragile — if a test fails before `afterEach` runs or test isolation changes, it could leak into other tests. Consider using `vi.spyOn` on the registry's `getOwnedOperation` to return a test-local operation object, keeping the real registry untouched.

apps/storefront/src/hooks/useFeatureFlag.test.ts
```diff
@@ -50,3 +54,65 @@ it('returns false when the flag is explicitly false', () => {
   });
   expect(result.result.current).toBe(false);
 });
+
+it('uses a local dev override when the requested flag is set to true', () => {
+  vi.stubEnv('DEV', true);
+  vi.stubEnv('MODE', 'development');
```
bc-lulu-yang, line 60: [Nit]

In a real Vite dev session, import.meta.env.MODE is usually `development`, but seems our `useFeatureFlag.ts` don't need it mocked.  probably we can remove it

PR #862, opened 2026-05-10
feat(orders): B2B-4825 Populate order summary, history, comments and cross-company banner in unified SF GQL adapter

bc-jeffteng:
Jira: [B2B-4825](https://bigcommercecloud.atlassian.net/browse/B2B-4825)

## What/Why?

Extends the unified SF GQL order detail adapter (`convertOrderDetail`) to populate the second tier of fields left blank after Task 1 (B2B-4824):

- **Order summary** (`orderSummary`, `money`, `currencyCode`) — subtotal, shipping, handling fee, discounts, tax, grand total; handling fee row omitted when zero; decimal places derived from the order's own currency
- **Order history** (`history`) — maps SF GQL `OrderHistoryEvent` to the legacy `OrderHistoryItem` shape including Unix timestamp conversion
- **Customer comments** (`orderComments`)
- **PO number** (`poNumber`) — already mapped in Task 1, end-to-end test added here
- **Cross-company banner** — fixed guard from `products?.length` to `orderId`; `isCurrentCompany` initialised to `true` to prevent banner flash before fetch resolves; uses `order.company.entityId` (SF GQL) instead of legacy `companyInfo.companyId`

All changes are gated behind the `B2B-4613.buyer_portal_unified_sf_gql_orders` feature flag. Legacy path is unaffected.

## Rollout/Rollback

Feature flagged via `B2B-4613.buyer_portal_unified_sf_gql_orders`. Rollback by disabling the flag.

## Testing
I use random mocking data to show what has been done and is available under the detail page
Screenshot 2026-05-14 at 1 36 37 pm

SF GQL backend is not yet available in dev, so testing is via the unified test file:

- `src/pages/OrderDetail/index.unified.test.tsx` — 8 tests covering:
  - Order summary renders with correct price rows
  - Handling fee row omitted when cost is zero
  - Order history section hidden when empty
  - Order history renders date, status rows when events present
  - PO number rendered in header
  - Cross-company banner shown when order belongs to a different company
- `src/pages/OrderDetail/index.test.tsx` — 35 existing legacy tests, all passing

All 43 tests pass.

[B2B-4825]: https://bigcommercecloud.atlassian.net/browse/B2B-4825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes unified order-detail fetching and state mapping (currency formatting, history timestamps, cross-company detection), which can affect what buyers see on order detail pages, though it is gated behind a feature flag.
> 
> **Overview**
> When `B2B-4613.buyer_portal_unified_sf_gql_orders` is enabled, the unified order-detail adapter now populates previously-empty fields: **currency-aware `orderSummary`** (including optional handling fee/coupon rows), `money`/`currencyCode`, `payment.dateCreateAt`, `orderComments`, and **mapped `history` events** (with event type normalization and Unix timestamps).
> 
> `OrderDetail` refactors the unified fetch to store the raw `order` separately and then dispatch converted state when either the order or currencies change, preventing stale updates and enabling currency formatting updates without a refetch; it also initializes `isCurrentCompany` to `true` and fixes the cross-company banner guard to use `orderId` (reducing banner flash / incorrect gating). Comprehensive unified-path tests were added to cover summary formatting, handling-fee omission, history visibility/rows, PO number display, cross-company banner, comments rendering, and currency reformatting without reloading.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b71247293206717968a180e40b5c176b4b61d028. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/OrderDetail/index.tsx
+++ b/apps/storefront/src/pages/OrderDetail/index.tsx
@@ -359,7 +359,7 @@ function OrderDetail() {
             )}
           </Grid>
         </Grid>
-        {products?.length && !isCurrentCompany ? (
+        {orderId && !isCurrentCompany ? (
           <Box
             sx={{
               marginTop: '24px',
--- a/apps/storefront/src/pages/OrderDetail/index.tsx
+++ b/apps/storefront/src/pages/OrderDetail/index.tsx
@@ -44,6 +44,7 @@ function OrderDetail() {
   const isAgenting = useAppSelector(({ b2bFeatures }) => b2bFeatures.masqueradeCompany.isAgenting);
 
   const companyInfoId = useAppSelector(({ company }) => company.companyInfo.id);
+  const currencies = useAppSelector(({ storeConfigs }) => storeConfigs.currencies);
   const { selectCompanyHierarchyId } = useAppSelector(
     ({ company }) => company.companyHierarchyInfo,
   );
@@ -68,14 +69,7 @@ function OrderDetail() {
   } = useContext(GlobalContext);
 
   const {
-    state: {
-      poNumber,
-      status = '',
-      customStatus,
-      orderSummary,
-      orderStatus = [],
-      digitalProducts,
-    },
+    state: { poNumber, status = '', customStatus, orderSummary, orderStatus = [], digitalProducts },
     state: detailsData,
     dispatch,
   } = useContext(OrderDetailsContext);
@@ -178,7 +172,7 @@ function OrderDetail() {
         if (order) {
           dispatch({
             type: 'all',
-            payload: convertOrderDetail(order, b3Lang),
+            payload: convertOrderDetail(order, b3Lang, currencies),
           });
 
           // BC omits `company` on Order — no cross-company check. B2B: compare order company to viewer context.
@@ -200,7 +194,7 @@ function OrderDetail() {
 
     fetchUnifiedOrderDetails();
     // eslint-disable-next-line react-hooks/exhaustive-deps -- dispatch stable
-  }, [isUnifiedOrders, orderId, preOrderId, currentCompanyId]);
+  }, [isUnifiedOrders, orderId, preOrderId, currentCompanyId, currencies]);
 
   useEffect(() => {
     if (!orderId) {
--- a/apps/storefront/src/pages/OrderDetail/index.tsx
+++ b/apps/storefront/src/pages/OrderDetail/index.tsx
@@ -17,6 +17,7 @@ import {
   getBcOrderStatusType,
   getOrderStatusType,
 } from '@/shared/service/b2b';
+import type { Order as UnifiedOrder } from '@/shared/service/bc/graphql/orders';
 import { getOrderDetail } from '@/shared/service/bc/graphql/orders';
 import { isB2BUserSelector, useAppSelector } from '@/store';
 import { AddressConfigItem, CustomerRole, OrderProductItem, OrderStatusItem } from '@/types';
@@ -89,6 +90,7 @@ function OrderDetail() {
   const [orderId, setOrderId] = useState('');
   const [isRequestLoading, setIsRequestLoading] = useState(false);
   const [isCurrentCompany, setIsCurrentCompany] = useState(true);
+  const [unifiedOrder, setUnifiedOrder] = useState<UnifiedOrder | null>(null);
 
   useEffect(() => {
     setOrderId(params.id || '');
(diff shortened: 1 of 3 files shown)


apps/storefront/src/pages/OrderDetail/index.tsx
```diff
@@ -202,7 +194,7 @@ function OrderDetail() {
 
     fetchUnifiedOrderDetails();
     // eslint-disable-next-line react-hooks/exhaustive-deps -- dispatch stable
-  }, [isUnifiedOrders, orderId, preOrderId, currentCompanyId]);
+  }, [isUnifiedOrders, orderId, preOrderId, currentCompanyId, currencies]);
```
bc-gary: [Suggestion] **`currencies` in useEffect dependency triggers full re-fetch on unrelated slice changes**

+1 to the Bugbot observation. The `currencies` selector grabs the entire `storeConfigs.currencies` slice (including `enteredInclusiveTax`). Any mutation to that slice — even one unrelated to formatting — creates a new reference, re-runs the effect, and re-fetches the order from the API.

`currencies` is only needed for the `convertOrderDetail` conversion, not the network call. Consider separating fetch from conversion so that currency changes re-format locally without a redundant API round-trip — e.g. store the raw `Order` in a ref and derive the converted state via `useMemo` keyed on `currencies`.

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
```diff
@@ -1,10 +1,149 @@
...
+      [labels.grandTotal]: formatPrice(order.totalIncTax.value, decimalPlaces),
+    },
+    priceSymbol: {
+      [labels.subTotal]: 'subTotal',
+      [labels.shipping]: 'shipping',
+      ...(hasHandlingFee && { [labels.handlingFee]: 'handlingFee' }),
+      [labels.discountAmount]: 'discountAmount',
+      ...couponSymbols,
+      [labels.tax]: 'tax',
+      [labels.grandTotal]: 'grandTotal',
```
bc-gary, line 89: [Question] **`nonCouponDiscountTotal` row shown even when value is zero**

The handling fee row is conditionally omitted when zero (`hasHandlingFee`), but the "Discount amount" row always renders — even when `nonCouponDiscountTotal.value === 0`. Is this intentional parity with the legacy path, or should it also be conditionally hidden to avoid a `$0.00` discount line?

apps/storefront/src/pages/OrderDetail/index.tsx
bc-jeffteng: Fixed. Narrowed the selector from `storeConfigs.currencies` (full `Currencies` slice) to `storeConfigs.currencies.currencies` (`Currency[]`) and updated `convertOrderDetail`/`buildMoneyFormat` to accept `Currency[]` directly. `enteredInclusiveTax` is not used in the conversion at all, so the broader selector was unnecessary — now currency slice mutations (e.g. `setEnteredInclusiveTax`) no longer create a new reference for this selector, preventing spurious re-fetches.

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
bc-jeffteng: Intentional parity with the legacy path. `B2BOrderData.ts:144` always renders `[labels.discountAmount]: formatPrice(discountAmount || '')` regardless of whether the value is zero. Matching that behaviour here avoids any visual inconsistency between the two paths during the migration window.

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
```diff
@@ -1,10 +1,149 @@
...
+      return 2;
+    default:
+      return 2;
+  }
+}
+
+function mapHistoryEvents(history: Order['history']): OrderHistoryItem[] {
+  return history.map((e) => ({
+    id: Number(e.id),
+    eventType: toNumericEventType(e.eventType),
```
bc-lulu-yang, line 108: [suggestion]

Seems this `toNumericEventType` is only introduced to map to legacy type `OrderHistoryItem`, but no need in UI. 

Probably we can add a TODO comment here or in the clean up jira ticket to remove this no-need mapping.

apps/storefront/src/pages/OrderDetail/index.unified.test.tsx
```diff
@@ -186,4 +207,351 @@ describe('Order detail path with unified SF GQL flag ON', () => {
...
+    });
+
+    server.use(
+      graphql.query('GetOrderDetail', () =>
+        HttpResponse.json(
+          buildOrderDetailResponseWith({
+            data: {
+              site: {
+                order: buildUnifiedOrderWith({
+                  entityId: 6696,
```
bc-lulu-yang: [suggestion]
1. For lines 229–249: we probably can extract the `buildUnifiedOrderWith` arguments/object into a named `orderDetails` const so the expect assertions at the bottom can reference these values, `order.subTotal.value` etc. instead of re-hardcoding `13.5`. One value change = one-line edit, and the test becomes self-documenting.

2. Seems totalIncTax: 100 doesn't add up. The correct sum of the fixture's own line items is `102 + 332 + 22.2 − 37.93 + 13.5 = 431.77` (correct me if I am wrong here). I know we pass totalIncTax straight through from the API without computing it. But feel like still worth to make it correct.

apps/storefront/src/pages/OrderDetail/index.tsx
bc-lulu-yang: Agreed with @bc-gary. Probably we can do something like below (haven't tested, just an idea)

1. Stabilize deps and make the `useEffect` depends on `storeCurrenciesKey` instead of `currencies`
```
// this is the line you updated already
const currencies = useAppSelector(({ storeConfigs }) => storeConfigs.currencies.currencies);
  const storeCurrenciesKey = useMemo(
    () => currencies.map((c) => c.currency_code).join(','),
    [currencies],
  );
```

2. Wait for currencies before fetch

Before `fetchUnifiedOrderDetails`, add below
```
    if (!currencies.length) {
      return;
    }
```

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
bc-jeffteng: Yeah, I just want the code to look cleaner. If you have any concerns, we can remove it

apps/storefront/src/pages/OrderDetail/index.unified.test.tsx
bc-jeffteng: fixed

apps/storefront/src/pages/OrderDetail/index.tsx
bc-jeffteng: Following Lulu's suggestion, I reviewed the code and found that the currencies field is not empty by default; it defaults to the US format. Therefore, using the currency's length is not the correct solution for this.

The core problem is that we want to avoid double fetching, meaning the fetch should only run once per load. My proposed solution is to separate the flow. The first part of the `useEffect` will fetch the order. Afterward, once the order is fetched, the second `useEffect` can dispatch the order detail to Redux with the correct currencies. This approach will prevent the double fetching

bc-gary (approved):
Nice work @bc-jeffteng 👏

bc-lulu-yang (approved):

bc-akshatkaushik:
[question]

Unit Tests run on UI
Can we add UI manual test video to PR description with mock data?

apps/storefront/src/pages/OrderDetail/index.tsx
```diff
@@ -161,48 +156,69 @@ function OrderDetail() {
...
+    let isCurrentRequest = true;
+
     const fetchUnifiedOrderDetails = async () => {
       const id = parseInt(orderId, 10);
       if (!id) {
+        setUnifiedOrder(null);
         return;
       }
 
+      setUnifiedOrder(null);
```
bc-veer:
```suggestion
setUnifiedOrder(null);
if (!id) {
     return;
}
```

bc-jeffteng: fixed

bc-launchbay:
success [20260514034608](https://launchbay.bigcommerce.net/projects/515/releases/284805) by Launchbay

bc-launchbay:
success [20260514034608](https://launchbay.bigcommerce.net/projects/515/releases/284805) by Akshat Kaushik

PR #864, opened 2026-05-12
fix(common): B2B-4848 replace anchor with button for Business Account Application nav

bc-lulu-yang:
Jira: [B2B-4848](https://bigcommercecloud.atlassian.net/browse/B2B-4848)

### Problem

The "Business Account Application" link injected into the storefront navigation had two accessibility violations:

- The element was an `<a>` tag with `href="javascript:;"` — a placeholder that gives no meaningful destination. Screen readers and keyboard users expect an anchor to navigate somewhere; using it purely as a click trigger is misleading and does not meet WCAG 2.1.1 (Keyboard) or WCAG 4.1.2 (Name, Role, Value).
- The `aria-label` was hardcoded as `"Gift Certificates"` introducing misleading.

### Fix

Replaced `href="javascript:;"` with `href="/registeredbctob2b"` and removed the incorrect `aria-label` in `useRegisteredbctob2b.ts`.

### Why `<a>` is kept instead of changing to `<button>`

The element is kept as an `<a>` tag because it represents navigation to a real destination, here `/registeredbctob2b` is a registered route in the application. In a nav bar where all sibling items are links, using `<a>` is semantically correct and consistent with user expectations.

The click handler calls `e.preventDefault()` so the SPA can handle routing via `setOpenPage`, opening the registration form as an overlay. And the `href` provides a meaningful, accessible destination while JavaScript handles the in-app experience. Without JavaScript, the `href` still points somewhere real rather than nowhere.

`<button>` would be appropriate if this element had no URL destination, for example, opening a modal with no corresponding route. Since `/registeredbctob2b` is a real route, `<a>` is the correct element.

### Why `aria-label` was removed

Per the [MDN ARIA spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label), `aria-label` is intended for elements that have no visible text label, for example an icon-only link or button. A `<a>`'s accessible name is derived automatically from its text content. Since the link already renders the visible text "Business Account Application" (from `global.registerB2B.linkText`), adding `aria-label` would override that visible text, hiding it from assistive technologies and risking a WCAG 2.5.3 (Label in Name) mismatch. Removing it means the accessible name and the visible label are identical, which is the most correct and robust state.

## Rollout/Rollback
Revert this PR if needed

## Testing

**After**

https://github.com/user-attachments/assets/a0edba4b-5f56-431c-9270-69eb5552f9d9

[B2B-4848]: https://bigcommercecloud.atlassian.net/browse/B2B-4848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storefront/src/hooks/dom/useRegisteredbctob2b.ts
+++ b/apps/storefront/src/hooks/dom/useRegisteredbctob2b.ts
@@ -53,12 +53,9 @@ const useRegisteredbctob2b = (setOpenPage: Dispatch<SetStateAction<OpenPageState
       const convertB2BNavNode = document.createElement('li');
       convertB2BNavNode.className = 'navUser-item navUser-convert-b2b';
       convertB2BNavNode.innerHTML = `
-        <button
-          class="navUser-action"
-          type="button"
-        >
+        <a class="navUser-action" href="/registeredbctob2b">
           ${b3Lang('global.registerB2B.linkText')}
-        </button>
+        </a>
       `;
       return convertB2BNavNode;
     };
@@ -86,9 +83,10 @@ const useRegisteredbctob2b = (setOpenPage: Dispatch<SetStateAction<OpenPageState
 
       accountNode?.parentNode?.insertBefore(convertB2BNavNode, accountNode);
 
-      const linkNode = convertB2BNavNode.querySelector('button');
+      const linkNode = convertB2BNavNode.querySelector('a');
       if (linkNode) {
-        linkNode.onclick = () => {
+        linkNode.onclick = (e) => {
+          e.preventDefault();
           setOpenPage({
             isOpen: true,
             openUrl: '/registeredbctob2b',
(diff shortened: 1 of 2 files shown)


apps/storefront/src/hooks/dom/useRegisteredbctob2b.ts
```diff
@@ -53,9 +53,12 @@ const useRegisteredbctob2b = (setOpenPage: Dispatch<SetStateAction<OpenPageState
       const convertB2BNavNode = document.createElement('li');
       convertB2BNavNode.className = 'navUser-item navUser-convert-b2b';
       convertB2BNavNode.innerHTML = `
-        <a class="navUser-action" href="javascript:;" aria-label="Gift Certificates">
+        <button
```
bc-akshatkaushik: [question]

Is it possible to add a test case for this?
i.e. render page and observe button with text to be in the document

bc-lulu-yang: I've added some unit tests for this button

bc-veer (approved):

bc-gary (approved):

bc-launchbay:
success [20260513231602](https://launchbay.bigcommerce.net/projects/515/releases/284777) by Launchbay

bc-launchbay:
success [20260513231602](https://launchbay.bigcommerce.net/projects/515/releases/284777) by Launchbay

PR #876, opened 2026-05-14
[B2B-3766] Fix shared component translations not loading on quote detail

bc-gary:
Jira: [B2B-3766](https://bigcommercecloud.atlassian.net/browse/B2B-3766)

## What/Why?

Merchant custom translations for shared quote components were not applied on the Quote Detail page.

**Root cause:** The buyer portal fetches translations per page based on the URL path segment.
When visiting `/quoteDetail/:id`, only `quoteDetail.*` keys are fetched from the translation API.
However, shared components rendered on that page (QuoteInfo, ContactInfo, QuoteAddress,
QuoteNote, QuoteAttachment, Message) use `quoteDraft.*` translation keys. Since `quoteDraft`
translations were never fetched on the quote detail page, merchant custom translations for
those keys were silently ignored — the UI fell back to default English values from `en.json`.

**Fix:** Add a `PAGE_DEPENDENCIES` map (alongside the existing `REPEATED_PAGES` map) in
`appAsyncThunks.ts`. When `getPageTranslations` is dispatched for a page that has dependencies,
it fetches the page and all its dependent pages in parallel using `Promise.all`, then merges
all translations into a single response. The existing `condition` guard on the thunk prevents
re-fetching pages that are already loaded.

Currently maps: `quoteDetail → ['quoteDraft']`. Extensible for future cross-page dependencies.

This is a single-file change — no modifications to the router or Redux slice were needed.

## Rollout/Rollback

- **Rollout:** Standard merge to main. No feature flags — the change is additive
  (fetches additional translations that were previously missing). Only affects stores
  with custom translations configured in the B2B Dashboard.
- **Rollback:** Revert the commit. No data migrations or side effects.

## Testing

- TypeScript compiles cleanly (`tsc --noEmit`)
- No new lint errors

Screenshot 2026-05-14 at 5 37 47 pm
Screenshot 2026-05-14 at 5 40 43 pm
Screenshot 2026-05-14 at 5 41 48 pm

[B2B-3766]: https://bigcommercecloud.atlassian.net/browse/B2B-3766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Low risk: localized to translation-loading thunks/state, with additive fetching/merging of extra translation pages and minimal impact outside i18n caching/network calls.
> 
> **Overview**
> Fixes missing merchant overrides for shared components by teaching `getPageTranslations` to also fetch and merge translations from dependent pages.
> 
> Adds a `TRANSLATION_DEPENDENCIES` map (currently `quoteDetail -> ['quoteDraft']`), skips already-fetched deps, merges successful dependency results into the returned `pageTranslations`, and records both the primary and dependency pages in `lang.fetchedPages` to prevent refetching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b8270515c5d669e247b61544e57c27ec754ad9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/store/appAsyncThunks.ts
+++ b/apps/storefront/src/store/appAsyncThunks.ts
@@ -33,6 +33,14 @@ const REPEATED_PAGES: Partial<Record<string, string>> = {
   'company-orders': 'orders',
 };
 
+/**
+ * Pages that render shared components whose translation keys belong to another page.
+ * When a page is visited, translations for its dependencies are also fetched.
+ */
+const PAGE_DEPENDENCIES: Partial<Record<string, string[]>> = {
+  quoteDetail: ['quoteDraft'],
+};
+
 export const getGlobalTranslations = createAppAsyncThunk<
   GetGlobalTranslationResponse,
   GetGlobalTranslationsParams
@@ -67,13 +75,24 @@ export const getPageTranslations = createAppAsyncThunk<
   'lang/getPageTranslations',
   async ({ channelId, page: pageKey }, { rejectWithValue }) => {
     const page = REPEATED_PAGES[pageKey] ?? pageKey;
-    const { message } = await getTranslation({ channelId, page });
+    const pagesToFetch = [page, ...(PAGE_DEPENDENCIES[pageKey] ?? [])];
 
-    if (typeof message === 'string') {
-      return rejectWithValue(message);
+    const results = await Promise.all(
+      pagesToFetch.map((p) => getTranslation({ channelId, page: p })),
+    );
+
+    const errorResult = results.find((r) => typeof r.message === 'string');
+
+    if (errorResult) {
+      return rejectWithValue(errorResult.message as string);
     }
 
-    return { pageTranslations: message, page };
+    const pageTranslations = Object.assign(
+      {},
+      ...results.map((r) => r.message as Record<string, string>),
+    );
+
+    return { pageTranslations, page };
   },
   {
     condition: ({ page: pageKey }, { getState }) => {
--- a/apps/storefront/src/store/appAsyncThunks.ts
+++ b/apps/storefront/src/store/appAsyncThunks.ts
@@ -75,21 +75,22 @@ export const getPageTranslations = createAppAsyncThunk<
   'lang/getPageTranslations',
   async ({ channelId, page: pageKey }, { rejectWithValue }) => {
     const page = REPEATED_PAGES[pageKey] ?? pageKey;
-    const pagesToFetch = [page, ...(TRANSLATION_DEPENDENCIES[pageKey] ?? [])];
+    const dependencyPages = TRANSLATION_DEPENDENCIES[page] ?? [];
 
-    const results = await Promise.all(
-      pagesToFetch.map((p) => getTranslation({ channelId, page: p })),
+    const [primaryResult, ...dependencyResults] = await Promise.all(
+      [page, ...dependencyPages].map((p) => getTranslation({ channelId, page: p })),
     );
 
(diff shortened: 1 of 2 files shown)


bc-lulu-yang (approved):
LGTM. Wondering if we can have some test to cover this case.

apps/storefront/src/store/appAsyncThunks.ts
```diff
@@ -67,13 +75,25 @@ export const getPageTranslations = createAppAsyncThunk<
...
-    return { pageTranslations: message, page };
+    const pageTranslations = Object.assign(
+      {},
+      primaryResult.message as Record<string, string>,
+      ...dependencyResults
+        .filter((r) => typeof r.message !== 'string')
+        .map((r) => r.message as Record<string, string>),
+    );
+
+    return { pageTranslations, page };
```
bc-gauravsobti: We need to be able to return dependency pages as well so that they are not fetched again. It's a 2 part scenario: dependency page shouldn't be fetched as a normal page request as well as a dependency page request.

apps/storefront/src/store/appAsyncThunks.ts
```diff
@@ -67,13 +75,25 @@ export const getPageTranslations = createAppAsyncThunk<
...
 
-    if (typeof message === 'string') {
-      return rejectWithValue(message);
+    const [primaryResult, ...dependencyResults] = await Promise.all(
+      [page, ...dependencyPages].map((p) => getTranslation({ channelId, page: p })),
+    );
+
+    if (typeof primaryResult.message === 'string') {
+      return rejectWithValue(primaryResult.message);
     }
```
bc-gauravsobti, line 88: We are not catching errors with dependencyResults from the api.

apps/storefront/src/store/appAsyncThunks.ts
```diff
@@ -67,13 +75,25 @@ export const getPageTranslations = createAppAsyncThunk<
...
```
bc-gauravsobti, line 89: We are not catching any network related errors as well.

bc-gauravsobti (changes requested):

apps/storefront/src/store/appAsyncThunks.ts
bc-gary: Dependency errors are already handled — the `.filter((r) => typeof r.message \!== 'string')` on the results array silently drops any failed dependency translations so they don't break the primary page. This is intentional: if `quoteDraft` fails, `quoteDetail` translations still load fine. Only the primary page failure rejects the thunk.

apps/storefront/src/store/appAsyncThunks.ts
bc-gary: You're right — with `Promise.all`, a network error in a dependency fetch would reject the entire thunk and lose the primary page's translations too. Fixed by separating the primary fetch from dependencies and using `Promise.allSettled` for the dependency calls. Now a dependency network failure is silently skipped, same as an API error response.

bc-gauravsobti (approved):
LGTM

bc-launchbay:
success [20260519000604](https://launchbay.bigcommerce.net/projects/515/releases/285234) by Launchbay

bc-launchbay:
success [20260519000604](https://launchbay.bigcommerce.net/projects/515/releases/285234) by Launchbay

PR #887, opened 2026-05-21
fix(common): B2B-3876 normalise uppercase currency token_location

bc-veer:
Jira: [B2B-3876](https://bigcommercecloud.atlassian.net/browse/B2B-3876)

## What/Why?

Currency location in prod coming as LEFT/RIGHT instead of left/right, causing the currency direction issue.

This pull request improves the handling of currency symbol placement in the storefront by making the placement logic case-insensitive and adding comprehensive tests to cover different configurations, especially when the `token_location` is set to uppercase values like `LEFT` or `RIGHT`. This ensures that the currency symbol appears in the correct position regardless of the casing in the configuration, and that the feature flag for fixing currency symbol placement works as intended.

**Currency formatting logic improvements:**

* Updated the `ordersCurrencyFormat`, `currencyFormatConvert`, and `currencyFormat` functions in `b3CurrencyFormat.ts` to make the currency symbol placement logic case-insensitive by using `.toLowerCase()` when checking `currency_location`. [[1]](diffhunk://#diff-a94673c9a25b5464d5be3777a128c1794d45fce493121acebd889791b9e5df19L56-R56) [[2]](diffhunk://#diff-a94673c9a25b5464d5be3777a128c1794d45fce493121acebd889791b9e5df19L108-R108) [[3]](diffhunk://#diff-a94673c9a25b5464d5be3777a128c1794d45fce493121acebd889791b9e5df19L118-R118) [[4]](diffhunk://#diff-a94673c9a25b5464d5be3777a128c1794d45fce493121acebd889791b9e5df19L144-R144)

## Rollout/Rollback

Revert the PR

## Testing

* Added tests in `QuoteDetail/index.test.tsx` and `QuotesList/index.test.tsx` to verify that the currency symbol is placed on the left when the `token_location` is set to uppercase `LEFT` and the relevant feature flag is enabled. [[1]](diffhunk://#diff-f0613f5f4f8e1051966f6ace5a5269833351c30ef13c0e571a11ffc0781ec752R1621-R1653) [[2]](diffhunk://#diff-a1bb8013ae4330146235b4e69700cf4ebf98749386c79c4256ca02942f3bb5b0R1438-R1463)
* Added a new test suite in `QuoteDraft/index.test.tsx` to check that the currency symbol is placed on the right when the active currency has an uppercase `token_location` of `RIGHT`.

[B2B-3876]: https://bigcommercecloud.atlassian.net/browse/B2B-3876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Low risk: small, localized change to currency formatting and store-config normalization plus added tests; main risk is subtle display regressions for currencies with unexpected `token_location` values.
> 
> **Overview**
> Fixes quote currency symbol placement when BigCommerce returns uppercase `token_location` values by **normalizing `storeConfigs.setCurrencies` to always store `token_location` as lowercase `left`/`right`**.
> 
> Makes `b3CurrencyFormat` placement checks case-insensitive (`currency_location?.toLowerCase()`), and adds tests in `QuoteDetail`, `QuotesList`, and `QuoteDraft` to cover uppercase `LEFT`/`RIGHT` scenarios (including the feature-flagged quote placement behavior).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62b20f708cc3aa54569deacedfbe851b3da65ea8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/utils/b3CurrencyFormat.ts
+++ b/apps/storefront/src/utils/b3CurrencyFormat.ts
@@ -53,7 +53,7 @@ export const ordersCurrencyFormat = (
       decimalPart ? `${moneyFormat.decimal_token}${decimalPart}` : ''
     }`;
     const priceStr =
-      moneyFormat.currency_location === 'left'
+      moneyFormat.currency_location?.toLowerCase() === 'left'
         ? `${showCurrencyToken ? moneyFormat.currency_token : ''}${newPrice}`
         : `${newPrice}${showCurrencyToken ? moneyFormat.currency_token : ''}`;
     return priceStr;
@@ -105,7 +105,7 @@ export const currencyFormatConvert = (
         moneyFormat.thousands_token,
       )}${decimalPart ? `${moneyFormat.decimal_token}${decimalPart}` : ''}`;
       const priceStr =
-        moneyFormat.currency_location === 'left'
+        moneyFormat.currency_location?.toLowerCase() === 'left'
           ? `${showCurrencyToken ? moneyFormat.currency_token : ''}${newPrice}`
           : `${newPrice}${showCurrencyToken ? moneyFormat.currency_token : ''}`;
       return priceStr;
@@ -115,7 +115,7 @@ export const currencyFormatConvert = (
       decimalPart ? `${moneyFormat.decimal_token}${decimalPart}` : ''
     }`;
     const priceStr =
-      moneyFormat.currency_location === 'left'
+      moneyFormat.currency_location?.toLowerCase() === 'left'
         ? `${showCurrencyToken ? moneyFormat.currency_token : ''}${newPrice}`
         : `${newPrice}${showCurrencyToken ? moneyFormat.currency_token : ''}`;
     return priceStr;
@@ -141,7 +141,7 @@ export const currencyFormat = (
       decimalPart ? `${moneyFormat.decimal_token}${decimalPart}` : ''
     }`;
     const priceStr =
-      moneyFormat.currency_location === 'left'
+      moneyFormat.currency_location?.toLowerCase() === 'left'
         ? `${showCurrencyToken ? moneyFormat.currency_token : ''}${newPrice}`
         : `${newPrice}${showCurrencyToken ? moneyFormat.currency_token : ''}`;
     return priceStr;
--- a/apps/storefront/src/pages/QuoteDetail/index.test.tsx
+++ b/apps/storefront/src/pages/QuoteDetail/index.test.tsx
@@ -1618,5 +1618,38 @@ describe('when the user is a B2B customer', () => {
       const summary = screen.getByTestId('quote-summary');
       expect(within(summary).getByText('€200.00')).toBeInTheDocument();
     });
+
+    it('places the token on the left when BC config has uppercase token_location LEFT and flag is enabled', async () => {
+      const eurWithUppercaseLeft = JSON.parse(
+        JSON.stringify({ ...eurOnLeftInBcConfig, token_location: 'LEFT' }),
+      );
+
+      renderWithProviders(<QuoteDetail />, {
+        preloadedState: {
+          ...preloadedState,
+          storeConfigs: {
+            currencies: {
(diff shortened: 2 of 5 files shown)


apps/storefront/src/utils/b3CurrencyFormat.ts
```diff
@@ -115,7 +115,7 @@ export const currencyFormatConvert = (
       decimalPart ? `${moneyFormat.decimal_token}${decimalPart}` : ''
     }`;
     const priceStr =
-      moneyFormat.currency_location === 'left'
+      moneyFormat.currency_location?.toLowerCase() === 'left'
```
bc-gary, line 118: **[Blocker]** — `currencyFormat()` at line 144 is missing the same fix

The `currencyFormat` function (line 128–152) has the same `=== 'left'` comparison at **line 144** that was fixed in the other three locations, but it wasn't updated. This is the most widely used currency formatter in the codebase (~30 call sites across invoices, quotes, shopping lists, orders, quick orders, etc.).

Its `moneyFormat` comes from `currencyFormatInfo()` which passes through `currentCurrency.token_location` without normalizing (line 20), so uppercase values from the backend will cause the same misplacement bug.

Line 144 should be:
```ts
moneyFormat.currency_location?.toLowerCase() === 'left'
```

bc-veer: updated

bc-gary (approved):
Nice work @bc-veer 👏

apps/storefront/src/utils/b3CurrencyFormat.ts
bc-lulu-yang, line 20: [suggestion]

`currencyFormatInfo()` still passes `token_location` directly. Display is correct today because every formatter lowercases at compare time, but Redux / MoneyFormat objects can still hold "LEFT". 

Consider normalising here (e.g. `(token_location ?? 'left').toLowerCase()`) so stored shape matches runtime expectations and we don’t rely on every future call site remembering to lowercase.

bc-lulu-yang (approved):
Thanks Veer for working on this.

I think the cleanest long-term fix is normalising when currencies are saved: either normalizing in `setCurrencies` after `getCurrencies()` call, or fix it in the API layer when  store currency config saving(not sure how much effort requires).

bc-veer:
> I think the cleanest long-term fix is normalising when currencies are saved: either normalizing in `setCurrencies` after `getCurrencies()` call, or fix it in the API layer when store currency config saving(not sure how much effort requires).

For now, it's just display part from b2b side.
Ref - https://bigcommerce.slack.com/archives/C063NCR6XGW/p1779327717900739

apps/storefront/src/utils/b3CurrencyFormat.ts
bc-veer: The main reason for this change is to support the old quote currency objects. If ff is off then we should render the currency as per quote object. if quote saved as `RIGHT` then our conversion from redux state on bc currencies won't work. 

I will convert the token_location while saving in redux 
``` Typescript
 setCurrencies: (state, { payload }: PayloadAction<Currencies>) => {
      state.currencies = {
        ...payload,
        currencies: payload.currencies.map((currency) => ({
          ...currency,
          token_location: currency.token_location?.toLowerCase() === 'right' ? 'right' : 'left',
        })),
      };
    })

bc-lulu-yang (approved):

bc-launchbay:
success [20260521062940](https://launchbay.bigcommerce.net/projects/515/releases/285577) by Launchbay

bc-launchbay:
success [20260521062940](https://launchbay.bigcommerce.net/projects/515/releases/285577) by Launchbay

PR #888, opened 2026-05-21
feat(orders): B2B-4629 Implement Order Detail Navigation with Cursor Pagination

bc-lulu-yang:
Jira: [B2B-4629](https://bigcommercecloud.atlassian.net/browse/B2B-4629)

## What/Why?
Implements prev/next order navigation on the Order Detail page for the unified Storefront GraphQL path. Previously, the `DetailPagination` component was hidden on the unified path because it relied on offset-based list positions which don't map to cursor-based pagination. This PR introduces a dedicated `CursorDetailPagination` component that carries cursor context from the list view and enables seamless order-to-order navigation without reloading the list.

## What changed

**New:**  A self-contained prev/next navigation component `CursorDetailPagination` component for the unified SF GQL path  which use a custom hook `useCursorDetailPagination`. 

**useCursorDetailPagination**: The generic layer (pure cursor state machine) takes care of `currentIndex`, `items`, `pageInfo`, `hasPrev/hasNext`, the in-page vs boundary branching logic, `exhaustBoundary`, and `syncHistory` logic. None of that is order-specific.

**CursorDetailPagination**: The order-specific layer is `fetchAdjacentPage` (which directly calls `getCompanyOrders`/`getCustomerOrders` and knows about `isCompanyOrder`, `OrdersFiltersInput`, `sortBy`) and the hardcoded `/orderDetail/${id}` route

Key behaviours of the new CursorDetailPagination(together with the new hook):
- **In-page navigation is instant (zero API calls).** The full orders page is passed from the list view via React Router location state and cached in component state. Navigating within that page just increments/decrements the index.
- **Page-boundary navigation fetches one adjacent page.** When the user reaches the first or last order of the cached page, the component fetches the previous or next page using `before`/`after` cursor arguments, lands on the last or first order of that page respectively, and updates the cached state.
- **Graceful degradation.** Renders nothing when no cursor context is present (e.g. direct URL access), satisfying the requirement that navigation controls only appear when meaningful.
- **History sync.** Each navigation step calls `navigate(..., { replace: true })` to keep the browser history entry in sync with the current cursor state, so back/forward browser navigation is consistent.
- Supports both customer (`GetCustomerOrders`) and company (`GetCompanyOrders`) order paths via the `isCompanyOrder` flag.

**Updated: `Order.tsx`**

- `goToDetail` now receives the full current page of list items and builds the `CursorLocationState` (orders + cursors, current index, pageInfo, active filters, sortBy) before navigating. This replaces the previous approach of passing a bare index and `totalCount: -1`.
- `mapSfGqlOrderToListItem` calls now forward the edge cursor so it's available on each list item.
- `navigateToOrderDetail` now branches on `isUnifiedOrders` (covering both company and customer unified paths) instead of `isUnifiedCustomerPath`, fixing an omission where company orders didn't get the cursor state.

**Updated: `DetailPagination.tsx`**

- Removed the `isUnifiedPath` guard (`totalCount === -1` check) and its early return, since unified navigation is now fully handled by `CursorDetailPagination`. `DetailPagination` is now legacy-only and no longer needs to know about the unified path.
- Cleaned up `aria-labelledby` and the order-position label which were conditioned on `!isUnifiedPath`.

**Updated: `OrderDetail/index.tsx`**

- Conditionally renders `CursorDetailPagination` (unified path) or `DetailPagination` (legacy path) based on `isUnifiedOrders`.
- Wrapped `handlePageChange` in `useCallback` to stabilise the reference passed as `onChange`.

**Updated: `mapSfGqlOrderToListItem.ts`**

- Added an optional `cursor?: string` field to `ListItem` and the mapper function so the SF GQL edge cursor travels with each list item through to the navigation state.

**Updated: `en.json`**

- Added `orderDetail.pagination.previousOrder` and `orderDetail.pagination.nextOrder` i18n keys used as `aria-label` on the navigation buttons.

## Test coverage

`CursorDetailPagination.test.tsx` (328 lines, new) covers:

- Renders nothing with no cursor state or empty orders list
- Button enabled/disabled states at page start, middle, and end
- In-page prev/next navigation (correct `onChange` calls, button state transitions)
- Page-boundary prev: fetches with correct `last`/`before` variables, lands on last order of returned page
- Page-boundary next: fetches with correct `first`/`after` variables, lands on first order of returned page
- Boundary fetch returning empty results disables the button without calling `onChange`
- Company order path uses `GetCompanyOrders`, not `GetCustomerOrders`

## Why this approach

Three cursor-based approaches were evaluated:

**Option 1: Pass the current item's cursor and adjacent cursors from the list view to the detail view (not chosen).** This would pass only three cursors to the detail view: the previous, current, and next order. The problem is that after the first navigation step those adjacent cursors are consumed — to move again you'd need to fetch the new order's neighbours, meaning every single click triggers an API call. It also breaks down entirely at page boundaries where the adjacent cursor simply doesn't exist in the set passed over.  A branch [B2B-4629](https://github.com/bigcommerce/b2b-buyer-portal/tree/B2B-4629) created based on this Option demonstrated how much effort need to maintain and update three cursors.

**Option 2: Fetch neighbours using `first: 1, after: currentCursor` and `last: 1, before: currentCursor` (not chosen).** This lets the detail page be fully self-sufficient without any state from the list view, but it pays an API round-trip on every navigation step. Stepping through five orders in a row means five sequential fetches, each returning a single item. That makes the feature feel slower than the offset-based legacy pagination it replaces, and it puts unnecessary load on the SF GQL API for the most common interaction pattern.

**Option 3: Cache the full page of order IDs from the list view and navigate within the cached set (chosen).** When the user clicks an order, the full current page of orders with their cursors and the active filter/sort state are passed to the detail view via React Router location state. Within-page navigation (the common case) is then instant with zero API calls — just an index increment/decrement over the cached set. Only crossing a true page boundary triggers a fetch, and that fetch retrieves the full adjacent page so subsequent steps within that page are also instant. The `CursorLocationState` interface is typed explicitly so the shape is clear and can be migrated to a shared store in future without changing the component's internal logic.

The tradeoff is that cursor context is lost on direct URL access or hard refresh. The component handles this by rendering nothing in that case rather than showing broken or misleading navigation controls.

## Rollout/Rollback
This feature is covered under FF, and we can revert this PR if needed

## Testing
All the unit tests added passed

When flag on, with mock data

https://github.com/user-attachments/assets/ff208db3-d9b4-4290-9d7b-8890dab26be4

[B2B-4629]: https://bigcommercecloud.atlassian.net/browse/B2B-4629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches core order list-to-detail navigation and boundary GraphQL fetches, but behavior is gated by the unified orders feature flag with graceful fallback when list context is missing.
> 
> **Overview**
> Adds **prev/next order navigation** on Order Detail for the unified Storefront GraphQL orders path (feature-flagged), replacing the previous behavior where pagination was hidden on that path.
> 
> The list view now passes a **`CursorLocationState`** in router history: the current page of order IDs and cursors, index, `pageInfo`, filters, sort, and page size. **`mapSfGqlOrderToListItem`** carries each edge’s cursor. **`CursorDetailPagination`** plus **`useCursorDetailPagination`** navigate within the cached page without API calls; at list page boundaries they fetch the adjacent page via **`getCustomerOrders`** / **`getCompanyOrders`** and update history with **`replace: true`**. Direct URL access or missing context hides controls. Legacy **`DetailPagination`** remains for non-unified orders; unified-specific guards were removed from it.
> 
> New copy in **`en.json`** covers navigation labels and fetch errors. Broad unit tests cover the hook and component.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4274174ef84a397f12d55c103be0fc0e4017dc79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.test.tsx
+++ b/apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.test.tsx
@@ -1,0 +1,315 @@
+import {
+  graphql,
+  HttpResponse,
+  renderWithProviders,
+  screen,
+  startMockServer,
+  waitFor,
+} from 'tests/test-utils';
+
+import { CursorDetailPagination } from '@/pages/OrderDetail/components/CursorDetailPagination';
+import type { CursorLocationState } from '@/pages/OrderDetail/components/CursorDetailPagination';
+import { OrdersSortInput } from '@/shared/service/bc/graphql/orders';
+
+const { server } = startMockServer();
+
+// ---------------------------------------------------------------------------
+// MSW response builders
+// ---------------------------------------------------------------------------
+
+function makeCustomerPage(
+  orders: Array<{ orderId: number; cursor: string }>,
+  hasNext = false,
+  hasPrev = false,
+) {
+  return {
+    data: {
+      customer: {
+        orders: {
+          edges: orders.map(({ orderId, cursor }) => ({ node: { entityId: orderId }, cursor })),
+          pageInfo: {
+            hasNextPage: hasNext,
+            hasPreviousPage: hasPrev,
+            startCursor: null,
+            endCursor: null,
+          },
+        },
+      },
+    },
+  };
+}
+
+function makeCompanyPage(
+  orders: Array<{ orderId: number; cursor: string }>,
+  hasNext = false,
+  hasPrev = false,
+) {
+  return {
+    data: {
+      customer: {
+        activeCompany: {
+          orders: {
+            edges: orders.map(({ orderId, cursor }) => ({ node: { entityId: orderId }, cursor })),
+            pageInfo: {
+              hasNextPage: hasNext,
+              hasPreviousPage: hasPrev,
+              startCursor: null,
+              endCursor: null,
+            },
+            collectionInfo: null,
+          },
+        },
+      },
+    },
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Base state: 3 orders on the page, current is the middle one (index 1)
+// ---------------------------------------------------------------------------
+
+const BASE_ORDERS = [
+  { orderId: '100', cursor: 'cursor-100' },
+  { orderId: '101', cursor: 'cursor-101' },
+  { orderId: '102', cursor: 'cursor-102' },
+];
+
+const BASE_STATE: CursorLocationState = {
+  isCompanyOrder: false,
+  currentIndex: 1,
+  orders: BASE_ORDERS,
+  pageInfo: { hasNextPage: true, hasPreviousPage: true },
+  filters: {},
+  sortBy: OrdersSortInput.CREATED_AT_NEWEST,
+};
+
+function renderComponent(state: CursorLocationState | null, onChange = vi.fn()) {
+  return renderWithProviders(<CursorDetailPagination onChange={onChange} color="#000000" />, {
+    initialEntries: state ? [{ pathname: '/orderDetail/101', state }] : ['/orderDetail/101'],
+  });
+}
+
+const getPrevButton = () => screen.getByRole('button', { name: /previous order/i });
+const getNextButton = () => screen.getByRole('button', { name: /next order/i });
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('CursorDetailPagination', () => {
+  describe('when there is no cursor state (e.g. direct URL navigation)', () => {
+    it('renders nothing', () => {
+      const { result } = renderComponent(null);
+      expect(result.container).toBeEmptyDOMElement();
+    });
+  });
+
+  describe('when cursor state is present', () => {
+    it('renders a navigation landmark with two buttons', () => {
+      renderComponent(BASE_STATE);
+      expect(screen.getByRole('navigation')).toBeInTheDocument();
+      expect(getPrevButton()).toBeInTheDocument();
+      expect(getNextButton()).toBeInTheDocument();
+    });
+
+    describe('initial button enabled/disabled states', () => {
+      it('enables both buttons when in the middle of the page', () => {
+        renderComponent(BASE_STATE);
+        expect(getPrevButton()).not.toBeDisabled();
+        expect(getNextButton()).not.toBeDisabled();
+      });
+
+      it('disables prev when on the first item and hasPreviousPage is false', () => {
+        renderComponent({
+          ...BASE_STATE,
+          currentIndex: 0,
+          pageInfo: { hasNextPage: true, hasPreviousPage: false },
+        });
+        expect(getPrevButton()).toBeDisabled();
+        expect(getNextButton()).not.toBeDisabled();
+      });
+
+      it('disables next when on the last item and hasNextPage is false', () => {
+        renderComponent({
+          ...BASE_STATE,
+          currentIndex: 2,
+          pageInfo: { hasNextPage: false, hasPreviousPage: true },
+        });
+        expect(getPrevButton()).not.toBeDisabled();
+        expect(getNextButton()).toBeDisabled();
+      });
+
+      it('enables prev when on the first item but hasPreviousPage is true (page boundary)', () => {
+        renderComponent({
+          ...BASE_STATE,
+          currentIndex: 0,
+          pageInfo: { hasNextPage: false, hasPreviousPage: true },
+        });
+        expect(getPrevButton()).not.toBeDisabled();
+      });
+
+      it('enables next when on the last item but hasNextPage is true (page boundary)', () => {
+        renderComponent({
+          ...BASE_STATE,
+          currentIndex: 2,
+          pageInfo: { hasNextPage: true, hasPreviousPage: false },
+        });
+        expect(getNextButton()).not.toBeDisabled();
+      });
+    });
+
+    describe('in-page navigation (zero API calls)', () => {
+      it('calls onChange with the previous orderId', async () => {
+        const onChange = vi.fn();
+        const { user } = renderComponent(BASE_STATE, onChange);
+        await user.click(getPrevButton());
+        expect(onChange).toHaveBeenCalledExactlyOnceWith('100');
+      });
+
+      it('calls onChange with the next orderId', async () => {
+        const onChange = vi.fn();
+        const { user } = renderComponent(BASE_STATE, onChange);
+        await user.click(getNextButton());
+        expect(onChange).toHaveBeenCalledExactlyOnceWith('102');
+      });
+
+      it('disables prev after navigating back to the first item with no previous page', async () => {
+        const { user } = renderComponent({
+          ...BASE_STATE,
+          currentIndex: 1,
+          orders: BASE_ORDERS.slice(0, 2),
+          pageInfo: { hasNextPage: false, hasPreviousPage: false },
+        });
+        await user.click(getPrevButton());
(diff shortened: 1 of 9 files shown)


apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
```diff
@@ -0,0 +1,285 @@
...
+          onChange(page.orders[newIndex].orderId);
+          syncHistory(page.orders, newIndex, page.pageInfo);
+        } else {
+          exhaustBoundary('prev');
+        }
+      }
+    } catch {
+      if (currentIndex === 0) {
+        exhaustBoundary('prev');
+      }
```
bc-gary: [Suggestion] This catch block (and its mirror at L244-247 in `handleNext`) calls `exhaustBoundary` on any error, permanently disabling the navigation button and persisting that state to the route history. A transient failure (network timeout, 5xx) is indistinguishable from "no more pages" — the user must go back to the list to recover navigation.

For a v1 behind a feature flag this is fine as a fail-safe, but before GA consider leaving the button enabled on error so the user can retry:

```typescript
} catch {
  // Transient error — leave button enabled so user can retry.
  // Only the empty-result path (above) should call exhaustBoundary.
}
```

Aligns with the Cursor Bugbot comment on this same area.

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+          },
+          onChange,
+        );
+        await user.click(getNextButton());
+        await waitFor(() => {
+          expect(onChange).not.toHaveBeenCalled();
+          expect(getNextButton()).toBeDisabled();
+        });
+      });
+    });
```
bc-gary, line 363: [Suggestion] The boundary tests cover empty results (correctly disables button) and successful fetches, but there's no test for when `fetchAdjacentPage` rejects (e.g., network error). Given that the catch block permanently disables the button (see comment on CursorDetailPagination.tsx:195), a test like "prev button remains enabled/disabled after a transient fetch error" would document the expected behavior and prevent regressions regardless of how the catch block is resolved.

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
bc-lulu-yang: updated

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.test.tsx
bc-lulu-yang: updated

bc-gary:
Can we have screenshots for this one @bc-lulu-yang ?

bc-gary (approved):
Nice work @bc-lulu-yang 👏

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
bc-veer, line 1: Can we make this a generic component?

Since cursor pagination will be used in quotes and potentially other areas, it might be worth abstracting this into a reusable HOC/component for better reusability and maintainability.

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const paginationArgs =
+    mode === 'before' ? { last: pageSize, before: cursor } : { first: pageSize, after: cursor };
+
+  if (isCompanyOrder) {
+    const result = await getCompanyOrders({
+      ...paginationArgs,
+      filters: filters as CompanyOrdersFiltersInput,
+      sortBy,
+    });
+    const connection = result.data?.customer?.activeCompany?.orders;
```
bc-jeffteng: If an API error occurs and the connection is set to `null`, this function will return `null`, then the caller execute `exhaustBoundary(...)`. Is that ideal? Should we consider the error handling?

bc-lulu-yang: Good point, updated with error handing, see the latest changes

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
bc-lulu-yang: Good point. I've abstracted the generic cursor navigation logic into `useCursorDetailPagination` hook with reuse in mind. Since there's no second caller yet and the implementation is still somewhat order-specific, I've kept it co-located in the `OrderDetail` folder for now. We can always promote it to a shared location when a second use case comes.

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
```diff
@@ -0,0 +1,230 @@
...
+    onNavigate,
+    onSyncHistory,
+  });
+
+  // Graceful degradation: hide when there is no cached page context
+  // (e.g. direct URL access, missing cursors, or an empty list page).
+  if (!init?.orders?.length) return null;
+
+  return (
+    <Box role="navigation" sx={{ display: 'flex', color }}>
```
bc-gary: [nit]

Unlabeled landmark — screen readers just announce "navigation" with no context. Legacy `DetailPagination` uses `aria-labelledby` but since this component omits position text, a static `aria-label` would fill the gap.

```tsx
<Box role="navigation" aria-label={b3Lang('orderDetail.pagination.orderNavigation')} ...>
```

apps/storefront/src/pages/OrderDetail/index.tsx
```diff
@@ -367,12 +368,16 @@ function OrderDetail() {
...
-                color={customColor}
-              />
-            )}
+            {location?.state &&
+              (isUnifiedOrders ? (
+                <CursorDetailPagination
+                  key={location.key}
+                  onChange={handlePageChange}
+                  color={customColor}
+                />
```
bc-gary, line 379: [nit]

`key={location.key}` is load-bearing — forces remount from list entry but stays stable during in-page prev/next (`replace: true` preserves the key). Worth a one-line comment so nobody removes it by accident.

bc-jeffteng (approved):
LGTM

apps/storefront/src/pages/OrderDetail/index.tsx
bc-lulu-yang: updated

apps/storefront/src/pages/OrderDetail/components/CursorDetailPagination.tsx
bc-lulu-yang: updated, nice to have a11y improved

bc-launchbay:
success [20260527010933](https://launchbay.bigcommerce.net/projects/515/releases/286133) by Launchbay

bc-launchbay:
success [20260527010933](https://launchbay.bigcommerce.net/projects/515/releases/286133) by Launchbay

PR #924, opened 2026-06-17
fix(common): B2B-4912 fix account page takeover failures

bc-jeffteng:
Jira: [B2B-4912](https://bigcommercecloud.atlassian.net/browse/B2B-4912)

## What/Why?

There were three cases where the Buyer Portal failed to take over native BC account pages:

1. **Child element clicks** — some themes wrap account link text in a `<span>` inside the `<a>`. Clicking the span meant `e.target` was the span, not the anchor, so the portal never caught it.
2. **Absolute-URL account links** — some themes render account hrefs as full URLs (`https://store.example.com/account.php`). The portal only watched elements matched by configured CSS selectors, so these fell through.
3. **Direct URL navigation** — navigating straight to `/account.php?action=address_book` in the browser was skipped by the portal's startup logic. Even in cases where the portal did open, it ignored the `?action=` param and always landed on the default page instead of the right one.

I added a shared utility that resolves the nearest anchor from any click target, normalises URLs to same-origin paths, and decides whether a link belongs to the portal. That utility is now used in all three interception points: the pre-init capture listener, the runtime click handler, and the startup logic in `App.tsx`. The startup logic also now maps the native `?action=` param to the correct portal route so direct loads and link clicks land in the same place.

## Rollout/Rollback

No feature flag — this is a bug fix on existing behaviour. Rolling back is a revert of this PR.

## Testing
| Before | After |
| --- | --- |
| https://github.com/user-attachments/assets/0c474acc-9575-4363-838c-73d51f20923a | https://github.com/user-attachments/assets/0c0788c4-32cd-463e-b4c7-c43e70d930f9 |

All changes are covered by the unit tests (34 pass, 0 skipped).

[B2B-4912]: https://bigcommercecloud.atlassian.net/browse/B2B-4912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes global click capture and app bootstrap routing for account/login entry points; behavior is gated by a feature flag but still affects how users reach the portal across themes and direct URLs.
> 
> **Overview**
> Fixes cases where the Buyer Portal failed to take over native BigCommerce **account/login** navigation: clicks on nested elements inside `<a>`, same-origin absolute `account.php`/`login.php` URLs outside configured selectors, and direct loads of `/account.php` with query params.
> 
> Adds **`nativeStorefrontLinks`** helpers (`getClosestAnchorFromTarget`, `getNativeStorefrontPath`, `isBuyerPortalNativeHref`, `shouldOpenAllowedPageOnInit`) and wires them into **pre-init** (`bindLinks` document capture), **runtime** (`useB3AppOpen` window capture, always registered), and **`App.tsx` init** when feature flag **`B2B-4912.buyer_portal_native_link_interception`** is on.
> 
> On startup with the flag enabled, **`shouldOpenAllowedPageOnInit`** replaces the old checkout/logged-in-no-hash heuristic so logged-in **`/account.php`** without a hash still opens the portal; when `?action=` is present, **`openPageByClick`** maps the native URL to the correct portal route (aligned with click interception). Checkout `#` placeholder links remain excluded.
> 
> Unit tests cover the utility, `bindLinks`, and `useB3AppOpen` interception behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 637e80981b6192bbdd6408a2cb1201f07d93e22c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/hooks/useB3AppOpen.ts
+++ b/apps/storefront/src/hooks/useB3AppOpen.ts
@@ -25,6 +25,10 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
     setCheckoutRegisterNumber(() => checkoutRegisterNumber + 1);
   }, [checkoutRegisterNumber]);
   const role = useAppSelector((state) => state.company.customer.role);
+  const isNativeLinkInterceptionEnabled = useAppSelector(
+    ({ global }) =>
+      global.featureFlags['B2B-4912.buyer_portal_native_link_interception'] ?? false,
+  );
   const authorizedPages = initOpenState?.authorizedPages || '/orders';
 
   const [openPage, setOpenPage] = useState<OpenPageState>({
@@ -101,7 +105,7 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
       const nativePath = anchor ? getNativeStorefrontPath(anchor.href) : null;
       const isNativeBuyerPortalLink = anchor ? isBuyerPortalNativeHref(anchor.href) : false;
 
-      if (isConfiguredTarget || isNativeBuyerPortalLink) {
+      if (isConfiguredTarget || (isNativeLinkInterceptionEnabled && isNativeBuyerPortalLink)) {
         const isSearchNode = handleJudgeSearchNode(e);
         const isCheckoutNormalHref = handleJudgeCheckoutNormalHref(e, anchor);
 
@@ -172,7 +176,7 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
       });
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [checkoutRegisterNumber, initOpenState, role]);
+  }, [checkoutRegisterNumber, initOpenState, isNativeLinkInterceptionEnabled, role]);
 
   useMutationObservable(config['dom.checkoutRegisterParentElement'], callback);
 
--- a/apps/storefront/src/hooks/useB3AppOpen.ts
+++ b/apps/storefront/src/hooks/useB3AppOpen.ts
@@ -5,7 +5,6 @@ import config from '@/lib/config';
 import { useAppSelector } from '@/store';
 import { CustomerRole } from '@/types';
 import { OpenPageState } from '@/types/hooks';
-
 import {
   getClosestAnchorFromTarget,
   getNativeStorefrontPath,
@@ -109,7 +108,8 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
         if (isSearchNode || isCheckoutNormalHref) return false;
         e.preventDefault();
         e.stopPropagation();
-        const isRegisterArrInclude = registerArr.includes(target) || (anchor ? registerArr.includes(anchor) : false);
+        const isRegisterArrInclude =
+          registerArr.includes(target) || (anchor ? registerArr.includes(anchor) : false);
         const tagHref = nativePath || anchor?.href || (e.target as HTMLAnchorElement)?.href;
         let href = tagHref || authorizedPages;
         if (!tagHref || typeof tagHref !== 'string') {
--- a/apps/storefront/src/hooks/useB3AppOpen.ts
+++ b/apps/storefront/src/hooks/useB3AppOpen.ts
@@ -69,12 +69,15 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
     return isSearchNode;
   };
 
-  const handleJudgeCheckoutNormalHref = (element: MouseEvent) => {
+  const handleJudgeCheckoutNormalHref = (
+    element: MouseEvent,
+    anchor: HTMLAnchorElement | null,
+  ) => {
     if (window?.location?.pathname !== CHECKOUT_URL) return false;
 
     const target = element.target as HTMLAnchorElement;
 
-    if (target.getAttribute('href') && target.getAttribute('href') === '#') return true;
+    if (target.getAttribute('href') === '#' || anchor?.getAttribute('href') === '#') return true;
 
     return false;
   };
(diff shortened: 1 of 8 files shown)


apps/storefront/src/hooks/useB3AppOpen.ts
```diff
@@ -86,80 +91,86 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
...
+      const isConfiguredTarget =
+        registerArr.includes(target) ||
+        allOtherArr.includes(target) ||
+        (anchor ? registerArr.includes(anchor) || allOtherArr.includes(anchor) : false);
+      const nativePath = anchor ? getNativeStorefrontPath(anchor.href) : null;
+      const isNativeBuyerPortalLink = anchor ? isBuyerPortalNativeHref(anchor.href) : false;
+
+      if (isConfiguredTarget || isNativeBuyerPortalLink) {
+        const isSearchNode = handleJudgeSearchNode(e);
+        const isCheckoutNormalHref = handleJudgeCheckoutNormalHref(e);
```
bc-gary: [suggestion]
Now that child clicks are intercepted via anchor resolution (`isConfiguredTarget` checks both `target` and `anchor`), `handleJudgeCheckoutNormalHref` still examines `e.target` directly. If a child element inside a configured `href="#"` anchor is clicked on checkout, `target.getAttribute('href')` returns `null` not `"#"`, so the safeguard gets bypassed.

Low probability but easy fix — pass the resolved `anchor` through and check it too:
```ts
const handleJudgeCheckoutNormalHref = (element: MouseEvent, anchor: HTMLAnchorElement | null) => {
    if (window?.location?.pathname !== CHECKOUT_URL) return false;
    const target = element.target as HTMLAnchorElement;
    if (target.getAttribute('href') === '#' || anchor?.getAttribute('href') === '#') return true;
    return false;
};
```

apps/storefront/src/utils/nativeStorefrontLinks.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  it('opens on account.php without a hash regardless of native query string (pathname strips search)', () => {
+    expect(
+      shouldOpenAllowedPageOnInit({
+        pathname: '/account.php',
+        hash: '',
+        customerId: 456,
+      }),
+    ).toBe(true);
+  });
+});
```
bc-gary, line 130: [suggestion]
These last two tests have identical inputs and assertions — both pass `{ pathname: '/account.php', hash: '', customerId: 456 }`. The second test name implies it verifies a query string doesn't affect the result, but since `pathname` never includes the search string they're testing the exact same thing.

Consider removing the duplicate or making the second one meaningfully different.

bc-veer (approved):

apps/storefront/src/utils/nativeStorefrontLinks.test.ts
bc-jeffteng: fixed

apps/storefront/src/hooks/useB3AppOpen.ts
bc-jeffteng: fixed

apps/storefront/src/utils/nativeStorefrontLinks.ts
```diff
@@ -0,0 +1,63 @@
...
+  pathname,
+  hash,
+  customerId,
+}: InitOpenDecisionInput): boolean {
+  if (pathname.includes('checkout')) {
+    return false;
+  }
+
+  if (pathname.includes('account.php') && !hash) {
+    return true;
```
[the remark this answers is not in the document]
bc-jeffteng: The `includes` check predates this PR, so we extract it as is rather than changing it. The same pattern is also in App.tsx :222 for the action mapping branch. `window.location.pathname` never includes query strings or hashes, and in BigCommerce's URL structure `account.php` only ever appears at root. So the substring match and an exact match would produce identical results in practice. Leaving it as-is

bc-gary (approved):
Nice work 👏

apps/storefront/src/load-functions.ts
```diff
@@ -44,23 +45,36 @@ export const initApp = async () => {
...
+
+  await clickLink(e);
+};
+
 export const bindLinks = () => {
   const links: NodeListOf<HTMLAnchorElement> = document.querySelectorAll(
     `${config['dom.registerElement']}, ${config['dom.allOtherElement']}`,
   );
 
+  document.addEventListener('click', clickNativeBuyerPortalLink, { capture: true });
```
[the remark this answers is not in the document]
bc-jeffteng: `clickNativeBuyerPortalLink` in `bindLinks()` run before Redux exists, so it's can't check the FF. With the FF `off`, it intercepts the click and loads React unnecessarily before the replay navigates to `account.php`. 

But there is not user impact: they still reach account.php and the portal opens there via `shouldOpenAllowedPageOnInit`. 

Also the flag tend to be `on` and only turn off when the incident happen, so it should be fine to leave it.

bc-veer (commented):
LGTM

bc-gary (approved):

bc-launchbay:
success [20260626053318](https://launchbay.bigcommerce.net/projects/515/releases/289408) by Launchbay

bc-launchbay:
success [20260626053318](https://launchbay.bigcommerce.net/projects/515/releases/289408) promoted by https://launchbay.bigcommerce.net/projects/515/releases/289549 being deployed. Deployed by Launchbay

PR #933, opened 2026-06-23
fix(register): B2B-5156 Buyer Portal: Clean up Log out issue

bc-vasu:
Jira: [B2B-5156](https://bigcommercecloud.atlassian.net/browse/B2B-5156)

## What/Why?

Fixes a bug on the Login page where the logout effect could fire multiple times, the second run wiping a freshly-fetched storefront token and breaking login until a manual page refresh.

In `Login/index.tsx`

A useEffect handles the loginFlag query param and decides whether to log the user out:

If `isLoggedIn && loginFlag === 'loggedOutLogin' `→ logout with banner.
If `!isLoggedIn && SHOULD_LOGOUT_FLAGS.includes(loginFlag)` → logout without banner.
The problem: calling `logout()` clears the company slice, which flips the customer role to GUEST. 

That makes `isLoggedIn` become false, the effect re-runs, and the `!isLoggedIn` branch fires a second logout. 
This second logout clears the `bcGraphqlToken` that the login flow had just fetched, leaving login broken until refresh.

### **Fix**
`Login/index.tsx`

Added a `handledLogoutFlagRef` ref that records which loginFlag has already triggered a logout, so logout runs once per flag.
Merged the two logout conditions into a single shouldLogout expression, then gated the actual `logout()` call on `handledLogoutFlagRef.current !== loginFlag`.
Preserved the original banner behavior: the logout banner only shows when an actually-logged-in user signs out `(showLogoutBanner: isLoggedIn && loginFlag === 'loggedOutLogin')`.

## Rollout/Rollback

Revert

## Testing

After Fix

https://github.com/user-attachments/assets/be772bcb-51dd-4ea9-aff8-df949884d2f1

Before Fix

https://github.com/user-attachments/assets/1d69059f-a6f1-4304-a484-c1f1f97a86a1

[B2B-5156]: https://bigcommercecloud.atlassian.net/browse/B2B-5156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches login/logout and token lifecycle on a critical path, but the change is a narrow guard with preserved behavior and test coverage.
> 
> **Overview**
> Fixes a **double-logout** on the Login page when `loginFlag` (e.g. `loggedOutLogin`) triggers `logout()` in a `useEffect` that depends on `isLoggedIn`.
> 
> The first logout clears auth state and sets `isLoggedIn` to false, which **re-runs the effect** and could hit the guest `SHOULD_LOGOUT_FLAGS` path a second time, clearing a freshly set `bcGraphqlToken` and breaking sign-in until refresh.
> 
> **`Login/index.tsx`** adds `handledLogoutFlagRef` so logout runs **once per flag**, unifies the two logout branches into `shouldLogout`, and keeps **banner behavior** (`showLogoutBanner` only when a logged-in user hits `loggedOutLogin`).
> 
> **`index.test.tsx`** adds a regression test that simulates logout clearing the company slice and asserts `logout` is called exactly once.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit efed0cfa8701da9a61f8b91c3790a23a13c054c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/Login/index.tsx
+++ b/apps/storefront/src/pages/Login/index.tsx
@@ -56,6 +56,13 @@ function Login(props: PageProps) {
 
   const [isLoading, setLoading] = useState(true);
   const isSubmittingRef = useRef(false);
+  /**
+    * Tracks which loginFlag has already triggered logout so it only runs once.
+    Without this, the first logout flips isLoggedIn to false,
+    the effect fires again, and a second logout clears the freshly fetched bcGraphqlToken,
+    breaking login until refresh.
+   */
+  const handledLogoutFlagRef = useRef<LoginFlagType | null>(null);
   const [isMobile] = useMobile();
 
   const [showTipInfo, setShowTipInfo] = useState<boolean>(true);
@@ -107,11 +114,19 @@ function Login(props: PageProps) {
         if (isLoginFlagType(loginFlag)) {
           setLoginFlag(loginFlag);
 
-          if (isLoggedIn && loginFlag === 'loggedOutLogin') {
-            await logout({ showLogoutBanner: true });
-            // All company-related flags have isLoggedIn set to false.
-          } else if (!isLoggedIn && SHOULD_LOGOUT_FLAGS.includes(loginFlag)) {
-            await logout({ showLogoutBanner: false });
+          const shouldLogout =
+            (isLoggedIn && loginFlag === 'loggedOutLogin') ||
+            (!isLoggedIn && SHOULD_LOGOUT_FLAGS.includes(loginFlag));
+
+          /* 
+            Guard against a second logout: logging out flips isLoggedIn to
+            false and re-runs this effect, which would otherwise re-trigger
+            logout via the !isLoggedIn branch. Only log out once per flag.
+          */
+          if (shouldLogout && handledLogoutFlagRef.current !== loginFlag) {
+            handledLogoutFlagRef.current = loginFlag;
+            // Banner only when an actually-logged-in user is signing out.
+            await logout({ showLogoutBanner: isLoggedIn && loginFlag === 'loggedOutLogin' });
           }
         }
 
--- a/apps/storefront/src/pages/Login/index.tsx
+++ b/apps/storefront/src/pages/Login/index.tsx
@@ -56,6 +56,13 @@ function Login(props: PageProps) {
 
   const [isLoading, setLoading] = useState(true);
   const isSubmittingRef = useRef(false);
+  /**
+   * Tracks which loginFlag has already triggered logout so it only runs once.
+   * Without this, the first logout flips isLoggedIn to false,
+   * the effect fires again, and a second logout clears the freshly fetched bcGraphqlToken,
+   * breaking login until refresh.
+   */
+  const handledLogoutFlagRef = useRef<LoginFlagType | null>(null);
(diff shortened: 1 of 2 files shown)


bc-jeffteng (approved):
LGTM

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -107,11 +114,19 @@ function Login(props: PageProps) {
...
+            (isLoggedIn && loginFlag === 'loggedOutLogin') ||
+            (!isLoggedIn && SHOULD_LOGOUT_FLAGS.includes(loginFlag));
+
+          /* 
+            Guard against a second logout: logging out flips isLoggedIn to
+            false and re-runs this effect, which would otherwise re-trigger
+            logout via the !isLoggedIn branch. Only log out once per flag.
+          */
+          if (shouldLogout && handledLogoutFlagRef.current !== loginFlag) {
+            handledLogoutFlagRef.current = loginFlag;
```
bc-gary, line 127: [question]
`handledLogoutFlagRef` is set once and never cleared. If the URL somehow changed from `?loginFlag=loggedOutLogin` to a different flag and back to `loggedOutLogin` within the same component lifecycle, the second visit would get skipped because the ref already holds that value.

Probably fine since login-flag navigation is typically one-shot, but worth a quick sanity check that there's no scenario where the same flag legitimately needs to re-trigger.

apps/storefront/src/pages/Login/index.tsx
```diff
@@ -56,6 +56,13 @@ function Login(props: PageProps) {
 
   const [isLoading, setLoading] = useState(true);
   const isSubmittingRef = useRef(false);
+  /**
+    * Tracks which loginFlag has already triggered logout so it only runs once.
+    Without this, the first logout flips isLoggedIn to false,
+    the effect fires again, and a second logout clears the freshly fetched bcGraphqlToken,
+    breaking login until refresh.
+   */
```
bc-gary, line 64: [nit]
JSDoc indentation is off — first line has `* ` prefix, rest don't. Either give all lines `* ` or switch to a `//` block. 💅

apps/storefront/src/pages/Login/index.tsx
bc-vasu: Tracks only the last processed flag, so any squence will still retrigger logout.
The only suppressed case is the same flag firing twice in a row, which is the bug we’re preventing: the first logout flips isLoggedIn, the effect runs again, and a second logout would wipe the fresh bcGraphqlToken.

A real same‑flag logout requires the user to log back in, which unmounts Login and resets the ref, so no valid retrigger is skipped.

bc-arezoo (approved):

bc-veer (approved):

bc-launchbay:
success [20260624012316](https://launchbay.bigcommerce.net/projects/515/releases/289056) by Launchbay

bc-launchbay:
success [20260624012316](https://launchbay.bigcommerce.net/projects/515/releases/289056) by Launchbay

PR #937, opened 2026-06-26
feat(orders): B2B-5157 Fix currency formatting on unified orders list for multi-currency stores

bc-lulu-yang:
Jira: [B2B-5157](https://bigcommercecloud.atlassian.net/browse/B2B-5157)

## What/Why?
The current unified SF GQL order list displays incorrect grand-total, for example, a USD order on a store with AUD as the default currency renders as A$90 instead of $90.00. This PR is to fix the display price issue in unified SF GQL order list when an order’s transaction currency differs from the shopper’s active session currency.

This incorrect grand-total already in current mobile version, even when unified order flag off. This PR includes the fix for mobile version too.

## Rollout/Rollback

Behind FF, and we can revert this PR if needed

## Testing

**Before**: 
**Flag off - legacy order**
Before - this order is $A174.05, but we can see the mobile version display $174.05, which is incorrect.

https://github.com/user-attachments/assets/b15abe40-a7a5-44be-b2ea-a516d5492ed4

After: Thanks @bc-gary for helping test local. 

**Desktop**
image

**Mobile**
image

[B2B-5157]: https://bigcommercecloud.atlassian.net/browse/B2B-5157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Display-only currency formatting with unit tests; legacy list still falls back to session formatting when `money` is absent.
> 
> **Overview**
> Fixes **incorrect grand totals** on the unified SF GQL order list (and mobile cards) when an order’s transaction currency differs from the shopper’s session/default currency.
> 
> Introduces shared **`orderMoneyUtils`**: resolves `MoneyFormat` from store currencies by order currency code (case-insensitive), builds the legacy double-JSON `money` string unified rows were missing, and **`formatOrderListGrandTotal`** for table and mobile display. Unified **`mapSfGqlOrderToListItem`** now takes store currencies and sets `money` from `order.totalIncTax.currencyCode` instead of an empty string. **`Order.tsx`** reads currencies from Redux, passes them into the mapper, includes `currencies` in unified list query keys, and routes grand-total rendering through the helper. **`convertOrderDetail`** drops its local money-format builder in favor of the same shared lookup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ede53df555d289d657696c5edbf3070a8a93b607. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bc-gary:
Currency configs

**Desktop**
Screenshot 2026-06-29 at 2 03 49 pm

**Mobile**
Screenshot 2026-06-29 at 2 23 32 pm

apps/storefront/src/pages/order/mapSfGqlOrderToListItem.ts
```diff
@@ -11,13 +14,17 @@ export interface ListItem {
   status: string;
   statusText?: string;
   createdAt: string;
-  companyName: string;
+  companyName?: string;
```
bc-gary, line 17: [question]
`companyName` changed from `string` to `string | undefined` — is there anything downstream (table column render, mobile card) that expects this to always be a string? Probably fine but worth a quick check.

bc-lulu-yang: Good question. I changed it is due to delete interface ListItem in `OrderItemCard`, which doesn't include the companyName. But after checking, seems nothing downstream requires top-level `companyName` to always be a string. 

Desktop version: The company column reads companyInfo?.companyName, not top-level `companyName`

Mobile card: `OrderItemCard` does not use `companyName` or `companyInfo` at all.

apps/storefront/src/pages/order/shared/orderMoneyUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function buildLegacyOrderListMoneyString(
+  currencies: Currency[],
+  currencyCode: string,
+): string {
+  return JSON.stringify(JSON.stringify(buildMoneyFormat(currencies, currencyCode)));
+}
+
+/** Formats a list row grand total using order-specific money when present, else session currency. */
+export function formatOrderListGrandTotal(totalIncTax: string | number, money?: string): string {
```
bc-akshatkaushik, line 47: [suggestion]

These methods seem a bit out of place for a file called `buildMoneyFormat` perhaps can rename this shared utility to be something a bit more generic like `utils/currency.ts` or `utils/money.ts`

bc-lulu-yang: updated, since these helpers only apply to order, renamed this file into `orderMoneyUtils.ts`

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
```diff
@@ -20,34 +19,13 @@ import type {
   OrderSummary,
 } from '@/types';
 
+import { getMoneyFormatByCurrencyCode } from '../../order/shared/orderMoneyUtils';
```
bc-jeffteng, line 22: [nit]
Seems like loading a share utility from different domain/page folder. Do we consider to create something like `shared/` under `pages/`?

apps/storefront/src/pages/order/shared/orderMoneyUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+  currencies: Currency[],
+  currencyCode: string,
+): string {
+  return JSON.stringify(JSON.stringify(getMoneyFormatByCurrencyCode(currencies, currencyCode)));
+}
+
+/** Formats a list row grand total using order-specific money when present, else session currency. */
+export function formatOrderListGrandTotal(totalIncTax: string | number, money?: string): string {
+  if (money) {
+    return ordersCurrencyFormat(JSON.parse(JSON.parse(money)), totalIncTax);
```
bc-jeffteng, line 49: [question]
`money` is only a string type, are we sure it always valid for `JSON.parse()` ?

Also, why double JSON encoding?

bc-lulu-yang:
will use another solution

PR #947, opened 2026-06-30
fix(common): B2B-4912 fix pending-approval loginFlag dropped on init navigation

bc-jeffteng:
Jira: [B2B-4912](https://bigcommercecloud.atlassian.net/browse/B2B-4912)

## What/Why?

PR #924 introduced a double-navigation bug in `App.tsx` `init()`: when a pending/inactive-company buyer visited the storefront, the `CompanyError` catch called `gotoPage('/login?loginFlag=…')`, but execution continued into the open-allowed-page block which called `gotoPage(...)` again, overwriting the loginFlag (last-write-wins). The pending-approval alert never showed.

**Fix:** Instead of navigating inside the catch, capture the error reason as `companyLoginFlag`. Then make a single navigation decision via a new pure function `resolveInitNavigation`, which returns a typed `InitNavigation` value. Company errors take priority (returned before any other branch), so the loginFlag can never be overwritten by a subsequent navigation.

**Also fixed:** `authorizedPages` was computed at render time from the Redux `role`, but `getCurrentCustomerInfo()` can update `userInfo.role` (e.g. B2C → SUPER_ADMIN) before `resolveInitNavigation` is called. `openPageByClick` returns `authorizedPages` verbatim for `SUPER_ADMIN`/`JUNIOR_BUYER`, so a stale value would land them on `/orders` instead of `/dashboard` or `/shoppingLists`. Fixed by recomputing `authorizedPages` from `userInfo.role` after `getCurrentCustomerInfo()` completes.

## Rollout/Rollback

No new feature flags introduced. The fix is a refactor of the existing `init()` navigation logic — same observable behavior for healthy users, corrected behavior for pending/inactive-company buyers.

Rollback: revert the two commits on this branch (`619baf18`, `0db5f209`).

## Testing

**Functional Test** Has been tested on integration with 2 previous failed testcase:
-  `tests/ui/domain/b2b/buyer-portal/auth/UnapprovedCompanyLogin.spec.ts`
- `tests/ui/domain/b2b/checkout/InvoicePayment/InvoicePaymentWithCompanyAdmin.spec.ts`

**Unit:** `yarn test run src/utils/resolveInitNavigation.test.ts` — 6/6 passing. Full suite (`yarn test run`) — 1163/1163 passing.

**E2E (manual / CI gate):** `functional-tests` repo — `tests/ui/domain/b2b/buyer-portal/auth/UnapprovedCompanyLogin.spec.ts`, Step 1 ("login-as-customer redirect is blocked"). Run against a store with a **pending** company, with flag `B2B-4912.buyer_portal_native_link_interception` **ON** and **OFF**.

Expected: final URL contains `#/login?loginFlag=<reason>` and the pending-approval alert renders in both flag states.

[B2B-4912]: https://bigcommercecloud.atlassian.net/browse/B2B-4912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes startup routing and global click handling for login/account flows; wrong ordering could still mis-route pending buyers or break theme-specific account links, though behavior is flag-gated and covered by new unit tests.
> 
> **Overview**
> Fixes **init navigation** so pending/inactive company buyers are not sent to `#/login?loginFlag=…` and then immediately overwritten by a second `gotoPage`. `App.tsx` now stores `companyLoginFlag` from `getCurrentCustomerInfo` errors and routes through **`resolveInitNavigation`**, which prioritizes that login URL over account-page mapping or `gotoAllowedAppPage`. **`authorizedPages`** is recomputed from the post-fetch `userInfo.role` so roles like SUPER_ADMIN land on the right default route.
> 
> Adds **native storefront link interception** behind `B2B-4912.buyer_portal_native_link_interception`: shared helpers in `nativeStorefrontLinks` (anchor resolution, `account.php`/`login.php` detection, init open rules, localStorage cache for pre-React `main.ts`). **`bindLinks`/`useB3AppOpen`** intercept same-origin account/login links (including child-of-anchor clicks) while leaving configured DOM targets on their full href when the flag is off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79497e2e86767e4a95b18492b1a07c5e89b436f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -216,9 +216,8 @@ export default function App() {
 
         // background login enter judgment and refresh
         const nativeLinkInterceptionEnabled =
-          store.getState().global.featureFlags[
-            'B2B-4912.buyer_portal_native_link_interception'
-          ] ?? false;
+          store.getState().global.featureFlags['B2B-4912.buyer_portal_native_link_interception'] ??
+          false;
         const shouldOpenAllowedPage = nativeLinkInterceptionEnabled
           ? shouldOpenAllowedPageOnInit({ pathname, hash: window.location.hash, customerId })
           : !pathname.includes('checkout') && !(!!customerId && !window.location.hash);
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -216,7 +216,7 @@ export default function App() {
           }
         }
 
-        const resolvedAuthorizedPages = isB2BUser
+        const resolvedAuthorizedPages = isB2BUserSelector(store.getState())
           ? b2bJumpPath(Number(userInfo.role))
           : PATH_ROUTES.ORDERS;
 
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -22,6 +22,7 @@ import { handleHideRegisterPage } from '@/utils/b3HideRegister';
 import { hideStorefrontElement } from '@/utils/b3HideStorefrontElement';
 import { getQuoteEnabled } from '@/utils/b3Init';
 
+import { shouldOpenAllowedPageOnInit } from '@/utils/nativeStorefrontLinks';
 import b2bVerifyBcLoginStatus from './utils/b2bVerifyBcLoginStatus';
 import { b2bJumpPath } from './utils/b3CheckPermissions/b2bPermissionPath';
 import clearInvoiceCart from './utils/b3ClearCart';
@@ -213,8 +214,15 @@ export default function App() {
         }
 
         // background login enter judgment and refresh
-        if (!pathname.includes('checkout') && !(customerId && !window.location.hash)) {
-          await gotoAllowedAppPage(Number(userInfo.role), gotoPage);
+        const shouldOpenAllowedPage = shouldOpenAllowedPageOnInit({
+          pathname,
+          hash: window.location.hash,
+          customerId,
+        });
+        const isAccountPageWithoutHash = pathname.includes('account.php') && !window.location.hash;
+
+        if (shouldOpenAllowedPage) {
+          await gotoAllowedAppPage(Number(userInfo.role), gotoPage, isAccountPageWithoutHash);
         } else {
           showPageMask(false);
         }
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -22,6 +22,7 @@ import { handleHideRegisterPage } from '@/utils/b3HideRegister';
 import { hideStorefrontElement } from '@/utils/b3HideStorefrontElement';
 import { getQuoteEnabled } from '@/utils/b3Init';
 import { shouldOpenAllowedPageOnInit } from '@/utils/nativeStorefrontLinks';
+import { resolveInitNavigation } from '@/utils/resolveInitNavigation';
 
 import b2bVerifyBcLoginStatus from './utils/b2bVerifyBcLoginStatus';
 import { b2bJumpPath } from './utils/b3CheckPermissions/b2bPermissionPath';
@@ -203,16 +204,21 @@ export default function App() {
           role: Number(role),
           isAgenting,
         };
+        let companyLoginFlag: string | null = null;
         if (!customerId) {
           const info = await getCurrentCustomerInfo().catch((error) => {
             if (isCompanyError(error)) {
-              gotoPage(`/login?loginFlag=${error.reason}`);
+              companyLoginFlag = error.reason;
             }
           });
           if (info) {
             userInfo.role = info?.role;
           }
         }
+
+        const resolvedAuthorizedPages = isB2BUser
+          ? b2bJumpPath(Number(userInfo.role))
+          : PATH_ROUTES.ORDERS;
 
         // background login enter judgment and refresh
         const nativeLinkInterceptionEnabled =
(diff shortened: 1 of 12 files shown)


apps/storefront/src/App.tsx
```diff
@@ -201,20 +204,46 @@ export default function App() {
...
-              gotoPage(`/login?loginFlag=${error.reason}`);
+              companyLoginFlag = error.reason;
             }
           });
           if (info) {
             userInfo.role = info?.role;
           }
         }
 
+        const resolvedAuthorizedPages = isB2BUser
```
bc-gary: [Question]

Just curious why the cursor review has been resolved. I'm seeing that `isB2BUser` is still the render-time Redux selector value. If a user transitions from B2B --> SUPER_ADMIN during init, `isB2BUser` will still be false. And in this case, it picks up `PATH_ROUTES.ORDERS`.

bc-jeffteng: fixed

apps/storefront/src/hooks/useB3AppOpen.ts
```diff
@@ -86,82 +94,88 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
...
+      if (isConfiguredTarget || (isNativeLinkInterceptionEnabled && isNativeBuyerPortalLink)) {
+        const isSearchNode = handleJudgeSearchNode(e);
+        const isCheckoutNormalHref = handleJudgeCheckoutNormalHref(e, anchor);
+
+        if (isSearchNode || isCheckoutNormalHref) return false;
+        e.preventDefault();
+        e.stopPropagation();
+        const isRegisterArrInclude =
+          registerArr.includes(target) || (anchor ? registerArr.includes(anchor) : false);
+        const tagHref = nativePath || anchor?.href || (e.target as HTMLAnchorElement)?.href;
```
bc-gary: [question]
`nativePath` is computed for every same-origin anchor, not just native buyer-portal links — so for pre-configured elements pointing at e.g. `https://store.example.com/#/orders`, this changes `tagHref` from the full URL to the normalized relative path `/#/orders`. Downstream `split('/#')` handles both, so it works, but was this intentional?

If not, could narrow it to only apply for native links:
```ts
const tagHref =
  (isNativeBuyerPortalLink ? nativePath : null) ||
  anchor?.href ||
  (e.target as HTMLAnchorElement)?.href;
```

bc-gary (approved):

bc-veer:
@bc-jeffteng Let's sync up on this.
It looks like this fix overlaps with, or is very similar to, the changes in this PR - https://github.com/bigcommerce/b2b-buyer-portal/pull/953

apps/storefront/src/main.ts
bc-jeffteng, line 1: The code change in this file is only for the short gap window, which is when the page has loaded but the React layer has not yet loaded. I use local storage to persist the feature flag we fetched **last time**. Then, we retrieve the value from local storage to fill in the parameter for calling the `bindLinks()/unbindLinks()`. 

This means that when the page is first loading, this short gap window might still have the failed result, but it's quite short. It only happens before the React layer is loaded, and second, this is only for the very first time because it's behind a feature flag. So, I think this is enough for now. 

I also made a comment just to make sure that when we clean up the feature flag, we need to clean up this area as well.

apps/storefront/src/hooks/useB3AppOpen.ts
bc-jeffteng: code udpated

apps/storefront/src/hooks/useB3AppOpen.ts
```diff
@@ -86,82 +94,91 @@ const useB3AppOpen = (initOpenState: OpenPageState) => {
...
-              href = parentHref || authorizedPages;
+    const handleTriggerClick = (e: MouseEvent) => {
+      const anchor = getClosestAnchorFromTarget(e.target);
+      const target = e.target as Element;
+      const isConfiguredTarget =
+        registerArr.includes(target) ||
+        allOtherArr.includes(target) ||
+        (anchor ? registerArr.includes(anchor) || allOtherArr.includes(anchor) : false);
+      const nativePath = anchor ? getNativeStorefrontPath(anchor.href) : null;
+      const isNativeBuyerPortalLink = anchor ? isBuyerPortalNativeHref(anchor.href) : false;
```
bc-gary: [question] These two lines run on every click regardless of `isNativeLinkInterceptionEnabled`. For configured elements pointing at `account.php`, `isNativeBuyerPortalLink` ends up true, so `tagHref` becomes the relative path (`/account.php`) instead of the full URL — even with the flag off.

Downstream `split('/#')` and the login checks handle both formats so it works, but it's a behaviour change outside the flag gate. Was this intentional, or should these be gated on `isNativeLinkInterceptionEnabled`?

apps/storefront/src/utils/nativeStorefrontLinks.test.ts
```diff
@@ -0,0 +1,166 @@
...
+  it('opens on account.php order status without a hash for a logged-in customer', () => {
+    expect(
+      shouldOpenAllowedPageOnInit({
+        pathname: '/account.php',
+        hash: '',
+        customerId: 456,
+      }),
+    ).toBe(true);
+  });
+});
```
bc-gary, line 172: [suggestion] `shouldOpenAllowedPageOnInit` special-cases `account.php` (always open if no hash) but `login.php` falls through to the generic check. Might be worth adding a test to document that asymmetry:

```ts
it('does not force-open on login.php for a logged-in user without a hash', () => {
  expect(
    shouldOpenAllowedPageOnInit({ pathname: '/login.php', hash: '', customerId: 123 }),
  ).toBe(false);
});
```

apps/storefront/src/utils/nativeStorefrontLinks.ts
```diff
@@ -0,0 +1,93 @@
+interface InitOpenDecisionInput {
+  pathname: string;
+  hash: string;
+  customerId?: number | string;
+}
+
+const NATIVE_BUYER_PORTAL_PATHS = ['/account.php', '/login.php'];
```
bc-lulu-yang: [Nit]

The const name is a bit misleading. From my understanding, `/account.php` and `/login.php` are native storefront paths, not buyer-portal routes like `/orders` or `/login`. In buyer portal maps them to its own routes (e.g. `/account.php?action=order_status` → `/orders`) via `openPageByClick`.

considering to rename it to make it easy to understand (for example, `NATIVE_STOREFRONT_PATH_SUFFIXES`)

bc-lulu-yang (approved):

apps/storefront/src/hooks/useB3AppOpen.ts
bc-jeffteng: updated

bc-gary (approved):
LGTM. Nice work 👏

bc-launchbay:
success [20260706041205](https://launchbay.bigcommerce.net/projects/515/releases/290281) by Launchbay

bc-launchbay:
success [20260706041205](https://launchbay.bigcommerce.net/projects/515/releases/290281) by Launchbay

PR #952, opened 2026-07-01
fix(companies): B2B-5056 Fix register company flow for Catalyst companies

shan-pandian:
Jira: [B2B-5056](https://bigcommercecloud.atlassian.net/browse/B2B-5056)

### Problem

The registerCompany BC Storefront GraphQL mutation was returning "Customer not authenticated" when submitted from a Catalyst storefront. The same flow worked correctly on Stencil.

Root cause: three gaps in how the buyer portal handled BC Storefront requests in a headless/cross-origin context.

* getBCGraphqlToken was guarded by platform !== 'bigcommerce', so the store-level BC Storefront bearer token (bcGraphqlToken) was never fetched in Catalyst. Without it, storefrontGQLRequest always routed to graphqlBCProxy rather than directly to BC.
* storefrontGQLRequest always used the proxy for non-Stencil platforms, even when a bearer token was available. In the proxy path, the guest request carries no customer authentication, so BC rejects registerCompany.
* graphqlBC did not include credentials: 'include', so the BC session cookie set by bcLogin was silently dropped by the browser on cross-origin responses and never sent on subsequent requests.
* Refactoring to remove hardcoded platform values everywhere

In Stencil, none of these matter: the page is served from the BC domain, making all Storefront API calls same-origin, so cookies work automatically and the bearer token is always fetched.

### Solution

Mirror the Stencil authentication flow for Catalyst:

*  Extend getBCGraphqlToken to run for platform === 'catalyst'. The storeFrontToken B2B mutation requires no authentication token, so this works before the user has a B2B session.
Route storefrontGQLRequest to graphqlBC (direct to BC) whenever bcGraphqlToken is present in the store, regardless of platform.
*  Add credentials: 'include' to graphqlBC fetches so the browser stores the BC session cookie returned by bcLogin and sends it on the registerCompany request. BC allows credentialed cross-origin requests from the Catalyst origin because storeFrontToken sets allowedCorsOrigins: [window.location.origin].
* Read bc_graphql_domain from window.B3.setting (injected by Catalyst) when building * BigCommerceStorefrontAPIBaseURL, so requests resolve correctly in local dev environments that use a non-production BC domain.

### Flow after this change

* ensureBcStorefrontGraphqlToken() → calls B2B storeFrontToken mutation → bcGraphqlToken stored in Redux
bcLogin → routes via graphqlBC directly to store-{hash}-{channel}.{domain}/graphql with credentials: 'include' → BC sets session cookie, browser stores it
*  registerCompany → same graphqlBC path, bearer token + session cookie → BC authenticates the customer 

## Rollout/Rollback
Revert 

## Testing

Manual testing of company registration on both Catalyst storefront and b2b-buyers portal

https://github.com/user-attachments/assets/8dcc4a55-cb93-4b41-9469-828b639b6e24

https://github.com/user-attachments/assets/37e4c2ae-2737-4930-af10-3c89cf9187b5

[B2B-5056]: https://bigcommercecloud.atlassian.net/browse/B2B-5056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Touches authentication (login/logout, session cookies, token prefetch) and company registration GraphQL paths across Catalyst and Stencil; regressions could leave stale BC sessions or break registration.
> 
> **Overview**
> Fixes **Catalyst company registration** failing with "Customer not authenticated" by aligning headless BC Storefront calls with Stencil: **`getBCGraphqlToken`** now runs for Catalyst via **`isBcStorefrontPlatform`**, **`bcLogin` / `bcLogoutLogin` / `registerCompany`** use direct **`graphqlBC`** (not the cookie-less proxy), and **`graphqlBC`** sends **`credentials: 'include'`** so the session cookie from login is stored and sent cross-origin.
> 
> **Config and routing:** **`PLATFORM_*` constants**, optional **`bc_graphql_domain`** for BC GraphQL host URLs, and **`bc_graphql_domain`** typed on **`window.B3.setting`**.
> 
> **Logout:** **`useLogout`**, **`CatalystLogin`**, and **`HeadlessController` `user.logout`** prefetch **`bcGraphqlToken`** with **`ensureBcGraphqlToken`** before **`bcLogoutLogin`**, then call BC logout so the storefront session cookie is cleared; tests cover token prefetch and logout behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd29446744003eff97be17839671847a6e5d90a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/HeadlessController/getSku.tsx
+++ b/apps/storefront/src/HeadlessController/getSku.tsx
@@ -2,7 +2,7 @@ import { z } from 'zod';
 
 import B3Request from '@/shared/service/request/b3Fetch';
 import { LineItem } from '@/utils/b3Product/b3Product';
-import { platform } from '@/utils/basicConfig';
+import { platform, PLATFORM_BIGCOMMERCE } from '@/utils/basicConfig';
 
 const Options = z.array(
   z.object({
@@ -25,7 +25,7 @@ const graphqlRequest: typeof B3Request.graphqlBC | typeof B3Request.graphqlBCPro
   query,
   variables,
 }) => {
-  return platform === 'bigcommerce'
+  return platform === PLATFORM_BIGCOMMERCE
     ? B3Request.graphqlBC({ query, variables })
     : B3Request.graphqlBCProxy({ query, variables });
 };
--- a/apps/storefront/src/shared/service/bc/graphql/client.ts
+++ b/apps/storefront/src/shared/service/bc/graphql/client.ts
@@ -1,4 +1,4 @@
-import { platform } from '@/utils/basicConfig';
+import { platform, PLATFORM_BIGCOMMERCE } from '@/utils/basicConfig';
 
 import B3Request from '../../request/b3Fetch';
 
@@ -6,7 +6,7 @@ export function storefrontGQLRequest<T = any>(data: {
   query: string;
   variables?: object;
 }): Promise<T> {
-  return platform === 'bigcommerce'
+  return platform === PLATFORM_BIGCOMMERCE
     ? B3Request.graphqlBC<T>(data)
     : B3Request.graphqlBCProxy<T>(data);
 }
--- a/apps/storefront/src/shared/service/bc/graphql/login.ts
+++ b/apps/storefront/src/shared/service/bc/graphql/login.ts
@@ -1,3 +1,4 @@
+import { isBcStorefrontPlatform } from '@/utils/basicConfig';
 import { mapToCompanyError } from '@/utils/companyUtils';
 
 import B3Request from '../../request/b3Fetch';
@@ -76,13 +77,16 @@ interface LoginVariables {
   password: string;
 }
 
+// bcLogin and bcLogoutLogin must go through graphqlBC (direct to BC Storefront with
+// credentials:include) on platforms where the session cookie matters (bigcommerce, catalyst).
+// Using storefrontGQLRequest would route Catalyst through the proxy, which never sets
+// or sends the BC session cookie that registerCompany requires.
 export const bcLogin = (variables: LoginVariables) =>
-  storefrontGQLRequest({
-    query: getBcLogin(),
-    variables,
-  });
+  isBcStorefrontPlatform
+    ? B3Request.graphqlBC({ query: getBcLogin(), variables })
+    : storefrontGQLRequest({ query: getBcLogin(), variables });
 
 export const bcLogoutLogin = () =>
-  storefrontGQLRequest({
-    query: logoutLogin(),
-  });
+  isBcStorefrontPlatform
+    ? B3Request.graphqlBC({ query: logoutLogin() })
+    : storefrontGQLRequest({ query: logoutLogin() });
--- a/apps/storefront/src/shared/service/request/b3Fetch.ts
+++ b/apps/storefront/src/shared/service/request/b3Fetch.ts
@@ -46,14 +46,23 @@ function request(path: string, config?: RequestInit, type?: RequestTypeKeys) {
   return b3Fetch(url, init);
 }
 
-function graphqlRequest<T, Y>(type: RequestTypeKeys, data: T, config?: Y) {
-  const init = {
+function graphqlRequest<T, Y>(
+  type: RequestTypeKeys,
+  data: T,
+  config?: Y,
+  extraInit?: Partial<RequestInit>,
+) {
+  const init: RequestInit = {
     method: 'POST',
     headers: {
       'content-type': 'application/json',
       ...config,
     },
     body: JSON.stringify(data),
+    // extraInit is spread last so callers can set non-header RequestInit options (e.g.
+    // credentials, signal). Do not pass `headers` here — it would replace the auth headers
+    // above instead of merging with them.
+    ...extraInit,
   };
 
   const url = GraphqlEndpointsFn(type);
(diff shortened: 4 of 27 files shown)


apps/storefront/src/shared/service/bc/graphql/client.ts
```diff
@@ -6,7 +7,12 @@ export function storefrontGQLRequest<T = any>(data: {
   query: string;
   variables?: object;
 }): Promise<T> {
-  return platform === 'bigcommerce'
+  const { bcGraphqlToken } = store.getState().company.tokens;
```
bc-veer: `client.ts` shouldn't be importing from @/store. It's a service layer file — pulling Redux state in here creates a hidden dependency that callers have no visibility into.

apps/storefront/src/shared/service/request/b3Fetch.ts
```diff
@@ -141,14 +141,24 @@ const B3Request = {
...
+   * causing BC to respond with Access-Control-Allow-Credentials: true.
    */
   graphqlBC: function post<T = any>(data: GQLRequest): Promise<T> {
     const { bcGraphqlToken } = store.getState().company.tokens;
-    const config = {
-      Authorization: `Bearer  ${bcGraphqlToken}`,
-    };
-    return graphqlRequest(RequestType.BCGraphql, data, config);
+    const url = GraphqlEndpointsFn(RequestType.BCGraphql);
+    return b3Fetch(url, {
```
bc-veer: `graphqlBC` was using `graphqlRequest` which handles response parsing and error handling centrally. Now it's dropping straight to b3Fetch and reconstructing headers manually — this bypasses all of that and duplicates fetch logic.

bc-veer (changes requested):

apps/storefront/src/shared/service/request/b3Fetch.ts
```diff
@@ -141,14 +147,18 @@ const B3Request = {
...
+   * Uses `credentials: 'include'` so the browser stores and sends the BC session
+   * cookie set by `bcLogin` — required for cross-origin headless (Catalyst) flows.
+   * The BC Storefront respects this because `storeFrontToken` sets allowedCorsOrigins,
+   * causing BC to respond with Access-Control-Allow-Credentials: true.
    */
   graphqlBC: function post<T = any>(data: GQLRequest): Promise<T> {
     const { bcGraphqlToken } = store.getState().company.tokens;
     const config = {
       Authorization: `Bearer  ${bcGraphqlToken}`,
     };
```
bc-gary, line 163: [question]
`credentials: 'include'` now applies to all `graphqlBC` callers globally. Are there any other direct `graphqlBC` calls outside the three functions in this PR? Just want to confirm there's no caller that shouldn't send credentials.

apps/storefront/src/shared/service/bc/graphql/login.ts
```diff
@@ -76,13 +77,16 @@ interface LoginVariables {
...
+    ? B3Request.graphqlBC({ query: getBcLogin(), variables })
+    : storefrontGQLRequest({ query: getBcLogin(), variables });
 
 export const bcLogoutLogin = () =>
-  storefrontGQLRequest({
-    query: logoutLogin(),
-  });
+  platform === 'bigcommerce' || platform === 'catalyst'
+    ? B3Request.graphqlBC({ query: logoutLogin() })
+    : storefrontGQLRequest({ query: logoutLogin() });
```
bc-gary: [nit] 💅
`platform === 'bigcommerce' || platform === 'catalyst'` appears in 4 places across 3 files. Not blocking, but if this keeps growing a shared `isBcStorefrontPlatform` constant in `basicConfig.ts` would keep it DRY.

apps/storefront/src/shared/service/request/b3Fetch.ts
```diff
@@ -46,14 +46,20 @@ function request(path: string, config?: RequestInit, type?: RequestTypeKeys) {
...
+) {
+  const init: RequestInit = {
     method: 'POST',
     headers: {
       'content-type': 'application/json',
       ...config,
     },
     body: JSON.stringify(data),
+    ...extraInit,
   };
```
bc-gary, line 66: [suggestion]
Since `...extraInit` is spread at the top level, a future caller passing `{ headers: {...} }` would silently replace the auth headers rather than merge them. Might be worth a brief comment that `extraInit` is for non-header options only (e.g. `credentials`, `signal`).

apps/storefront/src/shared/service/request/b3Fetch.ts
shan-pandian: It is global. All direct and indirect caller uses it.

apps/storefront/src/shared/service/request/b3Fetch.ts
shan-pandian: comment added

apps/storefront/src/HeadlessController/getSku.tsx
```diff
@@ -25,7 +25,7 @@ const graphqlRequest: typeof B3Request.graphqlBC | typeof B3Request.graphqlBCPro
   query,
   variables,
 }) => {
-  return platform === 'bigcommerce'
+  return platform === PLATFORM_BIGCOMMERCE
```
bc-veer, line 28: [Suggestion]

Could we move this cleanup into a separate PR? It would keep this PR focused on the issue at hand.

shan-pandian: Cleanup on this case have to be part of the PR. The cleanup is straightforward, not any refactoring. We wanted to define const for isBcStorefrontPlatform and it make sense to address the surrounding. Otherwise, we will move on to next task and we won't cleanup anything.

I would agree to do it separately if it is a complex refactoring or cleanup.

bc-veer:
Fall back to `createCompany` for non-BigCommerce platforms during Buyer Portal registration: [[PR #971](https://github.com/bigcommerce/b2b-buyer-portal/pull/971?utm_source=chatgpt.com)]

PR #960, opened 2026-07-07
feat(common): B2B-4938 update account settings via native BC Storefro…

bc-arezoo:
Jira: [B2B-4938](https://bigcommercecloud.atlassian.net/browse/B2B-4938)

**What/Why?**

Migrates account-settings updates from the legacy name-based b2b middleware mutations to the native BC Storefront GraphQL mutations, behind the PROJECT-7920.use_bc_account_settings flag:

- BC customers → customer.updateCustomer; B2B company users → company.updateCompanyUser.
- Form fields are sent as entityId-keyed typed groups. The fielconfig field_<id> fieldId; choice/checkbox option ids come fromsite.settings.formFields.customer.
- Only changed custom fields are sent (normalized comparison), d now clears it.
- reCaptcha token wiring is required for customer.updateCustomer and is shared via a new useStorefrontCaptcha hook (also adopted by ForgotPassword).
- Flag off → unchanged legacy path.

**Rollout/Rollback**

Gated by the PROJECT-7920.use_bc_account_settings LaunchDarkly e flag; roll back by disabling it (falls straight back to thelegacy middleware path). Code-only change, no data migrations.

**Testing**                                                                                                                                              
- Unit + integration tests added/updated (utils.test.ts, index.test.tsx) — 41 utils + 5 SF-GQL integration tests covering both native paths, change  detection, clearing, and entityId resolution. Full suite green;
- Manual (flag on): BC customer and B2B company user — edited scalar fields, custom text/number fields, and password; confirmed the updateCustomer / updateCompanyUser payloads carry the expected formFields keyed
- Manual (flag off): confirmed the legacy updateBCAccountSettings / updateB2BAccountSettings path is unchanged.
                                            
**Known limitation**: choice/checkbox custom fields require site.settings.formFields.customer, which currently returns empty through the b2b proxy; such edits surface an error rather than saving silently.

Before

https://github.com/user-attachments/assets/99d9ce0c-01f5-4012-bfaf-152e9c48c350

After
includes, b2b, b2c, junior buyer, senior buyer, super admin, masquerading tests

https://github.com/user-attachments/assets/3a7a2d88-397b-4b66-a953-aed09e292262

[B2B-4938]: https://bigcommercecloud.atlassian.net/browse/B2B-4938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Changes how customer profile, passwords, and custom fields are saved (including reCaptcha-gated mutations) across BC and B2B users; rollout is flag-gated but the native path is security- and data-sensitive.
> 
> **Overview**
> When **`PROJECT-7920.use_bc_account_settings`** is on, account-settings saves no longer go through the legacy B2B middleware **`updateBCAccountSettings` / `updateB2BAccountSettings`** path. BC customers use **`customer.updateCustomer`** (with optional reCaptcha v2) and **`customer.changePassword`** for password changes; B2B company users use **`company.updateCompanyUser`**.
> 
> Custom fields are built as **entityId-keyed typed `formFields`** (and name-keyed **`extraFields`** for contact-group B2B fields), using **`field_<id>`** parsing and **`site.settings.formFields.customer`** definitions for choice/checkbox option ids. Only **changed** custom fields are sent (with required fields resent for BC revalidation); values that cannot be represented (cleared numbers/dates, unmapped choices) surface **`fieldCannotBeSaved`** instead of silently dropping edits.
> 
> Save logic moves into **`useAccountSettingsSubmit`**; reCaptcha config loading is centralized in **`useStorefrontCaptcha`** (also used by **Forgot Password**). Init/submit helpers adjust date defaults and scalar payload shape for the native path vs legacy flag-off behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6414b0f34e9a6a4d05ae3ee9a8c5597c06fd1bfb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/AccountSetting/index.tsx
+++ b/apps/storefront/src/pages/AccountSetting/index.tsx
@@ -354,6 +354,24 @@ function AccountSetting() {
     formFields?: CustomerFormFieldsInput,
   ): Promise<boolean> => {
     if (useBcAccountSettings && isBCUser) {
+      const customerInput = buildUpdateCustomerInput(payload, formFields);
+      const hasProfileUpdate = Object.keys(customerInput).length > 0;
+
+      // Check the reCaptcha gates up front (they guard updateCustomer) so a gate can't fail
+      // AFTER the password has already been committed below.
+      if (hasProfileUpdate) {
+        if (isCaptchaConfigLoading) {
+          snackbar.error(b3Lang('global.error.genericMessage'));
+          return false;
+        }
+        if (isCaptchaEnabled && !captchaToken) {
+          snackbar.error(b3Lang('login.loginText.missingCaptcha'));
+          return false;
+        }
+      }
+
+      // Password change (its own mutation, no reCaptcha) runs first: its common failure — a
+      // wrong current password — then fails fast before any profile data is written.
       if (payload.newPassword) {
         const ok = await runMutation(() =>
           changeCustomerPassword(
@@ -366,21 +384,12 @@ function AccountSetting() {
         );
         if (!ok) return false;
       }
-      const customerInput = buildUpdateCustomerInput(payload, formFields);
-      if (Object.keys(customerInput).length === 0) return true;
 
-      // Don't submit until the reCaptcha config has loaded, otherwise isCaptchaEnabled is
-      // still false and we'd send customer.updateCustomer without a token the storefront needs.
-      if (isCaptchaConfigLoading) {
-        snackbar.error(b3Lang('global.error.genericMessage'));
-        return false;
-      }
-      if (isCaptchaEnabled && !captchaToken) {
-        snackbar.error(b3Lang('login.loginText.missingCaptcha'));
-        return false;
-      }
+      if (!hasProfileUpdate) return true;
+
+      let ok;
       try {
-        return await runMutation(() =>
+        ok = await runMutation(() =>
           updateCustomerDetails(customerInput, captchaToken || undefined).then((res) => ({
             errors: res.errors,
           })),
@@ -389,6 +398,13 @@ function AccountSetting() {
         // reCaptcha v2 tokens are single-use; drop it so a retry forces a fresh solve.
         setCaptchaToken('');
       }
+      // The password (if changed) already committed; if the profile update then failed, force a
+      // re-login so the session stays consistent with the changed password (and the user isn't
+      // stuck re-entering a now-stale current password on retry).
+      if (!ok && payload.newPassword) {
+        navigate('/login?loginFlag=loggedOutLogin');
+      }
+      return ok;
     }
 
     if (useBcAccountSettings && !isBCUser) {
--- a/apps/storefront/src/pages/AccountSetting/index.tsx
+++ b/apps/storefront/src/pages/AccountSetting/index.tsx
@@ -25,13 +25,17 @@ import {
   updateBCAccountSettings,
 } from '@/shared/service/b2b';
 import {
+  changeCustomerPassword,
   getCompanyUserDetails,
   getCustomerDetails,
   getCustomerFormFieldDefinitions,
   updateCompanyUserDetails,
   updateCustomerDetails,
 } from '@/shared/service/bc';
-import { CustomerFormFieldDefinition } from '@/shared/service/bc/graphql/accountSetting';
+import {
+  CustomerFormFieldDefinition,
+  CustomerFormFieldsInput,
+} from '@/shared/service/bc/graphql/accountSetting';
 import { isB2BUserSelector, useAppSelector } from '@/store';
 import { CustomerRole, UserTypes } from '@/types';
 import { Fields, ParamProps } from '@/types/accountSetting';
@@ -46,12 +50,12 @@ import { UpgradeBanner } from './UpgradeBanner';
 import {
   b2bSubmitDataProcessing,
   bcSubmitDataProcessing,
+  buildFormFieldsInput,
   buildUpdateCompanyUserInput,
   buildUpdateCustomerInput,
+  collectChangedExtraFields,
   collectChangedFormFields,
-  fieldTypeNeedsOptions,
-  findPasswordFieldEntityId,
-  getUnsendableFormFields,
+  CONTACT_GROUP_ID,
   initB2BInfo,
   initBcInfo,
   mapUserToAccountInfo,
@@ -219,12 +223,11 @@ function AccountSetting() {
 
         let accountSettings;
         if (useBcAccountSettings) {
-          // Scalar/text/number fields get their entityId from the `field_<id>` fieldId, so
-          // definitions are only needed to resolve option ids for choice/checkbox fields.
-          // Both customer.updateCustomer and company.updateCompanyUser use them, so fetch for
-          // either user type in parallel with the account read.
+          // Definitions supply choice option ids AND the entityId fallback (by label) for any
+          // custom field whose fieldId isn't `field_<id>`, so fetch them whenever the form has
+          // any custom field — not only choice fields.
           const needsDefinitions = [...contactInformation, ...additionalInformation].some(
-            (item) => item.custom && fieldTypeNeedsOptions(item.fieldType),
+            (item) => item.custom,
           );
           const definitionsPromise = needsDefinitions
             ? getCustomerFormFieldDefinitions()
@@ -321,10 +324,51 @@ function AccountSetting() {
     }));
   };
 
+  const runMutation = async (
+    run: () => Promise<{
+      errors?: Array<{ message: string }>;
+      resultErrors?: Array<{ message?: string }> | null;
+    }>,
+  ): Promise<boolean> => {
+    let result;
+    try {
+      result = await run();
+    } catch (error) {
+      b2bLogger.error(error);
+      snackbar.error(b3Lang('global.error.genericMessage'));
+      return false;
+    }
+    if (result.errors?.length || result.resultErrors?.length) {
+      const message = result.errors?.[0]?.message || result.resultErrors?.[0]?.message;
+      snackbar.error(message || b3Lang('global.error.genericMessage'));
+      return false;
+    }
+    return true;
+  };
+
   // Sends the prepared payload to the right backend for the current user/flag combination.
-  // Returns true on success; false means an error was already surfaced and the caller stops.
-  const dispatchUpdate = async (payload: Partial<ParamProps>): Promise<boolean> => {
+  // `formFields` is the pre-resolved entityId-keyed group (built once by the caller). Returns
+  // true on success; false means an error was already surfaced and the caller stops.
+  const dispatchUpdate = async (
+    payload: Partial<ParamProps>,
+    formFields?: CustomerFormFieldsInput,
+  ): Promise<boolean> => {
     if (useBcAccountSettings && isBCUser) {
+      if (payload.newPassword) {
+        const ok = await runMutation(() =>
+          changeCustomerPassword(
+            (payload.currentPassword as string) || '',
+            payload.newPassword as string,
+          ).then((res) => ({
+            errors: res.errors,
+            resultErrors: res.data?.customer?.changePassword?.errors,
+          })),
+        );
+        if (!ok) return false;
+      }
+      const customerInput = buildUpdateCustomerInput(payload, formFields);
+      if (Object.keys(customerInput).length === 0) return true;
+
       // Don't submit until the reCaptcha config has loaded, otherwise isCaptchaEnabled is
       // still false and we'd send customer.updateCustomer without a token the storefront needs.
       if (isCaptchaConfigLoading) {
@@ -335,45 +379,25 @@ function AccountSetting() {
         snackbar.error(b3Lang('login.loginText.missingCaptcha'));
         return false;
       }
-      let response;
       try {
-        response = await updateCustomerDetails(
-          buildUpdateCustomerInput(payload, customerFormFieldDefs),
-          captchaToken || undefined,
+        return await runMutation(() =>
+          updateCustomerDetails(customerInput, captchaToken || undefined).then((res) => ({
+            errors: res.errors,
+          })),
         );
-      } catch (error) {
-        b2bLogger.error(error);
-        snackbar.error(b3Lang('global.error.genericMessage'));
-        return false;
       } finally {
         // reCaptcha v2 tokens are single-use; drop it so a retry forces a fresh solve.
         setCaptchaToken('');
       }
-      if (response.errors?.length) {
-        snackbar.error(response.errors[0]?.message || b3Lang('global.error.genericMessage'));
-        return false;
-      }
-      return true;
     }
 
     if (useBcAccountSettings && !isBCUser) {
-      let response;
-      try {
-        response = await updateCompanyUserDetails(
-          buildUpdateCompanyUserInput(payload, customerFormFieldDefs),
-        );
-      } catch (error) {
-        b2bLogger.error(error);
-        snackbar.error(b3Lang('global.error.genericMessage'));
-        return false;
-      }
-      const companyUserErrors = response.data?.company?.updateCompanyUser?.errors;
-      if (response.errors?.length || companyUserErrors?.length) {
-        const message = response.errors?.[0]?.message || companyUserErrors?.[0]?.message;
-        snackbar.error(message || b3Lang('global.error.genericMessage'));
-        return false;
-      }
-      return true;
+      return runMutation(() =>
+        updateCompanyUserDetails(buildUpdateCompanyUserInput(payload, formFields)).then((res) => ({
+          errors: res.errors,
+          resultErrors: res.data?.company?.updateCompanyUser?.errors,
+        })),
+      );
     }
 
     // Legacy b2b middleware path (flag off).
@@ -419,24 +443,44 @@ function AccountSetting() {
           const dataProcessingFn = isBCUser ? bcSubmitDataProcessing : b2bSubmitDataProcessing;
           let payload = dataProcessingFn(data, accountSettings, decryptionFields, extraFields);
 
-          // The native SF GQL updates carry custom form fields collected directly from the
-          // form (the submit processor misses them). Only *changed* fields are sent, and a
-          // form-field-only edit is treated as non-pristine so it isn't dropped as "no edits".
+          // Native SF GQL form fields are resolved to their entityId-keyed groups once here;
+          // the same build reports any changed field that can't be sent so we fail loudly.
+          let customerFormFields;
           if (useBcAccountSettings) {
+            // BC keeps every custom field as an entityId-keyed form field; for B2B the
+            // company-user's own custom fields (contact group) are name-keyed extraFields,
+            // while any additional BC form fields stay entityId-keyed.
+            const formFieldConfig = isBCUser
+              ? accountInfoFormFields
+              : accountInfoFormFields.filter((item) => item.groupId !== CONTACT_GROUP_ID);
             const changedFormFields = collectChangedFormFields(
               data,
-              accountInfoFormFields,
+              formFieldConfig,
               accountSettings?.formFields || [],
+              customerFormFieldDefs,
             );
-            // Fail loudly if any changed field can't actually be sent (unmapped choice option,
-            // cleared number/checkbox, missing entityId) instead of reporting a save that
-            // silently dropped the edit.
-            if (getUnsendableFormFields(changedFormFields, customerFormFieldDefs).length > 0) {
+            const changedExtraFields = isBCUser
+              ? []
+              : collectChangedExtraFields(
+                  data,
+                  accountInfoFormFields,
+                  accountSettings?.extraFields || [],
+                );
+
+            const { formFields, unsendable } = buildFormFieldsInput(
+              changedFormFields,
+              customerFormFieldDefs,
+            );
+            // Fail loudly if a changed form field can't be sent (unmapped choice option, cleared
+            // number/checkbox, missing entityId) rather than reporting a save that dropped it.
+            if (unsendable.length > 0) {
               snackbar.error(b3Lang('global.error.genericMessage'));
               return;
             }
-            if (!payload && changedFormFields.length > 0) payload = {};
-            if (payload) payload.formFields = changedFormFields;
+            customerFormFields = formFields;
+            // Treat a field-only edit as non-pristine so it isn't dropped as "no edits".
+            if (!payload && (formFields || changedExtraFields.length > 0)) payload = {};
+            if (payload && changedExtraFields.length > 0) payload.extraFields = changedExtraFields;
           }
 
           if (payload) {
@@ -457,20 +501,7 @@ function AccountSetting() {
             return;
           }
 
-          // customer.updateCustomer sends the password as a form field; if we can't resolve
-          // its entityId we can't change it — bail before the logout-on-password-change below
-          // so we don't sign the user out on an update that never applied the new password.
-          if (
-            useBcAccountSettings &&
-            isBCUser &&
-            payload.newPassword &&
-            findPasswordFieldEntityId(customerFormFieldDefs) === undefined
-          ) {
-            snackbar.error(b3Lang('global.error.genericMessage'));
-            return;
-          }
-
-          const succeeded = await dispatchUpdate(payload);
+          const succeeded = await dispatchUpdate(payload, customerFormFields);
           if (!succeeded) return;
 
           if (
--- a/apps/storefront/src/pages/AccountSetting/index.tsx
+++ b/apps/storefront/src/pages/AccountSetting/index.tsx
@@ -6,11 +6,13 @@ import trim from 'lodash-es/trim';
 
 import { B3CustomForm } from '@/components/B3CustomForm';
 import CustomButton from '@/components/button/CustomButton';
+import { Captcha } from '@/components/captcha/Captcha';
 import { b3HexToRgb, getContrastColor } from '@/components/outSideComponents/utils/b3CustomStyles';
 import B3Spin from '@/components/spin/B3Spin';
 import { useFeatureFlag } from '@/hooks/useFeatureFlag';
 import { useMobile } from '@/hooks/useMobile';
 import useStorageState from '@/hooks/useStorageState';
+import { useStorefrontCaptcha } from '@/hooks/useStorefrontCaptcha';
 import { useB3Lang } from '@/lib/lang';
 import { CustomStyleContext } from '@/shared/customStyleButton';
 import {
@@ -22,10 +24,18 @@ import {
   updateB2BAccountSettings,
   updateBCAccountSettings,
 } from '@/shared/service/b2b';
-import { getCompanyUserDetails, getCustomerDetails } from '@/shared/service/bc';
+import {
+  getCompanyUserDetails,
+  getCustomerDetails,
+  getCustomerFormFieldDefinitions,
+  updateCompanyUserDetails,
+  updateCustomerDetails,
+} from '@/shared/service/bc';
+import { CustomerFormFieldDefinition } from '@/shared/service/bc/graphql/accountSetting';
 import { isB2BUserSelector, useAppSelector } from '@/store';
 import { CustomerRole, UserTypes } from '@/types';
 import { Fields, ParamProps } from '@/types/accountSetting';
+import b2bLogger from '@/utils/b3Logger';
 import { B3SStorage } from '@/utils/b3Storage';
 import { snackbar } from '@/utils/b3Tip';
 import { channelId, isCatalystPlatform } from '@/utils/basicConfig';
@@ -36,6 +46,10 @@ import { UpgradeBanner } from './UpgradeBanner';
 import {
   b2bSubmitDataProcessing,
   bcSubmitDataProcessing,
+  buildUpdateCompanyUserInput,
+  buildUpdateCustomerInput,
+  collectChangedFormFields,
+  fieldTypeNeedsOptions,
   initB2BInfo,
   initBcInfo,
   mapUserToAccountInfo,
@@ -136,6 +150,11 @@ function AccountSetting() {
   const dedupeStorefrontConfigFetchCalls = useFeatureFlag(
     'B2B-5309.dedupe_storefront_config_fetch_calls',
   );
+
+  // BC customers update via customer.updateCustomer, which needs a reCaptcha token
+  // when reCaptcha is enabled on the storefront.
+  const isCustomerUpdate = useBcAccountSettings && isBCUser;
+
   const [isMobile] = useMobile();
 
   const navigate = useNavigate();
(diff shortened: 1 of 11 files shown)


apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -267,6 +272,201 @@ export const bcSubmitDataProcessing = (
...
+        const values = Array.isArray(value) ? value : [value];
+        const fieldValueEntityIds = values
+          .map((choice) => optionEntityId(fieldEntityId, choice))
+          .filter((id): id is number => id !== undefined);
+        if (fieldValueEntityIds.length === 0) {
+          unsendable.push(formField);
+          return;
+        }
+        seen.add(fieldEntityId);
+        (formFields.checkboxes ??= []).push({ fieldEntityId, fieldValueEntityIds });
```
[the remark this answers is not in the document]
bc-arezoo: Fixed

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -242,6 +301,50 @@ function AccountSetting() {
...
+          captchaToken || undefined,
+        );
+      } finally {
+        // reCaptcha v2 tokens are single-use; drop it so a retry forces a fresh solve.
+        setCaptchaToken('');
+      }
+      if (response.errors?.length) {
+        snackbar.error(response.errors[0]?.message || b3Lang('global.error.genericMessage'));
+        return false;
+      }
```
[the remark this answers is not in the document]
bc-arezoo: Fixed

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -242,6 +301,50 @@ function AccountSetting() {
     }));
   };
 
+  // Sends the prepared payload to the right backend for the current user/flag combination.
+  // Returns true on success; false means an error was already surfaced and the caller stops.
+  const dispatchUpdate = async (payload: Partial<ParamProps>): Promise<boolean> => {
+    if (useBcAccountSettings && isBCUser) {
+      if (isCaptchaEnabled && !captchaToken) {
+        snackbar.error(b3Lang('login.loginText.missingCaptcha'));
+        return false;
```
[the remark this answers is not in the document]
bc-arezoo: Fixed

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -22,10 +24,18 @@ import {
   updateB2BAccountSettings,
   updateBCAccountSettings,
 } from '@/shared/service/b2b';
-import { getCompanyUserDetails, getCustomerDetails } from '@/shared/service/bc';
+import {
+  getCompanyUserDetails,
+  getCustomerDetails,
+  getCustomerFormFieldDefinitions,
```
bc-vasu, line 27: [Question]`getCustomerFormFieldDefinitions` is this basically replacement of Account form fields API ?

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
```diff
@@ -173,3 +173,175 @@ export async function getCustomerDetails(): Promise<CustomerResponse> {
...
+// fields the customer has no value for yet (which customer.formFields omits). This
+// is the authoritative source of fieldEntityId for the customer.updateCustomer input.
+// Choice fields also expose their options, so a selected label maps to its own entityId.
+export interface CustomerFormFieldOption {
+  entityId: number;
+  label: string;
+}
+
+export interface CustomerFormFieldDefinition {
+  __typename: string;
```
bc-veer: [Question]
is `__typename` is required ?

apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -267,6 +272,200 @@ export const bcSubmitDataProcessing = (
...
+    .map((item) => ({
+      name: item.bcLabel || '',
+      value: data[item.name || ''],
+      fieldType: item.fieldType,
+      fieldEntityId: parseFieldEntityId(item.fieldId),
+    }))
+    .filter((formField) => {
+      const original = originalFormFields.find((item) => item.name === formField.name);
+      return normalizeForCompare(original?.value) !== normalizeForCompare(formField.value);
+    });
```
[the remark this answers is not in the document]
bc-arezoo: I need confirmation for extra fields data structure

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -162,17 +190,49 @@ function AccountSetting() {
         const additionalInformation = (accountFormFields?.additionalInformation ??
           []) as Partial<Fields>[];
 
+        // Store the customer form-field definitions (option entityIds for choice fields),
+        // surfacing — not swallowing — a failed fetch since it's what blocks choice fields.
+        const applyFormFieldDefinitions = (
```
bc-vasu, line 196: [Question]
Are we guarding this against Feature Flag ?
[Required] Any new changes should be guard against Feature Flag.

apps/storefront/src/types/accountSetting.ts
```diff
@@ -30,6 +30,8 @@ export interface Fields {
 interface BcFormFieldsProps {
   name: string;
   value: any;
+  fieldType?: string;
+  fieldEntityId?: number;
```
bc-arezoo, line 34: yes, the flow is: collectChangedFormFields parses it from the config fieldId → carries it on the BcFormFieldsProps entry → buildFormFieldsInput consumes it to build customer.updateCustomer / company.updateCompanyUser input.

apps/storefront/src/pages/AccountSetting/index.tsx
bc-arezoo: No — they serve different roles; getCustomerFormFieldDefinitions is a narrow supplement to the account-form-fields API, not a replacement.

getB2BAccountFormFields (B2BAccountFormFields, b2b middleware) — the account form config. Returns every field with its fieldId, fieldName, fieldType, groupId, labelName, custom, visible, valueConfigs, etc. This is what actually drives the page: rendering the form (contact/additional groups), field metadata, defaults, and — for the update — the field type routing and the field_<id> → entityId parse. Still required and still the primary source.

getCustomerFormFieldDefinitions (CustomerFormFieldSettings → site.settings.formFields.customer, BC Storefront) — used for exactly one thing the config can't provide: resolving choice/checkbox option entityIds (options { entityId label }) and the Password field entityId needed to build the native customer.updateCustomer / company.updateCompanyUser input. It's only fetched when the form has a choice/checkbox custom field.

apps/storefront/src/pages/AccountSetting/index.tsx
bc-arezoo: the new behavior is guarded by the flag.

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -370,6 +512,12 @@ function AccountSetting() {
             setValue={setValue}
           />
 
+          {isCustomerUpdate && isCaptchaEnabled && (
```
bc-vasu, line 375: [Question]
Is FF needed ?

bc-arezoo: isCustomerUpdate has the check for feature flag `useBcAccountSettings`

apps/storefront/src/pages/AccountSetting/index.tsx
bc-vasu: Does the old Account form fields API was not supporting this behaviour getCustomerFormFieldDefinitions
Was it broken in previous implementation as well (I mean when FF is turned OFF)?

@bc-arezoo

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
```diff
@@ -173,3 +173,175 @@ export async function getCustomerDetails(): Promise<CustomerResponse> {
...
+
+const QUERY_CUSTOMER_FORM_FIELD_SETTINGS = `query CustomerFormFieldSettings {
+  site {
+    settings {
+      formFields {
+        customer {
+          __typename
+          entityId
+          label
+          ... on MultipleChoicesFormField { options { entityId label } }
```
bc-lulu-yang: [question]

I've added a custom filed to test, seems getting error in this request. Can you have a look pls?

Image

bc-lulu-yang: it works with new commits

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-arezoo: No, removed

apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -267,6 +276,293 @@ export const bcSubmitDataProcessing = (
...
+      return {
+        name,
+        value: data[item.name || ''],
+        fieldType: item.fieldType,
+        fieldEntityId: parseFieldEntityId(item.fieldId) ?? entityIdByLabel.get(name),
+      };
+    })
+    .filter((formField) => {
+      const original = originalFormFields.find((item) => item.name === formField.name);
+      return isValueChanged(original?.value, formField.value, formField.fieldType);
```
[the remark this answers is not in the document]
bc-arezoo: Addressed

bc-arezoo:
@bc-vasu As discussed yesterday, can you please confirm the `reCaptcha` changes with @bc-amandalan and how we need to move forward?

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
```diff
@@ -173,3 +173,228 @@ export async function getCustomerDetails(): Promise<CustomerResponse> {
...
+      customer {
+        firstName
+        lastName
+        company
+        phoneNumber: phone
+        email
+      }
+    }
+  }
+}`;
```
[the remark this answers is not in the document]
bc-vasu: we can ignore it as guarded against FF

bc-vasu (dismissed):

bc-amandalan:
Hi @bc-vasu, did you test on user masquerading? Is it working properly?

bc-vasu:
> Hi @bc-vasu, did you test on user masquerading? Is it working properly?

I’ve tested it to a reasonable extent, but I’m not fully confident about the masquerading coverage. I validated a few scenarios, though I am not sure if that includes everything. Since we have a blitz planned for it, we can expand the coverage there.
Else if you are aware of full coverage please do have a look
@bc-amandalan

bc-amandalan:
Can you add Screenshots for your manual tests on this Page?

bc-vasu:
> Can you add Screenshots for your manual tests on this Page?

Sure let me re-deploy on integration and will add shortly @bc-amandalan

bc-vasu:
> Can you add Screenshots for your manual tests on this Page?

@bc-amandalan I have added the testing videos in description for reference

bc-amandalan:
Hi @bc-vasu, thanks for adding the manual tests. Can you set up all form field types and extra field types, and each type twice to make sure all working properly. 🙇 As the code change also include these customised functionalities.

bc-vasu:
> PROJECT-7920.use_bc_account_settings

@bc-amandalan are we good with this ?


https://github.com/user-attachments/assets/9ad0b906-58a4-4102-a0b5-b3d6ac83371b

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -259,6 +326,103 @@ function AccountSetting() {
...
+          );
+        } finally {
+          // reCaptcha v2 tokens are single-use; drop it so a retry forces a fresh solve.
+          setCaptchaToken('');
+        }
+        if (!ok) return false;
+      }
+
+      // Details update succeeded (or there was none); now change the password if requested.
+      if (payload.newPassword) {
```
bc-lulu-yang: [Required]

Does this mean, current password is not required when changing password. From understanding, FE at least need to make sure current password is passed into server.  

So the useful FE is something like:
```
if (payload.newPassword && payload.currentPassword) {
  // allow the password change call
}
```

That only prevents an accidental empty submit. A determined attacker can still hit the API directly without currentPassword.

Real protection is: backend rejects password change when currentPassword is missing or wrong (and ideally rate-limits / lockout). Hi @bc-amandalan can you help check how currently backend server handle this case?

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-lulu-yang, line 122: [question/suggestion]

Not sure in which case we need `password` back. And what does PasswordFormFieldValue.password come back as empty, or masked (********), or actual plaintext.

If empty/masked only, then low real risk; if plaintext then may be a security issue.  But querying password is still unnecessary, we can consider drop password from the fragment if the functionality is still no effect after dropping

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -294,10 +458,55 @@ function AccountSetting() {
...
+              changedFormFields,
+              customerFormFieldDefs,
+            );
+            const { extraFields, unsendable: unsendableExtra } =
+              buildExtraFieldsInput(changedExtraFields);
+            // Fail loudly if a changed field can't be sent (unmapped choice option, cleared
+            // number/date, missing entityId, or an array-valued extra field) rather than
+            // reporting a save that silently dropped the edit.
+            if (unsendable.length > 0 || unsendableExtra.length > 0) {
+              snackbar.error(b3Lang('global.error.genericMessage'));
```
bc-lulu-yang: [Suggestion/Nit]

Feel like this message is to generic, user can't tell why save failed. Consider a specific field-level error message.

apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -267,6 +276,314 @@ export const bcSubmitDataProcessing = (
...
+}
+
+export function buildUpdateCustomerInput(
+  payload: Partial<ParamProps>,
+  formFields?: CustomerFormFieldsInput,
+): UpdateCustomerInput {
+  const input: UpdateCustomerInput = {};
+
+  assignSharedScalars(input, payload);
+  if (payload.company !== undefined) input.company = payload.company as string;
```
bc-lulu-yang, line 534: [question]

Trying to understand, why b2c customer has `company`, while updating company here, any business flow need to trigger (for example linked to a b2b company)? Or this is just for display personal info.

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -259,6 +326,103 @@ function AccountSetting() {
     }));
   };
 
+  const runMutation = async (
```
bc-lulu-yang: [Nit]

As this file getting bigger, seems the `runMutation` `dispatchUpdate`  and `handleAddUserClick` all related to save/update. Considering to move together to a new hook file to make this entry a bit clean and easy to read/maintain.

apps/storefront/src/pages/AccountSetting/index.tsx
bc-lulu-yang: Just tested, yes, the backend server handles correctly
Screenshot 2026-07-29 at 10 30 26 am

apps/storefront/src/pages/AccountSetting/utils.ts
bc-vasu: b2c can have company as you see in customers control panel company can be passed but it doe snot link

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
```diff
@@ -173,3 +173,228 @@ export async function getCustomerDetails(): Promise<CustomerResponse> {
...
+  multilineTexts?: Array<{ fieldEntityId: number; multilineText: string }>;
+  texts?: Array<{ fieldEntityId: number; text: string }>;
+}
+
+export interface UpdateCustomerInput {
+  firstName?: string;
+  lastName?: string;
+  company?: string;
+  phone?: string;
+  email?: string;
```
bc-lulu-yang, line 252: [question]

Does this mean after updating the email, the customer need to use the new email to login? I tried with the new email, seems get error. Seems it didn't get updated in bc control panel

cc: @bc-amandalan may I know if I need to wait for a longer time to try? or will bc endpoint will update immediately.

https://github.com/user-attachments/assets/1a122ef0-d51d-4f82-92ab-25875b341194

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -259,6 +326,103 @@ function AccountSetting() {
...
+    formFields?: CustomerFormFieldsInput,
+    extraFields?: CompanyUserExtraFieldsInput,
+  ): Promise<boolean> => {
+    if (useBcAccountSettings && isBCUser) {
+      const customerInput = buildUpdateCustomerInput(payload, formFields);
+      const hasProfileUpdate = Object.keys(customerInput).length > 0;
+
+      // Check the reCaptcha gates up front (they guard updateCustomer) so we don't attempt the
+      // details update — and its password follow-up — with a token that can't be accepted.
+      if (hasProfileUpdate) {
```
bc-jeffteng: [question]
Does this mean we only need reCaptcha for profile update not password change?

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
```diff
@@ -173,3 +173,228 @@ export async function getCustomerDetails(): Promise<CustomerResponse> {
...
+
+// customer.updateCustomer needs a reCaptchaV2 token when reCaptcha is enabled on the
+// storefront, so the token argument is only added when one is supplied.
+const updateCustomerMutation = (
+  withReCaptcha: boolean,
+) => `mutation UpdateCustomer($input: UpdateCustomerInput!${
+  withReCaptcha ? ', $reCaptchaToken: String!' : ''
+}) {
+  customer {
+    updateCustomer(input: $input${withReCaptcha ? ', reCaptchaV2: { token: $reCaptchaToken }' : ''}) {
```
bc-jeffteng, line 276: [question]
Do we need `error` selected in the schema? like:
```
updateCustomer(...) {
  customer { ... }
  errors {
    message
  }
}
```

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -259,6 +326,103 @@ function AccountSetting() {
...
+        let ok;
+        try {
+          ok = await runMutation(() =>
+            updateCustomerDetails(customerInput, captchaToken || undefined).then((res) => ({
+              errors: res.errors,
+            })),
+          );
+        } finally {
+          // reCaptcha v2 tokens are single-use; drop it so a retry forces a fresh solve.
+          setCaptchaToken('');
```
bc-jeffteng: [question]
We clean up the capture token because it is single-use, but is there any place we try to regenerate it for the further request?

apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -267,6 +276,314 @@ export const bcSubmitDataProcessing = (
...
+  return accountInfoFormFields
+    .filter((item) => item.custom)
+    .map((item) => {
+      // The field's display label is the key that matches the stored formFields name and the
+      // definition label. getAccountFormFields leaves bcLabel empty when the config supplies
+      // only labelName (kept on `label`), so fall back to label.
+      const name = item.bcLabel || item.label || '';
+      return {
+        name,
+        value: data[item.name || ''],
```
bc-jeffteng, line 598: The unset `date` field is initialise with `today` in `registerUtils.ts`. It will make this comparison treat it like user edit then saving unrelated field to `today` date. Can we only collect dirty fields, or avoid default empty dates on this?

bc-jeffteng (commented):
Overall looks good, but I found the test doesn't cover the mutation level error case. It would probably be good to have it covered.

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-vasu: @bc-amandalan seems we need inputs on API

good use case by @bc-lulu-yang

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-vasu: Good catch, 
checked and the storefront API returns this masked/empty, never plaintext, so there's no real exposure. 

We don't do anything with the value besides pre-filling the form default, so I'll leave it as-is for now to keep this PR scoped to the mutation migration.

Can be ignored

apps/storefront/src/pages/AccountSetting/index.tsx
bc-vasu: recaptcha is only for b2c not others thats why you see like this

apps/storefront/src/pages/AccountSetting/useAccountSettingsSubmit.ts
```diff
@@ -0,0 +1,329 @@
...
+            updateCustomerDetails(customerInput, captchaToken || undefined).then((res) => ({
+              errors: res.errors,
+            })),
+          );
+        } finally {
+          // reCaptcha v2 tokens are single-use; reset the widget so a retry can issue a new one.
+          resetCaptcha();
+        }
+        if (!ok) return false;
+      }
```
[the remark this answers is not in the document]
bc-vasu: bcSubmitDataProcessing unconditionally copied firstName/lastName/phone/email into the payload whenever any field was dirty, even if that specific scalar hadn't changed. Since the form always submits the full field set, this made buildUpdateCustomerInput non-empty on every save, forcing a spurious updateCustomer call (and its reCaptcha gate) before changePassword — even for a password-only edit. Fixed by only assigning each scalar when it actually differs from the stored value .

 Left b2bSubmitDataProcessing untouched, its test confirmed the full-snapshot behavior is intentional there (legacy middleware + a single atomic updateCompanyUser call, so no reCaptcha-sequencing issue).

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-vasu: added errors { message } to the mutation and response type, matching changePassword/updateCompanyUser, and wired it into resultErrors in dispatchUpdate

apps/storefront/src/pages/AccountSetting/useAccountSettingsSubmit.ts
```diff
@@ -0,0 +1,329 @@
...
+        return runMutation(() =>
+          changeCustomerPassword(
+            (payload.currentPassword as string) || '',
+            payload.newPassword as string,
+          ).then((res) => ({
+            errors: res.errors,
+            resultErrors: res.data?.customer?.changePassword?.errors,
+          })),
+        );
+      }
```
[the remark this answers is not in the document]
bc-vasu: dispatchUpdate now returns { ok, profileCommitted } instead of a single boolean, so when the profile details update succeeds but the trailing changeCustomerPassword call fails, the caller still knows the profile change was committed server-side and performs the corresponding client refresh (forced logout on email change / storage clear), rather than silently no-op'ing and leaving stale local state.

apps/storefront/src/pages/AccountSetting/useAccountSettingsSubmit.ts
```diff
@@ -0,0 +1,346 @@
...
+          if (!ok && !profileCommitted) return;
+
+          if (
+            (ok && data.password && data.currentPassword) ||
+            customer.emailAddress !== trim(data.email)
+          ) {
+            navigate('/login?loginFlag=loggedOutLogin');
+          } else {
+            B3SStorage.clear();
+            setIsFinishUpdate(true);
```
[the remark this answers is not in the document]
bc-vasu: It is due to other API as there are two APIs structured one for update Account other for password
so it can be ignored

bc-jeffteng (approved):

bc-lulu-yang (approved):
LGTM, with a tiny suggestion

apps/storefront/src/pages/AccountSetting/index.tsx
bc-lulu-yang: Just thinking to have a better user experience, we should give a alert to user to let them know they need to have current password when updating password. UI should block before backend check.

bc-vasu: We can have a separate ticket for the same and discuss with product on new UI designs

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-amandalan: If there is an issue, hold on before merge. Unless we are sure the behaviour is the same before and after.

apps/storefront/src/pages/AccountSetting/index.tsx
```diff
@@ -246,91 +294,28 @@ function AccountSetting() {
...
-    const userExtraFields = accountInfoFormFields.filter(
-      (item: CustomFieldItems) => item.custom && item.groupId === 1,
-    );
-    return userExtraFields.map((item: CustomFieldItems) => ({
-      fieldName: deCodeField(item?.name || ''),
-      fieldValue: data[item.name],
-    }));
-  };
-
-  const handleAddUserClick = () => {
```
bc-amandalan, line 262: [Question]
Why do we remove this code? Is this related?

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-vasu: @bc-amandalan Its specific to login behaviour which might be causing issue not update behaviour
That to at API level

bc-amandalan: If API level is broken, why do we want to rollout a broken feature?

apps/storefront/src/pages/AccountSetting/index.tsx
bc-vasu: This code hasn’t been removed — it’s now encapsulated and retrieved through the `useAccountSettingsSubmit` hook to reduce the overall file size.

@bc-amandalan

apps/storefront/src/shared/service/bc/graphql/accountSetting.ts
bc-vasu: We are not rolling out the login feature the FF is already off for login
so its login issue which is broken
And also we are not rolling out this update feature its guarded with Feature flag so FF will remain off
it will be easier for everyone to test at integration or staging when API is fixed for it

Also, the ticket which Areezo worked its frontend so  will it remain open till API is not resolved ?

bc-lulu-yang: This flow works when feature flag off.  And it also partial works when feature flag on now, see the video, the update only works when I updated one of the form field; I will get error if no change on form field value.

Off case:

https://github.com/user-attachments/assets/27bb9413-82b9-4f62-80ad-13a4cef77ae2

On case:

https://github.com/user-attachments/assets/562e9e4d-72c7-47d9-a9f7-08011d72566e

Only tested b2c user at the moment.

apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -267,6 +313,325 @@ export const bcSubmitDataProcessing = (
...
+        fieldEntityId: parseFieldEntityId(item.fieldId) ?? entityIdByLabel.get(name),
+        // The react-hook-form registered name, carried through so an unsendable field can be
+        // traced back to the control it belongs to (for a field-level error).
+        formName: item.name,
+      };
+    })
+    .filter((formField) => {
+      const original = originalFormFields.find((item) => item.name === formField.name);
+      return isValueChanged(original?.value, formField.value, formField.fieldType);
+    });
```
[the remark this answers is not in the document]
bc-amandalan: addressed

bc-amandalan:
When FF off: main and this branch behaviour the same
B2B user fetch and update all successful:


https://github.com/user-attachments/assets/e5745196-d952-4020-84ef-690ff0192635

bc-amandalan:
When FF off: main and this branch behaviour the same
B2C user fetch and update all successful:

https://github.com/user-attachments/assets/d58da56c-d222-4c9d-b5a4-53a8fbd3314a

bc-amandalan:
Super admin works as expected when FF off:

https://github.com/user-attachments/assets/0752001b-ca83-49dc-9ac5-a4dd0e4727a5

bc-lulu-yang:
buyer portal on Catalyst SF:

FF off case still using legacy flow: 

**B2B company admin** ✅

- account setting page load: B2BAccountFormFields + GetB2bAccountSettings
- update extra field, name, phone number: updateAccountSettings

https://github.com/user-attachments/assets/6d63b00d-e566-4593-ab26-e6ed91b3f58c

Update password: able to sign in with new password ✅

(Please note: in local, I can see when reset password, sign out behavior looks wired, jumping between pages)

https://github.com/user-attachments/assets/a7d4a123-7b97-40a7-91e5-cd50d000540f

https://github.com/user-attachments/assets/11c1294d-5f81-47d4-9d4d-98184fd09d5f

Update email: able to sign in with new email ✅

https://github.com/user-attachments/assets/07bad62e-1009-44df-b4a0-1a764ccb31c9

https://github.com/user-attachments/assets/6e3a945c-16b0-4bd2-bdfa-7a41af3a7ae4

Only update user works (first name TestLulu -> TestL)

https://github.com/user-attachments/assets/9fbd5b2b-184f-4464-bb24-dd7481416499


**B2C customer** ❌

Update email together with name and phone number ✅

https://github.com/user-attachments/assets/94ff0195-5565-422d-a90a-845d077d2350

https://github.com/user-attachments/assets/154df7c1-1f68-4600-b4a9-0b094c94f09d

Update password flow works ✅

https://github.com/user-attachments/assets/0eb2fd14-4703-4a05-a3b1-a6d3b5250b45

Only updating user name failed ❌

https://github.com/user-attachments/assets/e82a7e6e-f5e3-4d3e-b8e3-5fb0657ac87f


Have tested this with main buyer-portal branch, it works. so seems something may need to have a look here

https://github.com/user-attachments/assets/b1d6bc51-df00-40a2-8166-6f5d21ad4bc1

apps/storefront/src/pages/AccountSetting/utils.ts
```diff
@@ -200,6 +258,7 @@ export const bcSubmitDataProcessing = (
   accountSettings: any,
   decryptionFields: Partial<Fields>[],
   extraFields: Partial<Fields>[],
+  useBcAccountSettings: boolean = true,
```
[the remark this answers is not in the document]
bc-lulu-yang: @bc-amandalan this is the reason of the issue I found today: **b2c customer can't update the user name only when FF off**, since the `bcSubmitDataProcessing` is not under `useBcAccountSettings` false flow

bc-lulu-yang:
As confirmed with @snaderiBC , we are going to pause the unified account setting adoption. So close this PR until we need it

PR #955, opened 2026-07-05
feat(orders): B2B-5157 Fix currency formatting on unified orders list for multi-currency stores

bc-lulu-yang:
Jira: [B2B-5157](https://bigcommercecloud.atlassian.net/browse/B2B-5157)

## What/Why?
The current unified SF GQL order list displays incorrect grand-total, for example, a USD order on a store with AUD as the default currency renders as A$90 instead of $90.00. This PR is to fix the display price issue in unified SF GQL order list when an order’s transaction currency differs from the shopper’s active session currency. 

This fix applies for both desktop and mobile version. But the issue still there for legacy mobile at the moment, will not include fix for legacy in this ticket

## Rollout/Rollback
Behind FF `buyer_portal_unified_sf_gql_orders`, and we can revert this PR if needed

## Testing
With mocked data in local

Desktop:
Screenshot 2026-07-06 at 12 47 02 pm

Mobile:
Screenshot 2026-07-06 at 12 47 15 pm

[B2B-5157]: https://bigcommercecloud.atlassian.net/browse/B2B-5157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes how order totals are displayed on a customer-facing path; behavior is gated by the unified SF GQL orders flag and falls back to existing formatters, but misformatted API values would show incorrect amounts.
> 
> **Overview**
> Fixes **incorrect grand totals** on unified My Orders and Company Orders when an order’s transaction currency differs from the shopper’s active store currency (e.g. USD order shown with AUD symbols).
> 
> The SF GQL **`Money`** projection and order queries now request **`formattedV2`**. **`mapSfGqlOrderToListItem`** exposes it as **`formattedTotalIncTax`**, and the desktop table (**`Order`**) and mobile card (**`OrderItemCard`**) render that string first instead of **`currencyFormat`** / legacy **`money`** parsing. Legacy formatting remains when the pre-formatted value is missing.
> 
> Tests add **`buildSfGqlMoneyWith`**, preload store currency where needed, and assert the UI shows the API’s formatted string rather than client-side formatting.
> 
[...]

PR #961, opened 2026-07-09
[B2B-5226] Use formattedV2 from SF GQL Money for order detail price display

bc-gary:
Jira: [B2B-5226](https://bigcommercecloud.atlassian.net/browse/B2B-5226)

## What/Why?

Per Sylvia's directive ([Slack thread](https://bigcommerce.slack.com/archives/C0AM7FDJ6CE/p1782887279565399)):
*"B2B team should not come up with a separate way of rendering the formatting. If `formattedV2` is exposed, we should use it as-is."*

The order detail page previously used `buildMoneyFormat()` + `ordersCurrencyFormat()` with the Redux
`currencies` store config for all price formatting. This caused incorrect currency display on multi-currency
stores (e.g., AUD order on USD-default store rendered with wrong currency symbol).

This PR replaces manual FE formatting with `formattedV2` from the SF GQL `Money` type wherever a server-provided
Money object is available, and uses `Intl.NumberFormat` with the order's `currencyCode` for derived values
where no Money object exists.

### Changes

**Schema & types:**
- Added `subTotalSalePrice: Money` to `OrderLineItem` and `OrderDigitalLineItem` interfaces and query fragment
- Added `formattedPrice` and `formattedTotal` optional fields to `OrderProductItem` and `ProductItem`

**Converter (`convertOrderDetail.ts`):**
- Removed `buildMoneyFormat()`, `formatPrice()`, and `currencies` parameter
- Order summary uses `formatMoney()` → `money.formattedV2` for all 7 summary fields (sub total, shipping,
  handling, discount, coupons, tax, grand total)
- Product TOTAL uses `subTotalSalePrice.formattedV2` (server-authoritative)
- Product PRICE (unit) uses `subTotalListPrice.value / quantity` formatted with `Intl.NumberFormat`
  (no unit price Money field exists in SF GQL — intentional per Orders team)
- Shipping cost uses `formattedV2`

**Components:**
- `OrderAction`: dual-path display — legacy path uses `ordersCurrencyFormat`, unified path displays
  `formattedV2` values directly
- `B3ProductList`: added `currencyCode` prop, prefers `formattedPrice`/`formattedTotal` when available
  and no discounts modify the price
- `OrderCheckboxProduct`: uses `formattedPrice` for unit price display, `Intl.NumberFormat` for
  recalculated totals in re-order modal
- `OrderShipping`/`OrderBilling`: pass `currencyCode` to `B3ProductList`
- `OrderDialog`: threads `currencyCode` to `OrderCheckboxProduct`

**Context:**
- Removed `defaultMoneyFormat` from `OrderDetailsContext` init state
- `money: MoneyFormat` kept on context interface for legacy path backward compatibility

### Unit price decision

SF GQL `OrderPhysicalLineItem` intentionally has no unit price Money field — only subtotals
(`subTotalListPrice`, `subTotalSalePrice`). Confirmed by Shayne Even (Orders team) in
[#orders-dev](https://bigcommerce.slack.com/archives/C96P5T3J4/p1783492815069929):
promotions can create values that don't divide cleanly.

Jesse Bayley confirmed ([program thread](https://bigcommerce.slack.com/archives/C0AM7FDJ6CE/p1783555287448429)):
*"For the purposes of our b2b unification project, we can go with `subTotalListPrice / quantity`."*

**Depends on:** PR #955 (B2B-5157 — adds `formattedV2` to `Money` interface and `moneyFields` fragment)

## Rollout/Rollback

- Behind feature flag `B2B-4613.buyer_portal_unified_sf_gql_orders` — no production impact when flag is OFF
- Legacy (non-unified) path unchanged — `money: MoneyFormat` + `ordersCurrencyFormat` still works
- Rollback: revert this PR; legacy path continues to function

## Testing

- All 125 order detail tests pass (39 unified + 86 legacy)
- Unified summary tests verify `formattedV2` values render correctly for EUR currency
- Currency re-rendering test verifies `formattedV2` is stable regardless of Redux currency config changes
- Legacy path tests unchanged and passing — backward compatibility verified
- Live schema introspection confirmed `formattedV2` exists on all Money fields

[B2B-5226]: https://bigcommercecloud.atlassian.net/browse/B2B-5226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches all order-detail money display paths for the unified SF GQL flag; wrong fallbacks could show incorrect symbols or totals, though legacy orders and the feature flag boundary limit blast radius.
> 
> **Overview**
> Unified order detail pricing no longer builds amounts from Redux store currency config. **`convertOrderDetail`** drops `currencies` / `buildMoneyFormat` and fills summary, shipping, and line totals from **`Money.formattedV2`** (with **`Intl.NumberFormat`** only when `formattedV2` is missing). Line items now query **`subTotalSalePrice`** and expose **`formattedPrice`** / **`formattedTotal`** on products; unit price is derived from **`subTotalListPrice / quantity`** in the order’s currency.
> 
> **`B3ProductList`** accepts **`currencyCode`**, prefers pre-formatted strings when there are no discounts and quantity matches the original line (so partial shipments recalculate). **`OrderAction`** shows summary strings as-is when **`money`** is unset (unified path) and still uses **`ordersCurrencyFormat`** for legacy. **`currencyCode`** is threaded through shipping, billing, reorder/shopping-list dialogs, and context **`money`** is cleared on unified load so stale legacy format does not stick.
> 
> Tests drop Redux currency hydration / re-format-on-currency-change expectations and assert EUR **`formattedV2`** strings directly; fixtures add **`subTotalSalePrice`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2067279046f23fdf50cac5e400c462f2dca75bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/components/B3ProductList.tsx
+++ b/apps/storefront/src/components/B3ProductList.tsx
@@ -149,6 +149,8 @@ function getBackorderLayoutStyles(
 interface ProductProps<T extends ProductItem = ProductItem> {
   products: Array<T & ProductItem>;
   money?: MoneyFormat;
+  /** Order currency code (e.g. "USD"). When provided without `money`, uses Intl.NumberFormat. */
+  currencyCode?: string;
   renderAction?: (item: T & ProductItem) => ReactElement;
   actionWidth?: string;
   quantityKey?: string;
@@ -188,6 +190,7 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
     canToProduct = false,
     textAlign = 'left',
     money,
+    currencyCode,
     type,
     getCurrentProductUrls,
     catalogBackorderUiEnabled = false,
@@ -387,18 +390,33 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
           formatOnlyAvailable,
         });
 
-        const getDisplayPrice = (priceValue: number) => {
-          const newMoney = money
-            ? ordersCurrencyFormat(money, priceValue)
-            : currencyFormat(priceValue);
+        const getDisplayPrice = (priceValue: number, preFormatted?: string) => {
+          if (preFormatted) return showTypePrice(preFormatted, product);
 
-          return showTypePrice(newMoney, product);
+          let formatted: string;
+          if (money) {
+            formatted = ordersCurrencyFormat(money, priceValue);
+          } else if (currencyCode) {
+            formatted = new Intl.NumberFormat('en', {
+              style: 'currency',
+              currency: currencyCode,
+            }).format(priceValue);
+          } else {
+            formatted = currencyFormat(priceValue);
+          }
+
+          return showTypePrice(formatted, product);
         };
+
+        const hasDiscounts = discountAccountForSingleProduct > 0;
+        const safeFormattedPrice = hasDiscounts ? undefined : product.formattedPrice;
+        const safeFormattedTotal = hasDiscounts ? undefined : product.formattedTotal;
 
         const renderPrice = (
           priceLabel: string,
           priceValue: number,
           priceDiscountedValue: number,
+          preFormatted?: string,
         ) => {
           return (
             <FlexItem
@@ -424,14 +442,14 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
                 <Box
                   sx={{
                     '& #product-price': {
-                      textDecoration: discountAccountForSingleProduct > 0 ? 'line-through' : 'none',
+                      textDecoration: hasDiscounts ? 'line-through' : 'none',
                     },
                   }}
                 >
                   {isMobile && <span>{priceLabel}: </span>}
-                  <span id="product-price">{getDisplayPrice(priceValue)}</span>
+                  <span id="product-price">{getDisplayPrice(priceValue, preFormatted)}</span>
                 </Box>
-                {discountAccountForSingleProduct > 0 ? (
+                {hasDiscounts ? (
                   <Box
                     sx={{
                       color: '#2E7D32',
@@ -507,7 +525,7 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
               </Box>
             </FlexItem>
 
-            {renderPrice('Price', productPrice, discountedPrice)}
+            {renderPrice('Price', productPrice, discountedPrice, safeFormattedPrice)}
             <FlexItem
               textAlignLocation={textAlign}
               {...desktopQtyColumnStyle}
@@ -579,7 +597,7 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
               )}
             </FlexItem>
 
-            {renderPrice(totalText, totalPrice, discountedTotalPrice)}
+            {renderPrice(totalText, totalPrice, discountedTotalPrice, safeFormattedTotal)}
             {renderAction && (
               <FlexItem
                 {...desktopNumericColumnStyle}
--- a/apps/storefront/src/components/B3ProductList.tsx
+++ b/apps/storefront/src/components/B3ProductList.tsx
@@ -409,8 +409,12 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
         };
 
         const hasDiscounts = discountAccountForSingleProduct > 0;
+        // formattedTotal is only valid when qty matches the original order quantity
+        // (partial shipments show a different qty via quantityKey)
+        const qtyMatchesOriginal = quantity === originQuantity;
         const safeFormattedPrice = hasDiscounts ? undefined : product.formattedPrice;
-        const safeFormattedTotal = hasDiscounts ? undefined : product.formattedTotal;
+        const safeFormattedTotal =
+          hasDiscounts || !qtyMatchesOriginal ? undefined : product.formattedTotal;
 
         const renderPrice = (
           priceLabel: string,
--- a/apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
+++ b/apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
@@ -12,7 +12,7 @@ import { isB2BUserSelector, rolePermissionSelector, useAppSelector } from '@/sto
 import { Address, MoneyFormat, OrderProductItem } from '@/types';
 import { verifyLevelPermission } from '@/utils/b3CheckPermissions/check';
 import { b2bPermissionsMap } from '@/utils/b3CheckPermissions/config';
-import { currencyFormat, ordersCurrencyFormat } from '@/utils/b3CurrencyFormat';
+import { ordersCurrencyFormat } from '@/utils/b3CurrencyFormat';
 import { displayFormat } from '@/utils/b3DateFormat';
 import { b2bPrintInvoice, getPrintInvoiceUrl } from '@/utils/b3PrintInvoice';
 import { snackbar } from '@/utils/b3Tip';
@@ -91,6 +91,7 @@ interface OrderCardProps {
   invoiceId?: number | string | undefined | null;
   isCurrentCompany: boolean;
   switchCompanyId: number | string | undefined;
+  currencyCode?: string;
 }
 
 interface DialogData {
@@ -114,6 +115,7 @@ function OrderCard(props: OrderCardProps) {
     ipStatus,
     isCurrentCompany,
     switchCompanyId,
+    currencyCode,
   } = props;
   const displayAsNegativeNumber = ['coupon', 'discountAmount'];
   const b3Lang = useB3Lang();
@@ -198,8 +200,15 @@ function OrderCard(props: OrderCardProps) {
 
   if (typeof infos === 'string') {
     showedInformation = infos;
-  } else if (infos?.money) {
+  } else if (infos?.symbol) {
     const symbol = infos?.symbol || {};
+    const formatValue = (value: string) => {
+      // Legacy path: money is set, use ordersCurrencyFormat
+      // Unified path: values are pre-formatted via formattedV2, display as-is
+      if (infos.money) return ordersCurrencyFormat(infos.money, value);
+      return value;
+    };
+
     showedInformation = infoKey?.map((key: string, index: number) => (
       <Fragment key={key}>
         {symbol[key] === 'grandTotal' && (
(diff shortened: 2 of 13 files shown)


apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
```diff
@@ -187,9 +157,14 @@ function buildOrderProduct(
     imageUrl: string;
   },
 ): OrderProductItem {
-  const unitPrice = item.quantity > 0 ? item.subTotalListPrice.value / item.quantity : 0;
-  const formattedUnitPrice = formatPrice(unitPrice, decimalPlaces);
-  const formattedTotal = formatPrice(item.subTotalListPrice.value, decimalPlaces);
+  // Unit price from list price (per Jesse/Shayne decision — no unit price Money field exists)
+  const unitListPrice = item.quantity > 0 ? item.subTotalListPrice.value / item.quantity : 0;
```
bc-lulu-yang, line 166: [question]

Should `subTotalSalePrice` be used here to calculated instead of `subTotalListPrice`?

apps/storefront/src/pages/OrderDetail/index.unified.test.tsx
```diff
@@ -398,12 +398,12 @@ describe('Order detail path with unified SF GQL flag ON', () => {
...
-    expect(screen.getByRole('group', { name: 'Grand total' })).toHaveTextContent('€431,77');
+    expect(screen.getByRole('group', { name: 'Sub total' })).toHaveTextContent('€102.00');
+    expect(screen.getByRole('group', { name: 'Shipping' })).toHaveTextContent('€332.00');
+    expect(screen.getByRole('group', { name: 'Handling Fee' })).toHaveTextContent('€22.20');
+    expect(screen.getByRole('group', { name: 'Tax' })).toHaveTextContent('€13.50');
+    expect(screen.getByRole('group', { name: 'Discount amount' })).toHaveTextContent('-€37.93');
+    expect(screen.getByRole('group', { name: 'Grand total' })).toHaveTextContent('€431.77');
   });
 
   it('updates currency formatting without refetching the order detail', async () => {
```
bc-lulu-yang: [suggestion]

From my understanding, this test is no longer needed, since no currency call in order detail page now

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
```diff
@@ -27,41 +25,16 @@ import type { OrderDetailsState } from '../context/OrderDetailsContext';
...
-    thousands_token: currency.thousands_token,
-    currency_exchange_rate: currency.currency_exchange_rate,
-  };
-}
-
-function formatPrice(value: number, decimalPlaces: number): string {
-  return value.toFixed(decimalPlaces);
+/** Format a Money value for display, preferring the server-formatted string. */
+function formatMoney(money: Money): string {
+  return money.formattedV2 ?? money.value.toFixed(2);
```
bc-lulu-yang: [suggestion]

For the fallback, can we use the `intlFormat` as well instead of `money.value.toFixed(2)`. From my understanding,`intlFormat` handles currency symbol and decimal rules better than `toFixed(2)`

apps/storefront/src/pages/OrderDetail/components/OrderCheckboxProduct.tsx
```diff
@@ -249,7 +259,7 @@ export default function OrderCheckboxProduct(props: OrderCheckboxProductProps) {
             </FlexItem>
             <FlexItem textAlignLocation={textAlign} padding="10px 0 0" {...itemStyle.default}>
               {isMobile && <span>{b3Lang('orderDetail.reorder.price')} </span>}
-              {currencyFormat(product.base_price)}
+              {product.formattedPrice || formatPrice(product.base_price)}
```
bc-lulu-yang, line 259: [question/nit]

From what I read, this formattedPrice from the unit price, which calculated from 
  const unitListPrice = item.quantity > 0 ? item.subTotalListPrice.value / item.quantity : 0;

I am not too sure of this unit price is the current product list price, if not, might be cause issue.

<del>For example, I ordered 1 product during Xmas sale, and only paid 10AUD at that time. Now it's not sale time, the regular product is 20AUD. So in the reorder, we should not display 10AUD to mislead the buyer.  Not too sure if the component is for reorder or not, but from what I can think, the legacy `product.base_price` make more sense.<del>

apps/storefront/src/pages/OrderDetail/components/OrderCheckboxProduct.tsx
```diff
@@ -52,10 +54,18 @@ export default function OrderCheckboxProduct(props: OrderCheckboxProductProps) {
     catalogInventoryBySku,
     backorderUiEnabled = false,
     showReorderAtsHelper = false,
+    currencyCode,
   } = props;
 
   const b3Lang = useB3Lang();
 
+  const formatPrice = (value: string | number) =>
+    currencyCode
```
bc-lulu-yang: [suggestion]

Seems this format is used in a few place, probably a util function is better

apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
```diff
@@ -198,8 +200,15 @@ function OrderCard(props: OrderCardProps) {
 
   if (typeof infos === 'string') {
     showedInformation = infos;
-  } else if (infos?.money) {
+  } else if (infos?.symbol) {
```
bc-lulu-yang, line 203: [question]

Seems for unified order path, we don't have infos?.symbol. Does this mean the if branch will be skipped?

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
bc-jessebayley: i don't believe so, we are following the (admittedly imperfect) existing behaviour on gql powered orders pages in a non-b2bedition storefront.
Deriving the per unit price from the final discounted price could lead to rounding issues (promotion of 3x product for only $10). It would also be confusing with the rest of the order detail page which shows discounts as a separate line item. We shouldn't be displaying double discounts to a user

bc-lulu-yang: That's a good point. In this case, we might see some "inconsistence" in order detail page. For example, unit price is $20, then total is $10. But the right side summary contains more info, so probably fine for first release.


Just a screenshot (which is legacy order detail) for what I am talking about the "inconsistence" may show up.
Screenshot 2026-07-10 at 12 08 59 pm

apps/storefront/src/components/B3ProductList.tsx
```diff
@@ -387,18 +390,37 @@ export function B3ProductList<T extends ProductItem>(props: ProductProps<T>) {
...
+            }).format(priceValue);
+          } else {
+            formatted = currencyFormat(priceValue);
+          }
 
-          return showTypePrice(newMoney, product);
+          return showTypePrice(formatted, product);
         };
 
+        const hasDiscounts = discountAccountForSingleProduct > 0;
```
bc-lulu-yang, line 411: [question/nit]

seems for unified order path, the `applied_discounts` is never set anywhere for order products, which means `hasDiscounts` always be false for unified order. Is this something we intentional, or we need to follow up in later ticket - (I am not too sure, just thinking if we missed something)

apps/storefront/src/pages/OrderDetail/shared/convertOrderDetail.ts
bc-jessebayley: the inconsistencies is a matter of perspective, and it's partly due to how you interpret line items. I would say that product unit price _is_ 225.00$$$, and any discount gets applied after that. So it gets applied seperate to that price. Therefore showing the original price makes sense. If this order had 5 items, and a $10 discount, should each product have a $2 reduction applied to its per unit price? or some other weird fraction when you factor in qty? We shouldn't make users do maths to understand this page and their order. 
Ideally this page would match the checkout 1:1.

All that being said, i would like to leave the decision on how orders are displayed to shoppers in the storefront to the SMEs, the Orders team. We can influence, and let them know user requirements. But implementation details can be up to them.

bc-lulu-yang: Thanks @bc-jessebayley, make sense to me.

bc-gary: Based on discussion on call, it's agreed to use subTotalListPrice. Hence marking it resolved

apps/storefront/src/pages/OrderDetail/components/OrderCheckboxProduct.tsx
bc-gary: Based on discussion on call, it's agreed to use subTotalListPrice. Hence marking it resolved

apps/storefront/src/components/B3ProductList.tsx
bc-gary: Good catch checking this — `applied_discounts` is also empty on the legacy order path. The `applied_discounts` logic in `B3ProductList` is consumed by other contexts (shopping lists, quotes) but not order detail. Both paths show the same behavior — discount totals appear in the Summary section only, not per-product.

apps/storefront/src/pages/OrderDetail/components/OrderAction.tsx
bc-gary: The unified path does set `infos.symbol` — it comes from `priceSymbol` in `buildOrderSummary()` (convertOrderDetail.ts:78-86), which maps each summary row to its type (`'subTotal'`, `'shipping'`, `'discountAmount'`, `'grandTotal'`, etc.). This flows through `OrderAction` at line 492: `symbol: priceSymbol || {}`.

So the `infos?.symbol` branch IS entered on the unified path. The difference is `infos.money` — on the unified path it's `undefined` (we explicitly clear it), so `formatValue()` returns the value as-is (already formatted via `formattedV2`). On the legacy path, `money` is set and `formatValue()` uses `ordersCurrencyFormat`.

bc-lulu-yang (approved):
LGTM

apps/storefront/src/components/B3ProductList.tsx
bc-lulu-yang: Thanks for the checking, so surprised that legacy order path even didn't have this applied_discounts.

bc-veer (approved):

bc-launchbay:
success [20260713050812](https://launchbay.bigcommerce.net/projects/515/releases/291066) by Launchbay

bc-launchbay:
success [20260713050812](https://launchbay.bigcommerce.net/projects/515/releases/291066) by Launchbay

PR #991, opened 2026-08-06
B2B-5420: Fix unified orders queries against live SF GQL schema

bc-jeffteng:
Jira: [B2B-5420](https://bigcommercecloud.atlassian.net/browse/B2B-5420)

## What/Why?
My Orders, Company Orders, and Order Detail were coded against a schema that diverged from what actually shipped. This fixes the mismatches so those three surfaces work against the live schema, and removes/hides the pieces the server genuinely doesn't support.

Root cause for most of it: the order status pill and status filter were keyed off `status.label`, a merchant-configurable display string the resolver returns in unpredictable casing. The stable value is the enum (`status.value`), so I added one shared mapping and pointed every status lookup at it.

My Orders was also sending fields the server rejects — `companyName`, `companyIds`, and `sortBy` on `customer.orders` don't exist there, so every request 400'd. I removed `sortBy` (unblocking everything else), stopped sending the other two, and hid the search box, company selector, and sort headers on My Orders since none of them can do anything now.

Order Detail had six separate query mismatches. The worst is `returnableQuantity` — it 500s server-side, and since it's a non-null field, that error nulls the entire line-items list, so no line items rendered at all. Removed it along with two other dead selections (`quote`, `extraFields`), and fixed wrong shapes on four more (`downloads`, `payments`, `image`, `history`) so they match what the live schema actually returns.

## Rollout/Rollback
All gated by `B2B-4613.buyer_portal_unified_sf_gql_orders`, off by default. Rollback is flipping the flag back off, no migrations.

Still need to verify live against a real store before this goes on anywhere — haven't done that yet.

## Testing
Unit tests updated/added across every touched surface, all passing. No live verification yet.

[B2B-5420]: https://bigcommercecloud.atlassian.net/browse/B2B-5420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes GraphQL variables and field selections for order list/detail against the live API; incorrect status or filter mapping would break filtering and status pills, and Order Detail previously failed on bad fields like returnableQuantity.
> 
> **Overview**
> Aligns **My Orders**, **Company Orders**, and **Order Detail** with the deployed Storefront GraphQL schema so unified queries stop 400/500ing and status UI matches lookup tables.
> 
> **Status handling:** Adds `orderStatusValue` to map `status.value` (enum) to title-case system labels for colors/i18n, while keeping `status.label` for display. List/detail mappers and customer filters now send **`OrderStatusValue` enum members** (e.g. `PENDING`, `AWAITING_FULFILLMENT`) instead of display or system labels.
> 
> **My Orders (`customer.orders`):** Drops **`sortBy`** from the query and disables column sorting on the unified customer path. Removes **`search`**, **`companyName`**, and **`companyIds`** from customer filters and hides company selector / company filter field; status and date filters remain. **`CursorDetailPagination`** no longer passes `sortBy` for customer order fetches.
> 
> **Order Detail query & conversion:** Updates selections to live shapes—**payments** as a connection with `paymentMethodName`, **downloads** as a plain list (not `edges`), **history** uses `statusLabel` (not `status`/`createdBy`). Stops selecting **`returnableQuantity`** (server 500), **`quote`**, and **`extraFields`**; line item images use **`url(width: 80)`**. **`convertOrderDetail`** follows those shapes and sets **`canReturn: false`** without `returnableQuantity`.
> 
> **Company orders UI:** **Placed by** column sorting is turned off under unified orders (backend incomplete). **`usePlacedByUsers`** page size changes from 100 to 50.
> 
> All behind **`B2B-4613.buyer_portal_unified_sf_gql_orders`**; tests updated across touched surfaces.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0da2c5807ab2cb1ead8979b64f8b633a060eaf08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/pages/order/Order.tsx
+++ b/apps/storefront/src/pages/order/Order.tsx
@@ -121,7 +121,6 @@ function Order({ isCompanyOrder = false }: OrderProps) {
     orderStatuses: getOrderStatuses,
   });
   const customerFilterState = useCustomerOrdersState({
-    companyId: selectedCompanyId,
     orderStatuses: getOrderStatuses,
   });
   const companyFilterState = useCompanyOrdersState({
@@ -136,13 +135,17 @@ function Order({ isCompanyOrder = false }: OrderProps) {
     return legacyFilterState;
   };
 
-  const {
-    activeSort,
-    handleSearchChange,
-    handleFilterChange,
-    handleCompanyIdsChange,
-    handleSetOrderBy,
-  } = getActiveFilterState();
+  const { activeSort, handleFilterChange, handleSetOrderBy } = getActiveFilterState();
+
+  // The customer hook has no search or company-hierarchy filter, so resolve these
+  // directly from the company/legacy state rather than through the union above.
+  const handleSearchChange = isUnifiedCompanyPath
+    ? companyFilterState.handleSearchChange
+    : legacyFilterState.handleSearchChange;
+
+  const handleCompanyIdsChange = isUnifiedCompanyPath
+    ? companyFilterState.handleCompanyIdsChange
+    : legacyFilterState.handleCompanyIdsChange;
 
   const getFilterDataAndOrderBy = () => {
     if (isUnifiedCompanyPath) {
@@ -481,12 +484,19 @@ function Order({ isCompanyOrder = false }: OrderProps) {
             },
           }}
         >
-          {isEnabledCompanyHierarchy && (
+          {isEnabledCompanyHierarchy && !isUnifiedCustomerPath && (
             <Box sx={{ mr: isMobile ? 0 : '10px', mb: '30px' }}>
               <B2BAutoCompleteCheckbox handleChangeCompanyIds={handleCompanyIdsChange} />
             </Box>
           )}
           <B3Filter
+            filterMoreInfo={
+              // The customer filter state has no field to send a company filter to, so
+              // it's hidden here to avoid a visible-but-inert control — see B2B-5420.
+              isUnifiedCustomerPath
+                ? filterMoreInfo.filter((item) => item.name !== 'company')
+                : filterMoreInfo
+            }
             startPicker={{
               isEnabled: true,
               label: b3Lang('orders.from'),
@@ -499,7 +509,6 @@ function Order({ isCompanyOrder = false }: OrderProps) {
               defaultValue: filterData?.endDateAt || null,
               pickerKey: 'end',
             }}
-            filterMoreInfo={filterMoreInfo}
             handleChange={handleSearchChange}
             handleFilterChange={handleFilterChange}
             pcTotalWidth="100%"
--- a/apps/storefront/src/pages/order/Order.tsx
+++ b/apps/storefront/src/pages/order/Order.tsx
@@ -492,7 +492,7 @@ function Order({ isCompanyOrder = false }: OrderProps) {
           <B3Filter
             filterMoreInfo={
               // The customer filter state has no field to send a company filter to, so
-              // it's hidden here to avoid a visible-but-inert control — see B2B-5420.
+              // it's hidden here to avoid a visible-but-inert control
               isUnifiedCustomerPath
                 ? filterMoreInfo.filter((item) => item.name !== 'company')
                 : filterMoreInfo
--- a/apps/storefront/src/pages/order/Order.tsx
+++ b/apps/storefront/src/pages/order/Order.tsx
@@ -218,7 +218,6 @@ function Order({ isCompanyOrder = false }: OrderProps) {
     last?: number;
     before?: string;
     filters: OrdersFiltersInput;
-    sortBy: OrdersSortInput;
   }): Promise<{
     edges: ListItem[];
     totalCount: number;
@@ -324,7 +323,7 @@ function Order({ isCompanyOrder = false }: OrderProps) {
         key: 'orderId',
         title: b3Lang('orders.order'),
         width: '10%',
-        isSortable: true,
+        isSortable: !isUnifiedCustomerPath,
         render: ({ orderId }) => orderId,
       },
       {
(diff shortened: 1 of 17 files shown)


apps/storefront/src/pages/order/Order.tsx
bc-jeffteng, line 1: The current BC orders do not support a sorting feature, so that's why we turned it off if the flag is on

apps/storefront/src/shared/service/bc/graphql/orders.ts
```diff
@@ -102,15 +101,16 @@ export interface OrderDigitalLineItem {
   subTotalSalePrice: Money;
 }
 
-/** Projects OrderDownloadConsignment. */
+/** Projects OrderDownloadConsignment — a plain list element, not a connection node. */
 export interface DownloadConsignment {
-  entityId: number;
```
bc-lulu-yang, line 107: [question]

Even no id in each `DownloadConsignment`? If have more than one digital products, then how to tell which is which

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -355,32 +357,40 @@ function Order({ isCompanyOrder = false }: OrderProps) {
...
       },
     ],
-    [b3Lang, isB2BUser, isSuperAdminNotAgenting, isCompanyOrder],
+    [
+      b3Lang,
+      isB2BUser,
+      isSuperAdminNotAgenting,
+      isCompanyOrder,
+      isUnifiedCustomerPath,
+      isUnifiedOrders,
```
bc-lulu-yang, line 387: [question]

any different here between `isUnifiedCustomerPath`  and `isUnifiedOrders`?

bc-jeffteng: `isUnifiedOrders` =  when flag is ON, not matter what page
`isUnifiedCustomerPath` = when flag is ON and in My Orders page

For example, the sorting feature is supported for the Unified Company Orders page, but not the Unified Customer Orders page, so we need to use `isUnifiedCustomerPath`. 

Sort by `placedBy` is not supported under Unified Order at all, then we use `isUnifiedOrders`

apps/storefront/src/shared/service/bc/graphql/orders.ts
bc-jeffteng: No `ID` on the consignment itself, it's' the schema limit.

Each digital line item already carries its own `entityId`, which is what tells products apart.

bc-akshatkaushik:
[info]

https://refactoring.guru/smells/comments

bc-jeffteng:
> [info]
> 
> https://refactoring.guru/smells/comments

do you mean those in code comment are not appropriate? I can remove that.

bc-akshatkaushik:
> > [info]
> > https://refactoring.guru/smells/comments
> 
> do you mean those in code comment are not appropriate? I can remove that.

Have a read through the article and amend accordingly, just removing the comments isn't the solution unless they are already redundant

> Reasons for the Problem
Comments are usually created with the best of intentions, when the author realizes that his or her code isn’t intuitive or obvious. In such cases, comments are like a deodorant masking the smell of fishy code that could be improved.
The best comment is a good name for a method or class.
If you feel that a code fragment can’t be understood without comments, try to change the code structure in a way that makes comments unnecessary.

apps/storefront/src/shared/service/bc/graphql/orders.ts
```diff
@@ -102,15 +101,16 @@ export interface OrderDigitalLineItem {
...
+  recipientEmail: string;
```
bc-lulu-yang, line 106: [question]

Do we need this to be rendered somewhere? if yes, maybe we need it in convertOrderDetail.ts, otherwise we will lose this.

bc-lulu-yang (approved):
Overall LGTM. 

Seems there are a few decisions we made in this ticket/fix. Can you document it in somewhere to later reference? I am not too sure where is the right place, @bc-gary might can help here.

apps/storefront/src/pages/order/Order.tsx
```diff
@@ -417,7 +422,6 @@ function Order({ isCompanyOrder = false }: OrderProps) {
       return fetchUnifiedOrders({
         ...customerFilterState.paginationVariables,
         filters: customerFilterState.filters,
```
bc-gary, line 424: [nit]
`sortBy` is gone from the API call here but still sitting in the query key ~25 lines up. It's inert since sorting is disabled (value never changes), but might confuse a future reader into thinking it's actually being sent.

bc-gary:
> Overall LGTM.
> 
> Seems there are a few decisions we made in this ticket/fix. Can you document it in somewhere to later reference? I am not too sure where is the right place, @bc-gary might can help here.

@bc-lulu-yang Thanks for the callout. @bc-jeffteng Could you please document all the decisions in the JIRA ticket for this and also put as a note in the PR description please. 

We currently do not have a Decision Record document unfortunately. Let me see if one can be created.

bc-gary (approved):
Nice work @bc-jeffteng 👏 👏

bc-launchbay:
success [20260818015053](https://launchbay.bigcommerce.net/projects/515/releases/295351) by Launchbay

bc-launchbay:
success [20260818015053](https://launchbay.bigcommerce.net/projects/515/releases/295351) promoted by https://launchbay.bigcommerce.net/projects/515/releases/295511 being deployed. Deployed by Launchbay

PR #1005, opened 2026-08-18
B2B-5366: guard app init effect against re-entrant runs

bc-jeffteng:
Jira: [B2B-5366](https://bigcommercecloud.atlassian.net/browse/B2B-5366)

## What/Why?
The app's init effect in `App.tsx` depends on `isB2BUser`, which the effect itself changes partway through a run — `getCompanyInfo()` dispatches `setCompanyStatus`, and `isB2BUser` is derived from that. That retriggers the whole fetch-then-navigate sequence before the first run finishes, so a second run's `gotoPage()` call can fire after the user has already navigated elsewhere in the app, silently redirecting them back to the default page mid-interaction. Added a ref guard so the sequence only runs once per mount.

The loading mask was also tied to `isOpen`, which flips true as soon as a navigation decision is made rather than when init actually finishes, so the redirect happened with no visual indicator. The mask now waits for `isPageComplete`.

## Rollout/Rollback
Merge/revert.

### Tested on Catalyst store 
No impact

https://github.com/user-attachments/assets/db584e6d-8c7e-44d7-b11b-e3cde7c49466

Traced with a Playwright trace of the original bug (background init racing the user's own navigation to Company Orders). Full unit suite passes: 110/110 files, 1391 tests. No new test added for the specific re-entrant-effect scenario — flag if you want one before merge.

[B2B-5366]: https://bigcommercecloud.atlassian.net/browse/B2B-5366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches core app bootstrap, session verification, and default navigation on mount; incorrect guard behavior could block init or leave users stuck behind the mask, but the change is localized and revertible.
> 
> **Overview**
> Fixes **mid-session redirects** during buyer portal startup by ensuring the mount-time init sequence in `App.tsx` runs only once per successful completion.
> 
> The init `useEffect` used to depend on `isB2BUser`, which `getCompanyInfo()` can update partway through the same run. That retriggered the full fetch-and-`gotoPage()` flow and could override navigation the user had already started (e.g. opening Company Orders). A **`hasInitialized` ref** blocks repeat runs; it is cleared when BC session verification fails or init errors so a later dependency change can retry. **`isB2BUser` is removed** from the effect dependency list.
> 
> The **loading mask** now clears when **`isPageComplete`** is set instead of when **`isOpen`** flips, so the mask stays up until init finishes rather than disappearing as soon as a navigation decision is made.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fcda8785e305e8425e45b05292a1b6b9d57d8194. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -265,6 +265,8 @@ export default function App() {
       } catch (e) {
         b2bLogger.error(e);
         showPageMask(false);
+        // didn't actually initialize; let a future dep change retry it
+        hasInitialized.current = false;
         storeDispatch(
           setGlobalCommonState({
             isPageComplete: true,
@@ -278,7 +280,7 @@ export default function App() {
     // due they are functions that do not depend on any reactive value
     // ignore href because is not a reactive value
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [b2bId, customerId, emailAddress, isAgenting, isB2BUser, role]);
+  }, [b2bId, customerId, emailAddress, isAgenting, role]);
 
   useEffect(() => {
     if (quoteConfig.length > 0 && storefrontConfig) {
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -66,7 +66,7 @@ export default function App() {
   const isClickEnterBtn = useAppSelector(({ global }) => global.isClickEnterBtn);
   const isPageComplete = useAppSelector(({ global }) => global.isPageComplete);
   const isDefaultLoginStyling = useRef(shouldUseDefaultLoginStyling()).current;
-  const hasInitialized = useRef(false);
+  const isInitRunning = useRef(false);
   const currentClickedUrl = useAppSelector(({ global }) => global.currentClickedUrl);
   const isRegisterAndLogin = useAppSelector(({ global }) => global.isRegisterAndLogin);
   const { quotesCreateActionsPermission, shoppingListCreateActionsPermission } =
@@ -171,8 +171,8 @@ export default function App() {
     loginAndRegister();
 
     // getCompanyInfo() below can change isB2BUser mid-run and retrigger this effect.
-    if (hasInitialized.current) return;
-    hasInitialized.current = true;
+    if (isInitRunning.current) return;
+    isInitRunning.current = true;
 
     const init = async () => {
       try {
@@ -268,6 +268,8 @@ export default function App() {
             isPageComplete: true,
           }),
         );
+      } finally {
+        isInitRunning.current = false;
       }
     };
 
--- a/apps/storefront/src/App.tsx
+++ b/apps/storefront/src/App.tsx
@@ -66,7 +66,7 @@ export default function App() {
   const isClickEnterBtn = useAppSelector(({ global }) => global.isClickEnterBtn);
   const isPageComplete = useAppSelector(({ global }) => global.isPageComplete);
   const isDefaultLoginStyling = useRef(shouldUseDefaultLoginStyling()).current;
-  const isInitRunning = useRef(false);
+  const hasInitialized = useRef(false);
   const currentClickedUrl = useAppSelector(({ global }) => global.currentClickedUrl);
   const isRegisterAndLogin = useAppSelector(({ global }) => global.isRegisterAndLogin);
   const { quotesCreateActionsPermission, shoppingListCreateActionsPermission } =
@@ -171,8 +171,8 @@ export default function App() {
     loginAndRegister();
 
     // getCompanyInfo() below can change isB2BUser mid-run and retrigger this effect.
-    if (isInitRunning.current) return;
-    isInitRunning.current = true;
+    if (hasInitialized.current) return;
+    hasInitialized.current = true;
 
     const init = async () => {
       try {
@@ -184,6 +184,8 @@ export default function App() {
           if (!isBcLogin) {
             logoutSession();
             showPageMask(false);
+            // didn't actually initialize; let the retriggered run (now a guest) finish it
+            hasInitialized.current = false;
             return;
           }
         }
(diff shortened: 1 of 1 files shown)


apps/storefront/src/App.tsx
```diff
@@ -168,6 +169,11 @@ export default function App() {
   useEffect(() => {
     storeDispatch(setOpenPageReducer(setOpenPage));
     loginAndRegister();
+
+    // getCompanyInfo() below can change isB2BUser mid-run and retrigger this effect.
+    if (hasInitialized.current) return;
+    hasInitialized.current = true;
```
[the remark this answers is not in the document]
bc-jeffteng: Yeah, fixed — switched to an isRunning ref reset in a finally block, so it still blocks the mid-flight re-entry but allows a fresh run once the previous one (including the logout early-return) has actually finished.

apps/storefront/src/App.tsx
```diff
@@ -262,6 +268,8 @@ export default function App() {
             isPageComplete: true,
           }),
         );
+      } finally {
+        isInitRunning.current = false;
```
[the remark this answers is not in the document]
bc-jeffteng: Yeah, fair. Reverted the blanket finally reset — the flag only resets inside the logout branch now, so it stays permanently set after a normal completion and only allows the one retry that logout actually needs.

apps/storefront/src/App.tsx
```diff
@@ -168,6 +169,11 @@ export default function App() {
   useEffect(() => {
     storeDispatch(setOpenPageReducer(setOpenPage));
     loginAndRegister();
+
+    // getCompanyInfo() below can change isB2BUser mid-run and retrigger this effect.
+    if (hasInitialized.current) return;
```
bc-gary, line 174: [question]
No `App.test.tsx` exists so adding one just for this would be heavyweight, but the guard + scoped reset + mask timing change is subtle enough that future init-effect edits could regress it easily.

Is there a lighter-weight angle you've considered — e.g. a Playwright assertion beyond the trace you already ran, or a unit test for `resolveInitNavigation`? Or comfortable relying on the existing suite + manual trace?

apps/storefront/src/App.tsx
```diff
@@ -178,6 +184,8 @@ export default function App() {
           if (!isBcLogin) {
             logoutSession();
             showPageMask(false);
+            // didn't actually initialize; let the retriggered run (now a guest) finish it
+            hasInitialized.current = false;
```
bc-gary, line 188: [question]
The guard reset here handles the logout case nicely. One thing I'm curious about — if `init()` throws (e.g. `ensureBcGraphqlToken()` network blip), the catch block sets `isPageComplete: true` but the guard stays locked. Old code would accidentally re-run init on the next deps change, which could serve as a rough retry.

Probably fine since `isPageComplete: true` already signals "done" and the user would just refresh, but have you thought about whether there's a post-error scenario (masquerade toggle, session restore) where a re-init would actually be desirable?

apps/storefront/src/App.tsx
bc-jeffteng: I think it would be a tautology to write a test for just the two-line changes for this, but I'm planning to refactor the whole component and probably extract some parts to help us unit test and cover that code. This will be in a follow-up ticket, not this one.

apps/storefront/src/App.tsx
bc-jeffteng: Good catch. I added `hasInitialized.current = false` in the catch block, so once it failed, it was still making it do false

apps/storefront/src/App.tsx
bc-gauravsobti: [Question]
isB2Buser is the dependency responsible for re-render of the useEffect.
Since inside the effect we are getting isB2BUser directly from the redux state, why not just remove this dependency, then we don't need hasInitialized fix?

bc-lulu-yang: Seems in this useEffect block, no code is using `isB2BUser`. So I think we can remove isB2BUser from the dependency list. 

But in this useEffect, B2BUser is not the only one which can trigger the second init. For example, on the guest path L213`(!customerId)` the same effect calls getCurrentCustomerInfo(), which writes customerId, which can trigger the same second init. So the `hasInitialized` is still a valid fix to stops the flips.

So my suggestion is to remove isB2BUser from the dependency list and keep hasInitialized fix.

apps/storefront/src/App.tsx
```diff
@@ -168,6 +169,11 @@ export default function App() {
   useEffect(() => {
```
bc-lulu-yang, line 169: [suggestion]

Seems catalyst SF also using the component to load b2b buyer portal, pls verify the fix in catalyst SF as well. And ping me if you need a catalyst SF in integration store

bc-lulu-yang (approved):

apps/storefront/src/App.tsx
bc-gary: Has this been covered?

apps/storefront/src/App.tsx
bc-jeffteng: Tested

apps/storefront/src/App.tsx
bc-jeffteng: updated

bc-gary (approved):

bc-gauravsobti (approved):
LGTM

bc-launchbay:
success [20260820020620](https://launchbay.bigcommerce.net/projects/515/releases/295737) by Launchbay

bc-launchbay:
success [20260820020620](https://launchbay.bigcommerce.net/projects/515/releases/295737) promoted by https://launchbay.bigcommerce.net/projects/515/releases/295753 being deployed. Deployed by Launchbay

PR #1009, opened 2026-08-20
Revert "B2B-5366: guard app init effect against re-entrant runs"

bc-jeffteng:
Reverts bigcommerce/b2b-buyer-portal#1005

After PR#1005

https://github.com/user-attachments/assets/066346d6-6e18-418b-80bb-9f0ebebee64f

Expect behaviour

https://github.com/user-attachments/assets/f19a3ad3-ff68-496f-bcb7-1543542d3ea2

---

> [!NOTE]
> **Medium Risk**
> Restores duplicate init work and race-prone behavior when `isB2BUser` or session state changes during bootstrap, which motivated the original guard; mask timing on `isOpen` may reintroduce login flicker the guard was meant to avoid.
> 
> **Overview**
> **Reverts the B2B-5366 change** that used a `hasInitialized` ref to run the main `App` bootstrap `useEffect` only once, including resets on failed BC login or init errors.
> 
> Init again runs whenever its dependency list changes, with **`isB2BUser` added back** so a mid-init flip from `getCompanyInfo()` can trigger a full re-run instead of being skipped.
> 
> The loading mask is again cleared when the **portal iframe opens** (`isOpen`) rather than when `isPageComplete` becomes true; the separate effect that removes the **pre-mount login mask** still waits for `isPageComplete` so the native `login.php` form does not flicker through.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a015690968467f07a45cd16f1290b4c238b5b0b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1010, opened 2026-08-20
B2B-5366: Re-land init guard, fix hierarchy switcher regression

bc-jeffteng:
Jira: [B2B-5366](https://bigcommercecloud.atlassian.net/browse/B2B-5366)

## What/Why?
Re-lands the `hasInitialized` guard on `App.tsx`'s bootstrap effect (reverted in #1009), and extracts the init logic into `initializeApp()` in `utils/initializeApp.ts`.

#1005 guarded the bootstrap effect against re-entrant runs, but on a fresh login `setStorefrontConfig()` — which gates the company hierarchy switcher on redux permissions — ran before `getCurrentCustomerInfo()` dispatched those permissions. The old (unguarded) effect worked by accident: it re-ran a second time once `customerId`/`role` updated, and by then permissions were set. The guard blocked that second run, so the hierarchy switcher stopped showing on My Orders. I moved customer identity resolution before the permission-gated config fetch so it's correct in one pass, no second run needed.

I also pulled the ~110-line inline `init()` out of `App.tsx` into `initializeApp()`, with explicit params instead of closures, returning `{ completed }` so `App.tsx` only owns the guard ref.

This also fixes the jump back to My Orders that survived the guard: the post-login re-init re-dispatches the store configs, which re-assigned `isInit` — and every `isInit` assignment replays the link click stored before the app mounted. Once logged in, that stale "Sign in" click resolves to `/orders`, so the portal showed a full-page loading mask and yanked the user off whatever page they were on. The replay is now one-shot (`clickedLinkElement` is cleared when consumed) and `isInit` is only assigned when it isn't already true.

## Rollout/Rollback
Originally shipped this behind `B2B-5366.prevent_premature_orders_redirect`, read once at mount from a localStorage snapshot written by the previous page load. That means the fix only kicks in from a session's second page load onward — first load, and any fresh browser session, still runs the old unguarded path, which is exactly where this bug shows up. That kills the point of the PR, so I removed the flag. The guard now runs unconditionally, no fallback path. Rollback is a revert.

## Testing
Added `initializeApp.test.ts` — covers the identity-before-config ordering, the invalid BC session path, the error/retry path, and navigation delegation. Added `main.test.ts` — the stored click replays exactly once. All changes are covered by the unit tests.

[B2B-5366]: https://bigcommercecloud.atlassian.net/browse/B2B-5366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes core app bootstrap, login/session validation, and init navigation ordering; regressions could affect login flows, portal open behavior, and company hierarchy visibility.
> 
> **Overview**
> Re-lands a **bootstrap init guard** on `App.tsx` and moves the former inline startup flow into **`initializeApp()`**, so init can finish in one pass without relying on a second effect run when Redux updates `customerId` after login.
> 
> **Ordering fix:** `getCurrentCustomerInfo()` now runs **before** permission-gated work such as `setStorefrontConfig()`, so company hierarchy and similar UI see the correct role on the first init (fixing the hierarchy switcher regression when the guard blocked re-entry).
> 
> **Guard behavior:** Refs track in-flight init and the last successfully initialized customer id; mid-init identity updates no longer retrigger duplicate init. The helper returns `{ completed, resolvedCustomerId }` for the guard to distinguish retries vs. same-session identity.
> 
> **Post-login redirect/mask:** Page mask clearing waits on **`isPageComplete`** instead of portal `isOpen`. Stored pre-mount link clicks replay **once** (`clickedLinkElement` cleared in `main.ts`), and `isInit` is only set when not already true, avoiding stale “Sign in” clicks yanking users back to My Orders.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1446d7f374aff6ceb9e2c31ff365f5d1d2da269. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/storefront/src/App.tsx
```diff
@@ -311,12 +228,11 @@ export default function App() {
   ]);
 
   useEffect(() => {
-    if (isOpen) {
+    if (isPageComplete) {
       showPageMask(false);
```
bc-gary, line 234: [question]
This changed from watching `isOpen` to `isPageComplete` — so the mask is now hidden once at init completion rather than every time the portal opens. I traced through the post-init click flow and it looks safe (`handleAccountClick` → click handler effect → `showPageMask(false)` manages it independently), but want to confirm: is there any other code path that calls `showPageMask(true)` and previously relied on the `isOpen` effect to clear it?

apps/storefront/src/utils/initializeApp.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    vi.mocked(getCurrentCustomerInfo).mockResolvedValue({
+      role: CustomerRole.SENIOR_BUYER,
+      userType: UserTypes.MULTIPLE_B2C,
+    } as never);
+    vi.spyOn(store, 'getState').mockReturnValue({
+      company: {
+        customer: { id: 123, role: CustomerRole.SENIOR_BUYER, userType: UserTypes.MULTIPLE_B2C },
+        companyInfo: { status: CompanyStatus.APPROVED },
+      },
+      global: { featureFlags: {} },
```
bc-gary: [suggestion]
`Object.defineProperty` replaces `window.location` here but `afterEach` only calls `vi.restoreAllMocks()`, which doesn't undo property overrides. Subsequent tests that call `initializeApp` (which reads `window.location.hash`) inherit the modified object. Since `{ ...window.location }` spreads a `Location` instance whose properties are mostly prototype getters, the spread may produce a partial object.

Consider saving/restoring the original, or using `vi.stubGlobal`.

apps/storefront/src/utils/initializeApp.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    } as unknown as ReturnType<typeof store.getState>);
+    Object.defineProperty(window, 'location', {
```
bc-gary, line 157: [suggestion]
`Object.defineProperty` replaces `window.location` here but `afterEach` only calls `vi.restoreAllMocks()`, which doesn't undo property overrides. Subsequent tests that call `initializeApp` (which reads `window.location.hash`) inherit the modified object. Since `{ ...window.location }` spreads a `Location` instance whose properties are mostly prototype getters, the spread may produce a partial object.

Consider saving/restoring the original, or using `vi.stubGlobal`.

apps/storefront/src/App.tsx
bc-jeffteng: `showPageMask(true)` only use in 3 place:
- App.tsx: 73 
- B3StoreContainer.tsx:47
- Dashboard/index.tsx:146

None of them ever relied on a second or later `isOpen` toggle to get cleared. They all resolve at or before the `init` finish point that `isPageComplete` is now represented once.

apps/storefront/src/App.tsx
```diff
@@ -168,101 +158,28 @@ export default function App() {
...
-        } else {
-          showPageMask(false);
-        }
-
-        if (customerId) {
-          clearInvoiceCart();
-        }
+    // initializeApp() can dispatch a resolved customerId mid-call, retriggering this
+    // effect for the identity it's already handling; skip that self-retrigger.
+    if (isInitializing.current || initializedCustomerId.current === customerId) return;
```
bc-lulu-yang, line 164: [Nit]

Feel like a strict === on id can be risky if one side is 123 and the other is '123'.  Probably `Number(initializedCustomerId.current) === Number(customerId)`

bc-jeffteng: In JS, `Number(null) === 0`, so if `customerId` = 0 and `initialzedCustomerId.current` = null, `Number(initializedCustomerId.current) === Number(customerId)` will pass. As the `customerId` is only `number`, it's safe to keep the original way.

apps/storefront/src/utils/initializeApp.test.ts
bc-jeffteng: Fixed

bc-gary (approved):
Nice work 👏

bc-lulu-yang (approved):

bc-launchbay:
success [20260825055244](https://launchbay.bigcommerce.net/projects/515/releases/296208) by Launchbay

bc-launchbay:
success [20260825055244](https://launchbay.bigcommerce.net/projects/515/releases/296208) by Launchbay

